PR #10, opened 2015-02-05
Postgresql profile

leeevans:
Added a maven build profile for postgresql, so we can activate it and just override the spring.datasource.url to establish a postgresql database connection for integration testing in the CI build server.


alexfranken:
Cool, looks good.  Do you plan to have the CI server poll for changes in the repo and trigger a build? And deploy to your dev tomcat server too?

chrisknoll:
Lee, there's something up with this pull request.  It seems to show a very large merge into your branch as part of this pull request.  Do you think you can isolate just the changes you made to add the maven profile in a smaller pull request?

chrisknoll:
Ah, I could be mistaken, it looks like when looking at the files changed, it only has the +20 lines added to pom.xml.

leeevans:
Hi Alex,


Yes that's right. We should be able to use a github commit webhook to trigger CI builds and avoid polling for changes.

I'm thinking we may need three WebAPI build jobs in the CI server - one for each DBMS.  I hope to work on those build jobs tomorrow.  


We will need the three jdbc drivers on the CI server to execute the database integration tests which is no problem for postgresql because it's freely available in maven central but the other two are proprietary so maybe we need a locked down section of an OHDSI private repo for those. 


Thanks for all your help. I'm learning a lot.


regards,
Lee

---- On Wed, 04 Feb 2015 21:58:42 -0500 alexfranken&lt;notifications@github.com&gt; wrote ---- 

Cool, looks good. Do you plan to have the CI server poll for changes in the repo and trigger a build? And deploy to your dev tomcat server too?

leeevans:
yes, it's a really small change to the pom.xml file. 

The large merge was applied to my old github webapi fork, so I could first bring it up to date with everyone else's commits.  Maybe there's a better way to do that?


Lee.

---- On Wed, 04 Feb 2015 22:08:58 -0500 Chris Knoll&lt;notifications@github.com&gt; wrote ---- 

Ah, I could be mistaken, it looks like when looking at the files changed, it only has the +20 lines added to pom.xml.

chrisknoll:
I am not sure it will matter, when i look at the total files changed, it only says 1 file, 20 lines.  I'll merge it.

PR #12, opened 2015-02-06
Job Server - Spring batch implementation

alexfranken:
I added some documentation in the README.  
I also created JobServiceIT for integration testing (java client perspective).
Example usage from a service perspective can be seen in the exampleapplication package.
Obviously, when ready, we should be able to move this package/function to test source.
Spring Batch will try and create the tables,sequences for you upon startup, vendor derived from DataSource.  There is a isolation-related property in the pom should that become an issue with non-oracle DBs.
Let me know what questions/feedback you have.


testdev123:
I like the new spring batch feature. Spring is pretty cool.

Will the new batch tables be created in the cdm results schema? That would work well because many folks make their cdm schema read only.

I noticed some of the new code includes Regenstrief Institute Licence and Copyright 'boilerplate' text. All the code is already Apache open source licensed and I'm not aware of any OHDSI discussion to-date around what copyright for any OHDSI code.  Was this perhaps introduced in a copy paste of existing Regenstrief code? Suggest we remove it or there could be confusion for end users/developers who want to use the tools or contribute to further development of the tools.

alexfranken:
Lee,
Sorry about the Regenstrief copyright... I thought I had removed them all.  It is part of my Eclipse IDE config when I create a new class.  I removed it from the pull request.

Regarding the Spring Batch DDL...  Currently, the tables are auto-created in the schema associated with the datasource user.  So, for our usage at Regenstrief, this happens to be our results/ohdsi schema.  The batch tables should be located in this schema.  If the expectation is that the DataSource user is associated with the results/ohdsi schema, then we could leave the auto generation as is.  You all probably know better than I do if that is reasonable.  At Regenstrief, we do not have write access to the CDM schema.  I don't know enough yet to say that is consistent (actually, I believe I heard that there is a results table in the CDM schema which would imply that some may be specifying the CDM schema user?)...
Chris, Frank & I spoke a little about auto-generating DDL (e.g. having the app maintain its schema across versions) on Tuesday, but not landing on anything concrete.  Though that is possible, it is not uncommon to find .sql scripts in projects... There are projects like liquibase (though I haven't used it) that can help with maintaining your schema.
The Spring Batch DDL scripts for each supported DB vendor are embedded in the spring-batch-core jar and we have limited control when using the default auto-gen feature.  I can add a property to toggle the support for generating the DDL.  With that said, it doesn't change the behavior.  We could add support for property-driven DDL initialization, where the user would specify the location to the create and drop scripts (letting the user control schema prefixes to create table command, etc.).  We could even try the sql template / render/translation that you all came up with?
So, we have some options here.  At this time, there is nothing preventing the app from starting up if the batch tables don't exist so we could disable auto gen as the default...  If you all think this is the right move, let me know and I can add it to the pull request.

There is a 'tableprefix' property that drives the data access of the spring batch tables.  The default is 'BATCH_' so you will see tables created like 'BATCH_JOB_XYZ', etc.  So, if the batch tables are created as a pre-req step, and in a different schema than the user associated with the datasource in schema results, this will have to be adjusted (e.g. RESULTS.BATCH_).

leeevans:
No worries.

I agree with the approach that the batch tables should be located in the ohdsi (aka results) schema.  The physical name of that schema could vary with different end-user deployments but it will always be specified in the datasource.ohdsi.schema. property.

You may be referring to the cohort table in the CDM schema which is somewhat of an exception. I would recommend that the CDM schema should be treated as read-only for the OHDSI tools (which agrees with the approach here).

If it's feasible to include the end-user supplied datasource.ohdsi.schema. property value in the tableprefix for the BATCH tables, that would be preferred. With an Oracle database, user and schema are synonymous but that's not the case with Postgresql and SQLServer. 

Chris makes some really good points in the below discussion forum post around separating out the regular application account's database permissions from an admin account with higher elevated permissions to execute DDL for db migrations:
http://forums.ohdsi.org/t/managing-database-migrations/291

I think the same argument also applies to the initial DDL creation and we should run the auto-gen with a separate 'admin/migration' database account. Do you think it would make sense to add two additional pom.xml properties for admin user and password, to support that? 

I'm in favor of enabling auto-gen by default if it can be configured to connect to the database as a separate admin db user and it can create the batch tables in the ohdsi schema (using the datasource.ohdsi.schema. property value in the tableprefix).

chrisknoll:
Is this pull have auto-ddl disabled?  If not, can we get that set, and then merge this in?  I think we'd like to get started on firing jobs and also incorporate changes required for the JPA-Hibernate integration.

This leads to the question of where will the required DDL come from, but we can discuss that at the forums.

-Chris

alexfranken:
Chris, I just committed the change to disable any datasource initialization. Thanks.

cahilton:
Are we ready to merge then?

chrisknoll:
Done.

PR #13, opened 2015-02-11
Heracles cohorts

cahilton:
This is a pretty big commit, but most changes are isolated to cohort_analysis (Heracles run)/cohort_results (Heracles view). We'll certainly have more changes, but they will most likely be smaller, and because we were merging back from master, that added to the commits.

* Changes to `pom.xml` and `application properties` to support new properties used in Heracles sql
* Cohort Analysis - Tasks around submitting analyses for Heracles
    * Model Objects
    * Tasklet - used for submitting a task to the Spring batch job service
    * Service 
        * Get a list of all analyses
        * Get a list of all analyses and their status for a given cohort definition
        * Summary - get basic information about a cohort definition and base analyses (e.g. demographics)
        * Preview the SqlRender output of Heracles analysis SQL
        * Run the SqlRender output of Heracles analysis SQL using a Spring Batch job
     * SQL files to support service activities
     * Unit tests for submitting jobs
<br/>    
* Cohort Results - Tasks around view completed analyses for Heracles
     * Model Objects
     * Service
         * Get analysis results by name for a cohort (I implemented this generically, and may not be what we use going forward/sync up with Achilles, but it was a quicker way to get going to test if our jobs are working)
     * SQL files to get cohort results
<br/>
* Cohort
   * Remove CohortService since there's no current use case for direct access to cohort table data


alexfranken:
I notice there is a CohortAnalysisService and CohortResultsService.  Since we make the distinction at the service level, should we move the 'CohortAnalysis*' classes from cohortresults package to a new cohortanalysis package?

cahilton:
Probably. There's nothing really in cohort results yet, but there probably will be. I can go ahead and move for consistency.

chrisknoll:
I'm happy to merge this in if it's at a good point. I think once we get up to ~10k lines in a pull request, it's going to be hard to digest exactly what is involved in the change.  Any way to group the service properties into some logical unit? I imagine once we get a lot of service apis online, that file is going to be a heap.  But if this is the cost of a singular .warplatform, then that's the way it's gotta be.

Frank, any comments?

cahilton:
Yeah - I intended to make it smaller, but then it sorta became all the stuff we needed to get step 1 of Heracles off the ground. We'll try to group them up better going forward.

pom.xml
```diff
@@ -31,6 +32,18 @@
...
+		
+		<datasource.cohort.schema>[CDM_DATABASE].dbo</datasource.cohort.schema>
+		<datasource.cohort.table>cohort</datasource.cohort.table>
+		
+		<datasource.achilles.results.table>achilles_results</datasource.achilles.results.table>
+		<datasource.achilles.results_dist.table>achilles_results_dist</datasource.achilles.results_dist.table>
+		<datasource.heracles.results.table>heracles_results</datasource.heracles.results.table>
+		<datasource.heracles.results_dist.table>heracles_results_dist</datasource.heracles.results_dist.table>
+		
+		<!-- CDM properties -->
```
fdefalco: What are these properties being used for?

pom.xml
```diff
@@ -23,6 +23,7 @@
 		<spring.datasource.password></spring.datasource.password>
 		<spring.datasource.driverClassName>com.microsoft.sqlserver.jdbc.SQLServerDriver</spring.datasource.driverClassName>
 		<spring.datasource.url>jdbc:sqlserver://serverName;user=userName;password=password;</spring.datasource.url>
+		<spring.datasource.validationQuery>SELECT 1</spring.datasource.validationQuery>
```
fdefalco: The validation query here, how is this being used differently from other queries that we would want stored in resource files and run through sql translate? I just want to ensure we aren't requiring people to edit the property files to work with their particular database.

fdefalco:
I put a few comments inline, which I had never done before, but its cool. :) Also nice that they show up here in the main conversation thread. Gotta love github.  The only other comment I have is about the use of all caps SNAKE_CASE.  I appreciate that you were following through with what we had done for the achilles related results but Chris and I were discussing trying to get away from using it.  If you remember the history, we used it because originally the R analyses were using this naming convention.  For the CDMResultsService I was trying to maintain compatibility with the achilles application should we ever get around to porting it to use the service as opposed to reading the serialized JSON files.  That being said, I think you should break from that naming convention within Heracles if you are so inclined.  I think @chrisknoll agrees, but will let him confirm.

pom.xml
alexfranken: Not sure we will be able to avoid db-specific props (e.g. jdbc URL, dialect, etc).  With Flyway, you have to write db-specific .sql files.  That didn't seem to be a show stopper in the db migration thread.  The validation query is used by connection pooling libraries to validate a connection when it is handed out of the pool.  'SELECT 1' is said to work in quite a few databases (Oracle is an exception)

pom.xml
cahilton: They're properties in the SQL that Patrick wrote. I think there is some support in his scripts for v4 vs. v5, and I think source name is one analysis type.  So, I was just aligning our properties with his scripts. If I should hardcode those to something, just let me know.

cahilton:
Yea, re: SNAKE_CASE, just trying to align conventions, but for JSON I don't have a strong preference over another. I'm fine to go with just camelCase, which is what we would get by default.

src/main/resources/resources/cohortresults/sql/init/createHeraclesResultsTables.sql
```diff
@@ -0,0 +1,320 @@
...
+
+{DEFAULT @CDM_schema = 'CDM_schema'}    --CDM_schema = @CDM_schema
+{DEFAULT @results_schema = 'scratch'}   --results_schema = @results_schema
+{DEFAULT @cohort_schema = 'CDM_schema'}  --cohort_schema = @cohort_schema
+{DEFAULT @cohort_table = 'COHORT'}  --cohort_table = @cohort_table
+{DEFAULT @source_name = 'TRUVEN MDCD'}   --source_name = @source_name
+{DEFAULT @smallcellcount = 5}    --smallcellcount = @smallcellcount
+{DEFAULT @createTable = TRUE}    --createTable = @createTable
+{DEFAULT @runHERACLESHeel = TRUE}   --runHERACLESHeel = @runHERACLESHeel
+{DEFAULT @CDM_version = '4'}  --we support 4 or 5,   CDM_version = @CDM_version
```
cahilton: @fdefalco Here's the Heracles SQL that uses those properties.

cahilton: Also in runHeraclesAnalyses.sql, since we split out the init of the tables and the running of the analyses.

chrisknoll:
CammelCase would be great.  As far as the other properties, I know it makes sense from Patrick's perspective to have a setting specifically for the cohort table, but not sure if that's a good strategy for the application.  For the other applications (and I'm pretty sure there will be many) that want to use cohorts, will they need to all be configured similarly?  Or should there just be a well known, defined place to store cohorts?  Should I make the cohort definition table a configurable property?  I'm wondering if that will just lead to a heap of analysis specific tables...  Its easier to optimize a database when you know what tables are going to be in there.  having a defined table means you can grant permissions to a specific thing without opening up the entire schema to have create rights on it...  most db optimizations can compile plans based on the db schema. If every time we do an analysis it's on a different schema, that pretty much eliminates that possible optimization.

All this being said, I understand you were following the lead of the scripts you were given in Heracles so if there were parameters for tablenames, you used them.  We just might want to consider that these scripts were not written with system architecture in mind and they might need a little refinement in order to fit into a context of a 'system' or 'process'.

-Chris

pbr6cornell:
Translation:  Patrick didn't know what he was doing when he wrote those
heracles scripts, so don't treat them as gospel.

And I 100% completely agree:)

cahilton:
Fair enough - just trying align with what I see. I'm not always sure what is gospel and what isn't. :)  I'll mod these to look just for analyses to be in OHDSI/Results schema. That still works for you, right @jduke99?

jduke99:
Yep works for me!

chrisknoll:
Ok, great. Well, we have to start somewhere, and this looks like the best place to start. This works as a good model for setting up asynchronous jobs, uses sql render for the processing, etc.  I think only thing i'd echo from Frank's perspective is storing a select statement in one of the configuration properties (as if that's something that should be configured as part of the application) but I wasn't part of the decision making that put it there, and I'm also not here to secondguess anyone, so we'll just merge this in, and make adjustments if it makes sense.

cahilton:
Properties changes have been put into master.

cahilton:
camelCase has been put into master, along with corresponding changes to Heracles

pom.xml
fdefalco: Avoiding db specific queries has been something we've been able to achieve through the SqlRender library.  We've also adopted the approach of putting sql statements in resource files as opposed to in the properties.  I understand this is a short 1 liner but I would rather it be in a resource file and leverage SqlRender.  We can work with Martijn to ensure it supports what you need it to support.

chrisknoll: Yes, if this line 'select 1' is just a simple query to check for the health of the connection setting, there's probably other ways to do this in code where you can leverage the JDBC api to get meta out of the database without putting a SQL statement in the property file.  If that failes, it should at the very least be put into a separate propety file, sql translated, and executed on the datasource.

alexfranken: What would this look like in your template sql?  SELECT 1 FROM DUAL (oracle), SELECT 1 (others), etc.?

chrisknoll: Well, I did some reading and it feels (granted, I haven't been in the java space for a while) like the validationQuery is a sort of stop-gap from a time when the JDBC api did not support an 'isValid' method that does a database specific (as implemented in the JDBC driver) check that guarantees that the connection is open and valid.  I do not know why mondern day connection pools do not do this behind the scenes before returning a connection from the pool to the caller, but it feels like this mode of working is a throwback to those old days.  

Could we possibly find a more modern approach for determining the health of a connection in the pool?  If not, I'm hesitantly 'ok' with this _one_ case of storing a sql clause in the config because that's what the third party library deepens on (but I believe this is an optional paramater that you don't need to specify, so I'd be better OK with not doing this check in the framework, and just halding SQL exceptions at the place in code that the connection is used).
 
But just want to make clear that writing code for WebAPI should not have any custom SQL in our on properties.  That should be avoided at all costs.

In reading further, it seems that there is a few connetion pool implementations (for Tomcat, for example) which allow you to specify a 'validatorClass' which gets instantiated to do the check, instead of putting a query in a config file.  In a validation class, we can just call isValid() on the connection that we're validating and be done with it.

I don't think template sql would solve this particular issue because it is the spring framework that is using this query, and it won't know to translate the template sql before invoking it.

-Chris

alexfranken: We could use template sql & SqlRender, when setting up the DataSource, as the validationQuery is a tomcat pool property.  See the DataSourceConfig in my jpa branch for DataSource construction.  With that said, due to the issues with pooling, oracle, and temp tables, I have changed the DataSource to return a new connection for every request (i.e. no pooling).  Hopefully, we will be able to revert and use pooling.  When that time comes, we either use a validatorClass or the template sql (I am still curious what the template sql would look like in this case).

https://github.com/OHDSI/WebAPI/blob/jpa/src/main/java/org/ohdsi/webapi/DataAccessConfig.java

PR #14, opened 2015-02-16
Conceptset expression impelmentation

chrisknoll:
Unfortunately, this pull also grabbed 2 small changes I made with respect for the cohort definition generation, so I'm sorry for that mistake, please disregard those change

This pull request introduces the ConceptSetExpression to the WebAPI codebase.  This includes new classes to encapsulate the data, and a ConceptSetQueryExpressionBuilder which takes a ConceptSet Expression and creates 'template sql' out of it. (Sql that still should be translated and tokens such as @CDM_schema replaced).  Additionally, I've added a small unit test to verify a test expression contains the proper IN clauses in the output SQL.  

I placed these files under the Vocabulary service since querying concepts seems to be in the domain of the Vocabulary.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
+++ b/src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
@@ -17,14 +17,6 @@ public class ConceptSetExpression {
     public boolean includeDescendants;
     public boolean includeMapped;
   }
-  
-  public static class Concept {
-    public long conceptId;
-    public String conceptName;
-    public String conceptCode;
-    public String domainId;
-    public String vocabularyId;
-  }
 
   public ConceptSetItem[] items;
   
--- a/src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
+++ b/src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
@@ -1,0 +1,31 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.ohdsi.webapi.vocabulary;
+
+/**
+ *
+ * A Class that encapsulates the elements of a Concept Set Expression.
+ */
+public class ConceptSetExpression {
+  public static class ConceptSetItem
+  {
+    public Concept concept;
+    public boolean isExcluded;
+    public boolean includeDescendants;
+    public boolean includeMapped;
+  }
+  
+  public static class Concept {
+    public long conceptId;
+    public String conceptName;
+    public String conceptCode;
+    public String domainId;
+    public String vocabularyId;
+  }
+
+  public ConceptSetItem[] items;
+  
+}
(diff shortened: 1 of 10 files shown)


src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
cahilton: This is minor, but I think you have some artifact of your IDE here in the header.

src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+  
+  public static class Concept {
+    public long conceptId;
+    public String conceptName;
+    public String conceptCode;
+    public String domainId;
+    public String vocabularyId;
+  }
+
```
cahilton: Could this use org.ohdsi.webapi.vocabulary.Concept? It looks like all the fields are there, although it does use SNAKE_CASE json naming.

src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
chrisknoll: Yep, all my files look like this. Do we have a standard language we want to put into our headers?

src/main/java/org/ohdsi/webapi/vocabulary/ConceptSetExpression.java
chrisknoll: Few things on the Vocabluary.Concept:
1. JSON snake case naming is used, and the front end that manipulates this structure is using camelCase, so i introduced this Concept class as a static-inner class to keep it hidden and unobtrusive to existing concepts. 
2. the GetInvalidReason and GetStandardConcept are in the Vocabulary.Concept class, I didnt' want to include it, as it has special behavior in it that I'm not sure how I would handle.
3. I planned on including this 'inner' concept type just for the pull request, if we do identify that we can indeed just standardize on one form, that can be applied in a future commit?  I am worried about the cascading changes that may be required in Hermes if the concept structure changes.

Addressing the duplicate Concept type can be definitely be the first thing to address after incorporating this code base. Your are absolultely right that there's a bit of codesmell having 2 Concepts defined that serve seemingly the same purpose.

-Chris

chrisknoll:
Hi, Charity.  I switched it up to use the concept and removed the Concept class I gave you. Other than that, it was just the headers, right? I'm not sure how to address that but it's not isolated to the pull, i think all of our files have this issue.

cahilton:
Yeah - that's it.

We might want to start thinking of a group what we want to put in our headers. We have a common one at RI that can be setup automatically in Eclipse.

chrisknoll:
Ok, I applied the standard language that came with netbetans for Apache 2.0. Let me know if there's any problems with this.

And if there aren't any other concerns, i'll merge this branch later tonight.

PR #15, opened 2015-02-16
Flyway - Database Change Management

alexfranken:
Flyway will create a schema_version table and apply migrations upon startup.  The idea is when a feature is introduced the appropriate DDL would be created in the src/main/resources/db/migration/${vendor} directory, with the version incremented (e.g. next feature would introduce a V1.0.0.2__new_feature.sql for each supported db).  The commits should illustrate this.  No two files can have the same version so, the convention proposed is to model after the artifact version (1.0.0 major.minor.maintenance) and add an "incremental" type of digit for each feature.  Let me know what questions you have.

* You will see a failure if running WebAPI and the spring batch tables already exist.  Remove first… (see src/main/resources/db/migration/${vendor}/rollback/1.0.0.1__schema-drop_spring_batch.sql)
* Configure Flyway datasource (datasource credentials for ohdsi/results schema).  See flyway.schemas property to define Flyway managed schemas (defaults to property 'datasource.ohdsi.schema')
* Supporting 2 DataSources, which includes the change of property names (e.g. spring.datasource.x -> datasource.x).
* Also, note the shift to remove the USER/PASS from the jdbc.url string.  Passwords can end up in log files, etc. Use of separate properties is encouraged.  See the oracle profile for an example.  We will need to have Lee update the postgresql profile for the CI server.


chrisknoll:
Thanks Alex. I'll play with it.  I'm feeling a little daunted by the different configuration profiles, but I'm getting the impression that the main POM files have different environment profiles defined, but in the ~/.m2 folder's settings.xml you can provide additional (possibly private) configuration options.  Are those our options (either in main POM or local settings.xml) or did you have another option in mind to pass external file properties along into a maven process?  I guess it would be helpful for me to just have some examples of how you'd specify different settings in the different locations.  Should I ask this on the developer forums?

This is very good tho, it seems like I can set up a local DB on sql, and set the jdbc url and credentails in a property, and fire up the app and I'll have tables.  This is good because we need to implement a few more tables which at this point I think the opinion is that the tables are 'just there', but this isn't actually the case.  

Thanks again, Alex!

alexfranken:
Chris, you have it right.  Settings.xml should be used for sensitive properties (e.g. passwords) and any one-off configurations, overriding defaults in pom.xml... For example, we really don't have to have the oracle and postgres profiles in the pom.  They are really just for reference, giving an idea of the properties that commonly vary across environments.
An alternative to settings.xml profiles is specifying java/jvm properties. `mvn spring-boot:run -Ddatasource.password=sensitive -Ddatasource.dialect=oracle`
These properties trump all.  It is common to see these in CI server jobs/plans.  IDE support should consider settings.xml so I have a lot of profiles in my settings.xml.  That way, I can run via CLI or within the IDE.
Let me know if that doesn't answer your question(s).

chrisknoll:
I think it does. I'll take a look at the sample profiles taht are in the POM and try to get this branch running on my local env where I can see the tables get created. I'd like to move on to getting this pull integrated because persistence tables is a high priority need.

I'd also like to get a cohort generation job running too.

-Chris

PR #17, opened 2015-02-25
Cohort Definition Persistence Implementation

chrisknoll:
This pull request will be used to coordinate changes related to cohort definition persistence.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .project
 /nb-configuration.xml
 /nbactions.xml
+/nbactions-webapi-mssql.xml
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 .project
 /nb-configuration.xml
 /nbactions.xml
+/nbactions-webapi-local.xml
 /nbactions-webapi-mssql.xml
--- a/nb-licenseheader.txt
+++ b/nb-licenseheader.txt
@@ -1,0 +1,19 @@
+<#if licenseFirst??>
+${licenseFirst}
+</#if>
+${licensePrefix}Copyright ${date?date?string("yyyy")} ${organization!user}.
+${licensePrefix?replace(" +$", "", "r")}
+${licensePrefix}Licensed under the Apache License, Version 2.0 (the "License");
+${licensePrefix}you may not use this file except in compliance with the License.
+${licensePrefix}You may obtain a copy of the License at
+${licensePrefix?replace(" +$", "", "r")}
+${licensePrefix}     http://www.apache.org/licenses/LICENSE-2.0
+${licensePrefix?replace(" +$", "", "r")}
+${licensePrefix}Unless required by applicable law or agreed to in writing, software
+${licensePrefix}distributed under the License is distributed on an "AS IS" BASIS,
+${licensePrefix}WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+${licensePrefix}See the License for the specific language governing permissions and
+${licensePrefix}limitations under the License.
+<#if licenseLast??>
+${licenseLast}
+</#if>
--- a/pom.xml
+++ b/pom.xml
@@ -361,6 +361,14 @@
               <flyway.datasource.password>FLYWAY_PASS</flyway.datasource.password>
               <flyway.locations>classpath:db/migration/sqlserver</flyway.locations>
             </properties>
+            <dependencies>
+              <dependency>
+                <groupId>com.microsoft.sqlserver</groupId>
+                <artifactId>sqljdbc4</artifactId>
+                <version>4.0</version>
+                <type>jar</type>
+              </dependency>
+            </dependencies>	 
           </profile>                  
 	</profiles>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,7 @@
 				<version>1.5.1</version>
 				<executions>
 					<execution>
+						<phase>package</phase>
 						<goals>
 							<goal>restdoc</goal>
 						</goals>
@@ -368,7 +369,7 @@
                 <version>4.0</version>
                 <type>jar</type>
               </dependency>
-            </dependencies>	 
+            </dependencies>	            
           </profile>                  
 	</profiles>
 </project>
--- a/src/main/java/org/ohdsi/webapi/Filter.java
+++ b/src/main/java/org/ohdsi/webapi/Filter.java
@@ -18,7 +18,7 @@ public class Filter implements ContainerResponseFilter {
         cres.getHeaders().add("Access-Control-Allow-Origin", "*");
         cres.getHeaders().add("Access-Control-Allow-Headers", "origin, content-type, accept, authorization");
         cres.getHeaders().add("Access-Control-Allow-Credentials", "true");
-        cres.getHeaders().add("Access-Control-Allow-Methods", "GET, POST");
+        cres.getHeaders().add("Access-Control-Allow-Methods", "GET, POST, PUT");
         cres.getHeaders().add("Access-Control-Max-Age", "1209600");
     }
 }
--- a/src/main/java/org/ohdsi/webapi/cohortanalysis/CohortSummary.java
+++ b/src/main/java/org/ohdsi/webapi/cohortanalysis/CohortSummary.java
@@ -2,7 +2,7 @@ package org.ohdsi.webapi.cohortanalysis;
 
 import java.util.List;
 
-import org.ohdsi.webapi.model.CohortDefinition;
+import org.ohdsi.webapi.cohortdefinition.CohortDefinition;
 
 public class CohortSummary {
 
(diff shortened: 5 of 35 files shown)


chrisknoll:
The first thing I'd like to work out is the CohortDefinition JPA repository:  We already have a CohortDefinition class out there, I do not know if Heracles depends on it or not.  There are some fields in there that I don't quite understand (cohortDefintionTypeConceptID..i'm not sure what this would be set to).  We have a cohort definition row mapper defined, but not sure if this would be needed in the case of JPA.  Also, there's a getCohortDefinitions.sql templateSQL out there, but i don't think we need to use template sql if JPA handles our CRUD (and query) operations.

I am hesitant to change/remove any existing code that others have produced that may introduce breaking changes.  I cam introduce new (possibly redundant classes) but I'd like some feedback about how this code should be structured.

-Chris

cahilton:
Those were the methods we created just to get going on Heracles, pre-JPA.

I would introduce the new code for now, let me know. Then I can switch Heracles to use the JPA version. Then, we can remove the old.

I need the code for my Heracles autocompleter, essentially just a list of cohort_definition with id, name, description.

chrisknoll:
This first batch of changes contains DDL for oracle and Postgres, but I'm unable to test them.  If someone with access to those platforms could checkout this branch and see if flyway can take it, that'd be helpful.

With the changes, I've had to change some of the signatures of cohort definition related references (cohort analysis for example)

cahilton:
I can work on oracle testing tomorrow.

chrisknoll:
Thank you, Charity.  If you're worried about applying migrations off of a branch and then switching back to some other branch (where the migrations do not exist) there will not be an issue except flyway warning that the schema version is newer thant he migration scripts you have. If you want to rid yourself of the warning, just delete the migration version from schema_version table.

cahilton:
Ah - good point. Thanks.

cahilton:
The other thing we could do for oracle testing is to setup a VirtualBox image with postgres, oracle xe, sql server express for testing. Of course, we couldn't really distribute it as open source, but we could share how we created it, and maybe share internally.

BTW - not suggesting you do this, just an idea that might be handy for us. We've done this when we needed a new database for integration testing, or needed to make ddl changes experimentally, that the DBAs just won't let us do.

For a bonus, we put our IDE on there, and have a sandbox environment. Anyway, just an idea.

chrisknoll:
It's a good idea, I have virtual box on my local box, but my poor machine can barely handle a sql server mgmt studio + netbeans + email + browesr (for postman) set up...Buuuuut getting it set up on one of our servers....dooable.  And we have profiles in our own envrionment testing it would be just a matter of swtiching a profile (and hence jdbc urls).  But..godz...learning how to set up 2 more database platforms...my head...

fdefalco:
We could also work with Lee to leverage the amazon environment.  He already has a postgresql database that is publicly available.  Setting up the SQL and Oracle shouldn't be too much of a challenge.  Sharing the databases across development environments will likely prove challenging with ongoing development branches but we'll need to deal with those types of things as different community members adopt the platform.  We should get a very good look at how gracefully the software can behave.

leeevans:
An OHDSI  virtual box developer environment is a potential option.  We could also leverage the OHDSI AWS environment as Frank mentioned.  Either way, someone will need to maintain the environment.  I'm currently keeping the AWS Postgresql database up to date as we add new Spring capabilities.

I already have Postgresql, Oracle XE and Sql Server running on the AWS OHDSI Jenkins CI server (it's a Windows machine).  I needed them for the sqlrender web app.    However currently the CI server WebAPI build is using a different Postgresql database on the LAERTES AWS server.

It is certainly possible to setup 3 WebAPI CI Server build jobs - one per DBMS.  It would just take some time to get the V5 "unrestricted" vocabulary setup in all three DBMSs and create the ohdsi schema and setup the three CI server build jobs. I've held off on doing that so far because as I mentioned on the call yesterday we are individually working and building with all three DBMSs anyway - but it's certainly an option.

Sharing access to those three DBMSs across all developers for builds is more challenging.  We would need to consider how we would support multiple users in each database (something we should think about anyway) to avoid one developer build impacting another developer build and keep/refresh test data separately for multiple developers.

chrisknoll:
Lee, could you try to pull the sql out of db/migration/postgresql/V1.0.0.3__cohort_definition_persistence and try a 'dry run' on a postgres database to see if there are any errors?  I think this pull is ready to go (save and load is implemented from a cohort definition perspective) unless there's other issues with other client apps that I am not aware of.

leeevans:
Hi Chris,

I just ran the postgresql migration sql successfully and manually dropped the sequence and tables it created.

It looks good so I think you can go ahead and merge.

chrisknoll:
Ok, i'll wait to hear from Charity about her experience with the oracle side.

chrisknoll:
Added some updates to deal with lazy loading one to one mappings (CohortDefinitionDetails is a one to one mapped to CohortDefinition to extract the LOB column for Expression into it's own entity).  Still having problems make it work the way I'd like, but it does work.

I'd like to request some assistance writing a unit test for this.  I'm not sure what type of annotations I'd need to add to a test class in order for it to inject the spring JPA elements into the class.  I'd simply like to write a save new -update existing -delete existing test that will test the @GeneratedId behavior across all platforms.

cahilton:
There were some problems.

`FK_cohort_definition_details_cohort_definition` was too long.

Error at Command Line : 5 Column : 13
Error report -
SQL Error: ORA-00972: identifier is too long
00972. 00000 -  "identifier is too long"
*Cause:    An identifier with more than 30 characters was specified.
*Action:   Specify at most 30 characters.

Also: 
`expression [varchar](max) NOT NULL` is not a valid type

You can do:
`expression VARCHAR2(4000) NOT NULL`
or
`expression CLOB NOT NULL`

depending on whether you need the space or the text manipulation utility, but there's no varchar max in Oracle.

chrisknoll:
Would you mind pushing in a change that would make it deploy properly?  I'd say that CLOB makes more sense than varchar(4k) only because these will be large text blobs.  But not sure if there's special notation you need to apply to the entity.  Do you want to give it a shot?

chrisknoll:
Actually, Charity, hold off on that. I have a different approach that will split the expression across rows.  I'll have an updated branch pull in a few hours.  The Columns will be deifined as varchar2(4k), and I'll set the split size as 4k (not sure if I should expose this as a property in the app, might make sense to do so).

cahilton:
Okay, will do.

Sent from my iPhone


On Mar 1, 2015, at 2:27 AM, Chris Knoll <notifications@github.com<mailto:notifications@github.com>> wrote:


Actually, Charity, hold off on that. I have a different approach that will split the expression across rows. I'll have an updated branch pull in a few hours. The Columns will be deifined as varchar2(4k), and I'll set the split size as 4k (not sure if I should expose this as a property in the app, might make sense to do so).

chrisknoll:
Ok, new commits pushed: major change is that i'm returning specific DTOs (not repository entities) from the service. While working out save/load I noticed that if we ever return entities that have lazy load fields that we do not actually pull in for a given operation, those lazy fields will be left with a proxy, which can't be serialized down to json.  So, I think the better strategy is to create 'data view' dtos that load from the repository all elements that a particular operation needs to work with. I've defined 2: CohortDefinitionListItem and CohortDefinitonDTO: list item is used to get list of all definitions (where returning a CLOB per item would be expensive for the dlient) and the DTO does include the CLOB data (but the DTO is returned int he cases where we load a specific definition).

I also shortened the foreign key references in the oracle migration script. You should be able to apply it without errors.

-Chris

chrisknoll:
Charity, in the latest commits, i assigned the column type of expression to CLOB.  I am not sure if I need to specify @LOB on the expression data because I don't think any special 'large object' handling needs to be performed for mssql and postgres (they are both 'varchar-ish' types).  I'm not sure if the more recent versions of oracle require you to specify @LOB on string fields that are stored on CLOB fields.  Could you run a test on your envirionment hosting this branch, and use postman to post the following:
```json
    {
      "name": "Test CLOB INSERT",
      "description" : "My Description",
      "expressionType" : "SIMPLE_EXPRESSION",
      "expression" : "{ PrimaryCriteria: {} }"
    }
````

PUT'ing that to webapi/cohortdefinition should trigger a create.  If it saves then I guess we don't need to make a special annotations for @LOB.  However, it may require that we do a test for larger text payloads, but I'm assuming saving to a clob column behaves the same in the small text length case as well as large text length.

cahilton:
Chris, I have a question. Should we just assume that our current cohort_definition tables are not there? 

I have some data in there, but I can keep it, and drop them.

chrisknoll:
Ok, Charity. You should be able to pull this branch and run the PUT. It will save using the @Lob annotation.  This works on SQL2012.

cahilton:
Also, another question, maybe will get addressed today. But who is going to create the OHDSI.COHORT? I can make that part of my heraclas-viz branch, unless you already have plans for it.

We don't have any WebAPI hitting it, just needs to be there for Heracles analyses.

cahilton:
Happy to report that this works fine in Oracle-land now.

cahilton:
How far are we away from merging this in to master? I can merge it into my branch too, just want to make sure there aren't too many more changes, so I can move forward using cohort_definition for Heracles.

cahilton:
Thanks.

PR #18, opened 2015-03-03
Initial .gitattributes file. Normalize eol. Fix Flyway checksum validation error

alexfranken:
.gitattributes should help ensure that eol characters are normalized when writing to the repository.  This should prevent messy diffs (e.g. where it looks like the whole file changed) & Flyway checksum validation errors, regardless of the value of 'git config --global core.autocrlf'.

For Flyway, the .gitattributes basically says always convert *.sql to 'lf' when writing to the object database as well as the working directory.
Below are steps to take to “refresh” existing cloned repo, after modifying .gitattributes
1.)	mvn clean (remove target dir)
2.)	git rm –cached –r *
3.)	git reset --hard

To confirm a file had LF, I used *nix `file src/main/resources/db/migration/oracle/V1.0.0.1__schema_create_spring_batch.sql`  The output should mention 'ascii text' (not 'ascii text with CRLF').

For our environment, where the schema was created via Flyway on a Windows machine (with core.autocrlf=true), I then dropped all Flyway managed objects and re-ran migrations.

More on the Flyway issue: https://github.com/flyway/flyway/issues/253

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,15 @@
+* text=auto
+*.sql text eol=lf
+*.xml text
+*.groovy text
+*.java text
+*.html text
+*.css text
+*.js text
+*.vm text
+*.properties text
+
+*.png binary
+*.jpeg binary
+*.jpg binary
+*.gif binary


cahilton:
Following your instructions to reset worked for me.

chrisknoll:
One thing we may need to be aware of is that if you are working on a migration script and you have fly away deploy it before a commit (and even after a commit, you may get a message that line endings will be modified for commit but left native on your local filesystem) you may run into checksum errors when someone else pulls your changes into their own local environment. If your editor supports specifying line endings, set it to LF.  I found this plugin for netbeans that you can install via the IDE:
http://plugins.netbeans.org/plugin/36810/show-and-change-line-endings

I noticed that before doing the hard reset per above instructions, the files opened up and they said they were 'crlf'.  After reset, the line endings were detected 'LF'. After launching the app, i got a checksum validation error.  I repaired it using the Flyway CLI tool: flyway repair.

-Chris

alexfranken:
@chrisknoll  The line `*.sql eol=lf` should use LF in both the repository db as well as the working directory, so a Windows user would have previously seen these files as CRLF (due to default core.autocrlf=true setting).  Like you saw, the reset will change the working directory, so these change to LF.  We opted to drop and recreate versus "repair" but since there's little we can do to prevent the case you mention (where someone creates a new .sql file in Windows - resulting in CRLF), I imagine the "repair" functionality will be used more than should be necessary.  Maybe we set a convention/guideline of "committing" to your branch prior to running app (i.e. initiating flyway)...  What a pain.

chrisknoll:
Yeah, it's just something to look out for.  I think that this is going to be resolved in the next version of Flyway.  I only think this is an issue for editors that default to CRLF, and even netbeans seem to open up new files in the 'last line ending mode' so I think we'll be fine in the short term.

PR #19, opened 2015-03-03
Heracles flyway

cahilton:
I like to go ahead and merge the Heracles flyway configs just so I don't conflict with anyone else. As long as these work. Can you guys try in SQL Server and Postgres?

Thanks.


chrisknoll:
This script works on sql server.  Assuming that the last modified columns should be 'datetime' and not 'timestamps'. Timestamp in sql is not exactly a date time, you can't default a timestamp to 'getdate()'.

chrisknoll:
I'm ready to merge this request if there are no issues.

cahilton:
Works for me, if it works on Postgres.

PR #21, opened 2015-03-05
Cohort generate job

chrisknoll:
This pull request introduces cohort generation capability.  This has been tested on MS Sql only.  Our postgres database does not have CDM tables with data, and I do not have access to an Oracle instance. Those of us who have those environments, please check out this branch and let me know of any issues.

Main points:
New Task: GenerateCohortTask - encapsulates the params to the GenerateCohortTasklet
New Tasklet: GenerateCohortTasklet - takes the cohort definition as a parameter, and converts it to SQL. Then executes SQL against database.
New application.properties field: cohort.targetTable. Set this to the table the GenerateCohortTasklet will write to.  This is specified in the default pom.xml as dbo.COHORT, but we can make this a placeholder value with instructions to override in settings.xml.
New CohrotDefinitionService endpoint: /{id}/generate.  GET request which triggers the cohort generation.

To Do:
I think we need a status table for cohort generation to track things like execution duration, cohort counts and error messages. I know the job subsystem has a record of executions but I was thinking of tracking more task-specific status.

Need testing on other platforms.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,9 @@
 		<source.name>CDM_NAME</source.name>
 		<cdm.version>5</cdm.version>
 
+                <!-- cohort generation propertise -->
+                <cohort.targetTable>dbo.COHORT</cohort.targetTable>
+
 		<spring.batch.repository.tableprefix>BATCH_</spring.batch.repository.tableprefix>
 		<spring.batch.repository.isolationLevelForCreate>ISOLATION_READ_COMMITTED</spring.batch.repository.isolationLevelForCreate>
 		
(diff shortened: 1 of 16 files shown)


chrisknoll:
latest push now provides the ability to check the status of a cohort generation.  This is very simple at this point (when it was executed, duration in milliseconds, status (running, pending, complete) and if it is valid (exceptions thrown during execution).

I haven't tested this on postgres yet (will try tomorrow) but if someone on oracle land could deploy this migration script (1.0.0.3.1) and let me know of any issues, I'd appreciate it.

chrisknoll:
This was tested and runs on postgres.  If I don't hear any problems on the oracle side, I'll merge this to master later today.

cahilton:
I just checked it out. I'll let you know.

cahilton:
Starts up fine, and flyway scripts ran without issue.

chrisknoll:
I've made post-merge updates to this to address 2 items that came up as I was doing additional circe work. This was checked in to master, please pull from master to see the updates. i will be pushing CIRCE with a 'generate' button shortly.

PR #23, opened 2015-03-13
Laertes REST calls added to the WebAPI

rkboyce:
Dear Team,

I have created a new 'evidence' service implemented the following calls:

1) /evidence/drughoi/{key}  # key: <drug>-<hoi> using concept ids. 
Ex:http://localhost:8080/WebAPI/evidence/drughoi/1000640-137682

2) /evidence/drug/{id}
Ex: http://localhost:8080/WebAPI/evidence/drug/1000640

3) /evidence/hoi/{id}
Ex:http://localhost:8080/WebAPI/evidence/hoi/320073

As of right now, these all work to return counts and also return functioning URLs that a web client can use for the "drill down" use case.

Please let me know if you have any questions or comments.

-R


.gitignore
```diff
@@ -1,3 +1,7 @@
+*application.properties
+sandbox/
+pom.xml
+pom.xml*
```
fdefalco: pom.xml shouldn't be ignored, any particular reason for this?

src/main/resources/application.properties
```diff
@@ -17,7 +17,7 @@ cdm.version=${cdm.version}
 
 #Flyway database change management.
 #DataSource for Change Managment / Migration
-flyway.enabled=true
+flyway.enabled=false
```
fdefalco: I'm not familiar with the flyway implementation, but is this something you were looking to disable for Laertes only?  @chrisknoll or @alexfranken would have to determine how this should be handled.  

Along the same lines however, it looks like your service relies upon your evidence tables existing in the CDM schema, which I know they aren't by default.  I would imagine they would be more appropriately located in the OHDSI schema.

src/main/java/org/ohdsi/webapi/service/EvidenceService.java
```diff
@@ -0,0 +1,221 @@
...
+     */
+    @GET
+    @Path("drughoi/{key}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Collection<DrugHoiEvidence> getDrugHoiEvidence(@PathParam("key") final String key) {
+	String[] par = key.split("-");
+	String drug_id = par[0];
+	String hoi_id = par[1];
+        String sql_statement = ResourceHelper.GetResourceAsString("/resources/evidence/sql/getDrugHoiEvidence.sql");
+        sql_statement = SqlRender.renderSql(sql_statement, new String[] { "drug_id", "hoi_id",  "CDM_schema" },
```
fdefalco: You are passing CDM_schema in here, but you don't reference the @CDM_schema parameter anywhere in your SQL, so it won't be replaced.  I mentioned it elsewhere, but I think we would want the evidence tables created in the OHDSI schema and potentially created by flyway if you want to leverage it for versioning of feature sets from the evidence base.  You can take a look at the hermes queries for examples of how to replace the schema parameters in the SQL.

.gitignore
rkboyce: Ok. Fixed now.

src/main/java/org/ohdsi/webapi/service/EvidenceService.java
rkboyce: I was aware of the situation but left it so as a placeholder for when I know what schema parameter to pass. Looking at PGAdmin the current schema is "public". How hard would it be to migrate? Can we leave it as it is until after AMIA?

src/main/resources/application.properties
rkboyce: Flyway was presenting an issue with starting the server because I don't have a username or password to use for connecting. I have no opinion about using it for Laertes right now except that I would have to learn how. Please advise.

alexfranken: I think we want the default to be true. If someone wants to disable flyway then that is OK, just probably not committed.

chrisknoll: Yes, flyway should be left enabled. You may be able to override this to your local settings config to false, but it should be not comitted as 'false' in the repo.

-Chris

rkboyce:
Lee's specification for Flyway user and pword worked so I turned back on flyway in that source application.properties file. Also, I removed the CDM_schema filter from EvidenceService.java query build code and wrote the following comment write before each call to SqlRender.renderSql  we can address after the F2F and AMIA next week:

	// TODO: modify this to use a parameter for the schema where
	// Laertes' tables reside. For now, we assume they are in the
	// 'public' schema. If they reside in a new schema (i.e., not
	// ohdsi or cdm) then a new property datasource.laertes.schema
	// property will need to be created in application.properties,
	// filtered in pom.xml, read in AbstractDaoService.java, and
	// passed to renderSql here.

PR #24, opened 2015-03-13
Clearer distinction between DataSource handling (CDM vs OHDSI)

alexfranken:
@chrisknoll & @fdefalco , I realized that the datasource plumbing wasn't really in line with what I believe to be our intent.  That is, the CDM schema should be read-only and that we want the OHDSI/RESULTS schema  to be transactional.  
Flyway config was right but the Jpa & Job config was really using the default datasource, which was the "CDM DataSource".
So, I tweaked the configuration to make a cleaner distinction between CDM & OHDSI DataSources (including JdbcTemplates).  So, within WebAPI, the idea is that when you need transaction support, you should work with the ohdsiDataSource/ohdsiJdbcTemplate/transactionTemplate/crudrepository (any of) beans.  For example, the /vocabulary/vocabularies endpoint would use the CDM JdbcTemplate (getJdbcTemplate in AbstractDAOService), and the tasklets, etc., that write data, would use the OHDSI JdbcTemplate (getOhdsiJdbcTemplate in AbstractDAOService).
I think this makes sense from the WebAPI perspective and supports variable configurations.  I don't think it was an issue for any of us because we were all configuring WebAPI with DB accounts that were associated with the OHDSI schema, or had a "CDM" account that could write to the OHDSI schema...

To take it a step further, we could rename the "flyway.datasource" properties to represent the ohdsi.datasource but I didn't go there due to the impact to others and their settings.xml.

Let me know any thoughts you all have.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
+++ b/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
@@ -4,11 +4,13 @@ import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
 import org.springframework.core.env.Environment;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.datasource.DriverManagerDataSource;
 import org.springframework.orm.jpa.JpaTransactionManager;
 import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
@@ -31,7 +33,7 @@ public class DataAccessConfig {
     
     @Bean
     @Primary
-    public DataSource primaryDataSource() {
+    public DataSource cdmDataSource() {
         
         String driver = this.env.getRequiredProperty("datasource.driverClassName");
         String url = this.env.getRequiredProperty("datasource.url");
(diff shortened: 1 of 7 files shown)


chrisknoll:
Actually, the cdm schema being read only is not defined as being read only, just happens to be so in certain contexts.  But regardless, i think we only need 1 datasource: the connection the ohdsi schema.  (I should say +1 for the flyway datasource which has special configuration, but isn't used in normal app operations) At least how I'm assuming things would work is that i connect to the ohdsi datasource and access CDM tables through some alias (such as a schema prefix).  I don't open 2 connections during processing: I execute a query that queries directly aginast the cdm data but writes directly to an ohdsi schema in the same batch execution. this should be the same way for your herecles results: you query cdm, write to herecles, but use a single connection (the ohdsi one).

alexfranken:
That is much cleaner.  Maybe I was making up hypothetical cases from our discussions :)

chrisknoll:
Hey, Alex, maybe I was misunderstanding you.  ohdsi and cdms are defintely distinct/separate things.  I just see us doing querys that will bridge those things.  Like getting data from the cdm to do your analysis into heracles. But there could be other cases...and it might be advantageous to have a distinct line between a cdm datasouce and a ohdsi one...I can think of one such case in our own environment where we have a cdm on PDW and our ohdsi schema might be on a sql2014 instance.  In this case, I wouldn't be able to join across the different instances, but then i'd have to resort to pulling data from a query from the CDM into memory and dump it back into Sql2014 for processing. That sounds possible but posslby a bit I/O intensive to get the data. And then to take the results of taht processing and join it back into a query against the PDW server, i don't know how that would be possible.

What circumstances were you thinking about where we'd have distinct datasources?  Maybe something was discussed but I'm just missing it.

-Chris

PR #26, opened 2015-03-19
Cohort importer

cahilton:
This is a feature done by our team for @jduke99 by @cbeesley. It allows importing and exporting from our NLP app into and from the CDM. The idea is that you already have a list of cohort entities exposed in the source application that are the same identifiers as the CDM.


src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -186,9 +186,12 @@ public CohortDefinitionDTO createCohortDefinition(CohortDefinitionDTO def) {
             .setDescription(def.description)
             .setCreatedBy("system")
             .setCreatedDate(currentTime)
-            .setExpressionType(def.expressionType);
+            .setExpressionType(def.expressionType).setId(0);
```
chrisknoll: I don't think you should set Id here. This is a new cohort definition. It will be treated as an insert if id is NULL.

chrisknoll:
Other than my comment about the ID, I'm not sure what the use case is for this pull request: is this to copy data from the OHDSI result schema into the CDM schema cohort?  If so, there's going to be a challenge to have a conflict with cohort_definiton_id generated out of the ohdsi_scheama.cohortDefiniton table with any cohort that was created ad-hoc that inserted their own cohort_definitoin IDs into the cdm.  We have examples of people doing this with making 'custom observation periods' by inserting data into the cdm_schema.cohort table.

cahilton:
Maybe @jduke99 can describe it better, but basically we have an NLP tool that can generate cohorts outside of Circe. It already has the ids and notion of cohorts, but it's outside of the CDM, the idea is to take this cohort and link them to the CDM

chrisknoll:
Ok, the Issue I see is that we're tagging this as a 'simple cohort expression' and so circe will think that this is an actual cohort definiton, it will show up in circe to edit, and someone could go in and change the definition and run it (thus blowing away your cohort import).

Would you consider defining a new type of cohort expresison type (we currently have SIMPLE_EXPRESSION and CUSTOM_SQL) for purposes of defining an external definition:  we can define a new JSON object that serializes down to the expression but this JSON object can just have a note: field on it that you can put in a description, or a description, or any other type of metadata info you want to save into an expression so that you have some information about where the cohort came from.

Then we can have circe only look for cohort definitoin types of 'SIMPLE_EXPRESSION' (because that's the only type of cohort defintiion it can edit), we can add some new cohortDeifinitionRepository CRUD methods to filter by expression type...and you'll be able to have a new type of cohort defininition that comes from an external source.

chrisknoll:
Additional thought: I mentioned that we could define a cohort as hard coding up specfic PERSON_ID, start date, end date combos, and he liked that idea (that you could specify the person ids in a cohort right in an expression) but he said he'd still want an import type because some of the cohorts are very large (and would be not great to store in JSON serialized form).

So i guess i'm proposing 2 additional cohort expresison types: EXTERNAL_SOURCED and VERBATUM (external source just has descritpion of the cohort, verbatum has the physcical person ids in it suited to things < 1000 members)

-Chris

cahilton:
I think that will probably work. Knowing most of these will probably fit into the EXTERNAL_SOURCED category.

chrisknoll:
Ok, we can either introduce it in this branch, or i can make a master change that you can rebase on.  I'm sorta leaning towards introducing it here since it's tied to this feature (and I'd like to demonstrate to people how they might want to extend the CIRCE model...)

cahilton:
Fine with it being in this branch.

chrisknoll:
Charity, I'm looking at the changed files, and I realize that I'm only seeing half the picture: I see how you modified the CohortDefService to default to an empty expression (of type SIMPLE_EXPRESSION) but i'm not seeing what is being populated on the client side.  So, maybe the way forward is to offer suggestions about what I think you should account for and then you can see how that folds into your side:

The first thing I'd consider changing is where you default the definition expression to what looks like an empty simple expression. I don't know what value you are passing for DefinitionType but the way you are defaulting it is that it will look like a simple expression criteria, but the definition you are providing is coming from a cohort list.  So, I'd remove the 'default if null' so that we actually throw an error if they do not pass a value, and decide on a JSON string that can represent the cohort definition expression that you can pass into the PUT call.  I think I heard you had 2 modes: 1 was passing in a list of ids, and second is just an external load was performed on the cohort table that you want to represent in the cohort definition table.

For the first case: I'd add a new ENUM value to ExpressionType.java called PERSON_LIST, and the json object can be something like { personId: [], startDate: [], endDate: []}.  Also note this string is going into a string field of the cohort definition JSON, so you will stringify that value so that it looks like this: "{ \"personId\":....

The second case, is another ENUM value of EXTERNAL_SOURCE which can have a json object that looks like this : { description: "{some text here"}}.

In both cases, you set the cohort definition expression to one of these two object structures, and you can remove the null check from your commits.  Also call setExpressionType(ExpressiontType.EXTERNAL_SOURCE) (or the other enum) based on what type of cohort you define.

If you do this, then I can change CIRCE to only view cohort definitions that have a type SIMPLE_EXPRESSION.  But I can do this independently of your branch here, I'm just saying this is how I'd exclude your custom cohorts from being visible in the Circe editor.

-Chris

cahilton:
Thanks for following up on this, Chris. For us right now, the most common option will EXTERNAL_SOURCED, because generally they'll be hundreds or thousands of patients. So, that is my priority, but if you want to implement both, that works for me.

I'm happy with populating a basic JSON as you described.

For now, what we have on our side is pretty minimalistic because there's no way right now they can interact with Circe, we just need to a way to insert a cohort_definition to be selectable from Heracles.

For what that's worth, here's an example of what we're doing in our NLP app (and I can easily switch it away from using SIMPLE_EXPRESSION to EXTERNAL_SOURCED).

### Controller
```
	    	CohortDefinition def = new CohortDefinition();
	    	def.setName(cohortName);
	    	def.setDescription(cohortDescription);
	    	def.setExpressionType("SIMPLE_EXPRESSION");
	    	def = cohortDao.saveToCDMNewCohortDefintition(def);
```

### Service call to WebAPI
```
	/**
	 * Does a put to a web service to create a new cohort
	 * 
	 * @param def
	 * 
	 * @return The newly create cohort def
	 */
	public CohortDefinition saveToCDMNewCohortDefintition(CohortDefinition def){
		RestTemplate restTemplate = new RestTemplate();
		HttpHeaders headers = new HttpHeaders();
		headers.setContentType(MediaType.APPLICATION_JSON);
		final ObjectMapper mapper = new ObjectMapper();
		String json = "";
		try {
			json = mapper.writeValueAsString(def);
		} catch (JsonGenerationException e) {
			e.printStackTrace();
		} catch (JsonMappingException e) {
			e.printStackTrace();
		} catch (IOException e) {
			e.printStackTrace();
		}
		if(Strings.isNullOrEmpty(json)){
			return def;
		}
		

		HttpEntity<String> entity = new HttpEntity<String>(json,headers);
		//restTemplate.put(host + "cohortdefinition/", entity);
		ResponseEntity<String> responseEntity = 
	            restTemplate.exchange(host + "cohortdefinition/", HttpMethod.PUT, entity, String.class);
		
		try {
			CohortDefinition newDef = mapper.readValue(Strings.emptyToNull(responseEntity.getBody()), CohortDefinition.class);
			return newDef;
		} catch (JsonParseException e) {
			e.printStackTrace();
		} catch (JsonMappingException e) {
			e.printStackTrace();
		} catch (IOException e) {
			e.printStackTrace();
		}
		return def;
	}
```

chrisknoll:
Ok, I think the only thing you'd need to do to specify EXTERNAL_SOURCED is to add an enum to ExpressionType in the cohortdefinition package.  You can make up any json object structure you want since it just stores as a string in the definitionDetails table.

The enum is required because it is going to resolve that string into the appropriate enum and i it doesn't exist, i believe you'll get a serialization error.

cahilton:
Can we merge this now?

All the changes are isolated to the cohort importer, and I added the EXTERNAL_SOURCED to the enum.

chrisknoll:
Yes, my apologies for the slow turnaround.

cahilton:
No problem-thanks!

Sent from my iPhone


On Apr 10, 2015, at 11:48 PM, Chris Knoll <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #26<https://github.com/OHDSI/WebAPI/pull/26>.

PR #28, opened 2015-03-20
bringing in drug rollup by ingredient, clinical drug, and branded drug

rkboyce:
Hi, I added three new calls so that we can get data from the drug rollup table (laertes_summary) by ingredient, clinical drug, and branded drug:

By ingredient: http://localhost:8080/WebAPI/evidence/drugrollup/ingredient/1000632

By clinical drug: http://localhost:8080/WebAPI/evidence/drugrollup/clinicaldrug/19074181

By branded drug: http://localhost:8080/WebAPI/evidence/drugrollup/brandeddrug/1000640

Please let me know if you have any questions. The Makefile is just something that helps me with the build and teset process.


rkboyce:
Hi, I think I fixed the issues that were preventing Laertes calls from working on the public WebAPI instance. I now make copy of that evidence tables at load time from the public to the unrestricted schema. In the evidence servioce within WebAPI all calls pass "@CDM_schema" which is configured to hit the unrestricted database.  Testing my local instance with user 'webapi' seemed to work fine.

chrisknoll:
Rich,
  What's the plan about how these LARATES tables being created in the cdm schema? Normally if we are introducing new tables, they get Flyway migration scripts (so that they are created during startup if needed) but these tables also get put into the OHDSI schema. In your case the're being put into the CDM schema (which in some environments is read only).  So, where do they come from?

rkboyce:
Hi Chris,

I don't claim this is the best solution but for now I am just copying the tables and data from the public schema (where Laertes data is initially loaded) to the unrestricted schema so that the user 'webapi' should be able to read them. The script that does this also creates the initial table schemas: https://github.com/OHDSI/KnowledgeBase/blob/master/LAERTES/Schema/postgresql/ohdsiTest3.psql

I don't think that being read-only in some environs is a problem. Also, Lee didn't think that using Flyway for these tables was a good solution (see below). What are your thoughts?
"My opinion is that the evidence base tables are a different animal that falls outside the scope of the CDM and the ohdsi schemas. They are not part of the official CDM as far as I'm aware (but could be added) and they are not results tables because they are pre-populated by the LAERTES ETL process (not written to by any OHDSI app calling the WebAPI). Flyway is only currently used for creating/maintaining ohdsi (results) schema tables.

The simplest option would be to create the LAERTES tables (manually) in the CDM ("unrestricted") schema.  Another option would be to create a new schema (maybe called 'laertes' or 'kb'?) for the LAERTES tables and add an additional WebAPI datasource.ohdsi.schema property for that new schema. We would need to grant the webapi user read-only access on the new laertes/kb schema and I would have to add the new property on the Jenkins server and the server where WebAPI is deployed (in tomcat)."

chrisknoll:
Ok, we can merge this now so that the public api has access to it, but here are the concerns:

1. Anyone else who pulls from master and wants to use the LARATES api will have to take additional steps to get the app working, and the steps are not clearly understood.  
2. The course we take now may have negative impact on how this is going to be handled in the future later. If we move towards a migration tool like flyway, there could be already deployed LARATES infrastructure that may get in the way.  Are we ready to face the pain later by taking the easy path now?

If we are going to manage the LARATES tables as something separate from CDM or OHDSI, then it seems that we would be introducing a new schema as Lee describes.  This is fine, I think flyway can manage this.  What i'd suggest the path forward is that after this is merged, you create a new branch, have your queries reference this new schema, and create the flyway migration scripts that will create the tables.  The population of these tables from the LARATES etl process can be left as a separate step and a manual process, but the problem remains that these tables need to be created in order for the WebAPI to function.

If you are ok with the above, I can merge this.

leeevans:
Keep in mind that the "public" schema is the default schema that always exists in Postgresql.  There is no default "public" schema in Oracle or Sql Server so it would need to be manually created or created by Flyway.

pbr6cornell:
My two cents:   I think the LAERTES tables belong in the OHDSI schema, and
would recommend doing that rather than creating yet another schema.  Data
can be loaded into tables within the OHDSI schema, there's nothing
forbidden about that.  Ultimately, we'll want to enable WebAPI to call a
community server that hosts LAERTES, but we'll also want to support users
who want to download the LAERTES schema and run it within their local
environment.   This is directly analogous to how HERMES can point to a
public install of API on OHDSI server, but can also point locally to a
install of the vocab database.

I defer to Chris on how to integrate flyway in the process, but it seems
we're only talking about the DDLs for a handful of tables, so shouldn't be
a big deal to do that using flyway, just like we're doing with all other
apps.

rkboyce:
Ok. I am flexible. At the end of the day, I just need a pathway that allows us to easily update the table schemas and reload data when we have updates (which right now can be fairly frequent). Please just tell me what to do once decided. I plan to do a full update next week when the new vocab release is loaded so its pretty good timing if I have to learn flyway. thanks! -R

cgreich:
Rich et alii:

Let's also sit down and clean the tables up, before they get too blessed by applications and users. Want me to give it a shot and we have a session with interested folks?

rkboyce:
Sure. Please fork the code and make the changes you think best. Erica and myself plan to completely update the evidence base next week after the next release of the vocab is in. I can then take your changes as a pull request if you generate one, and migrate to flyway per Lee and Chris's instructions as to how and which schema.

We can also discuss plans/issues/questions at next week's KB meeting on Wed.

cgreich:
Where are the DDLs? In which folder?

rkboyce:
The folder is https://github.com/OHDSI/KnowledgeBase/tree/master/LAERTES/Schema/postgresql

The specific file is: ohdsiTest3.psql

chrisknoll:
Ok, a few comments here just so that we know what the target is:
Flyway can be configured to manage any schema, public for postgres, dbo for sqlserver, whatever for oracle...it's just the default schema that the flyway process attaches to when running migration scripts.

WebAPI has configuration settings to specify the schema that you can use in your SqlRender to replace {ohdsi_schema} or {cdm_schema} tokens in your scripts.  If you're just reading from larates tables then you should ahve an ohdsi_schema token as a prefix for all your tables, because it's a site-specific configuration setting that will specify what schema the tables are in.  That being said, you should be doing sqlTranslate() on all your custom sql queries you perform against your database.  Example:

```java
    sql_statement = SqlRender.renderSql(sql_statement, new String[] { "query", "CDM_schema", "filters" }, new String[] {
    search.query.toLowerCase(), getCdmSchema(), filters });
    sql_statement = SqlTranslate.translateSql(sql_statement, "sql server", getDialect());
```

Btw: getCdmSchema() is a base class function that if you extend AbstractDaoService, you'll have access to it, and that will return the value based on the property files.

An example SQL statement that has tokens in it looks like this:

```sql
select CONCEPT_ID, CONCEPT_NAME, ISNULL(STANDARD_CONCEPT,'N') STANDARD_CONCEPT, ISNULL(INVALID_REASON,'V') INVALID_REASON, CONCEPT_CODE, CONCEPT_CLASS_ID, DOMAIN_ID, VOCABULARY_ID
from @CDM_schema.CONCEPT
where CONCEPT_ID = @id
```

This example would use the @CDM_schema to replace, but you would put something like @OHDSI_schema in your own scripts and replace them using standard java string replacement or the sqlRender functions.

I'd take a look at VocabularyService and CohortDefinitionService if you are unclear as to how to load up sql queries, translate and execute them.

Finally, on the sql DDL you provided above: it looks like it's a mix of table creates and inserts.  I'd split that up so that all the tables are created up front (that Larates the web service requires) is created, and your population scripts, just assume that the tables you need are created.  Your stand-alone LARATES setup documentation can just say the same scripts that FlyWay will use to create the tables should be a pre-setup step that needs to be performed once before running the population scripts.  On that note: you shoudl have your data population scripts truncate before inserting (assume that you already have talbes there and they might have bad data in it).  Of course, disregard this if your strategy is to ammend data overtime and not refresh each run.

Let me know if you have any questions.

rkboyce:
Alright. Thanks for the instructions Chris. I plan a full update next week and will work on this. Am sure I will have questions with flyway simply because its new to me. We'll get through it though :)

chrisknoll:
Is this pull request obsolete since @leeevans  made the other pull request?

leeevans:
yes @chrisknoll. my pull request incorporated the code changes @rkboyce made in this earlier pull request

PR #29, opened 2015-03-23
Feasibility study

chrisknoll:
This pull requests adds the feasibility service that is used for CALYPSO.

Migration scripts have been tested on postgres and mssql.


chrisknoll:
I'd appreciate it if someone on the oracle side could pull this branch down locally and see if anything breaks.

New flyway migration script was added:
V1.0.0.5__feasability_tables.sql

PR #33, opened 2015-04-09
get circe working on oracle

cahilton:
@chrisknoll 

@jduke99 needed Circe to generate in our environment. To get that to work, I had to make these changes. I don't want to break anything, but doesn't work in Oracle without the target schema being used.

Thanks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -333,6 +333,7 @@
     <profile>
       <id>webapi-oracle</id>
       <properties>
+        <cohort.targetTable>OHDSI.COHORT</cohort.targetTable>
         <datasource.driverClassName>oracle.jdbc.OracleDriver</datasource.driverClassName>
         <datasource.url>jdbc:oracle:thin:@127.0.0.1:1521:cdm</datasource.url>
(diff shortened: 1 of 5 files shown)


chrisknoll:
Charity:
See this post about the new property that was added to pom.xml
http://forums.ohdsi.org/t/circe-generate/431/3?u=chris_knoll

cahilton:
ok - I can specify the pom property for now, but it might be cleaner eventually to just used ohdsi_schema + cohort, since  I think we've moved towards that with Heracles.

Trying session id now.

cahilton:
Also that new translate method requires the oracleTempTable schema (which for us is the OHDSI schema), but going to try null for now.

cahilton:
Okay, the changes I have in this branch now work.

So, I added the schema name + cohort name in the oracle profile, and I passed in the session id on the generate cohort tasklet

chrisknoll:
Perfect. I'll draft up a pull request to remove cohort.targetTable property from being required.  Great work.

PR #35, opened 2015-04-15
Heracles viz

cahilton:
I know this is a HUGE pull request, but it's all the base work for Heracles.  Knowing there may be some more queries we add in the future, I think this is most of it. It all stays in cohortanalysis and cohortresults packages, so shouldn't affect anything existing in master.


src/main/resources/resources/cohortresults/sql/condition/byConcept/sqlPrevalenceByGenderAgeYear.sql
```diff
@@ -0,0 +1,38 @@
...
+	SELECT CAST(num.stratum_1 AS INT) AS num_stratum_1,
+		CAST(num.stratum_2 AS INT) AS num_stratum_2,
+		CAST(num.stratum_3 AS INT) AS num_stratum_3,
+		CAST(num.stratum_4 AS INT) AS num_stratum_4,
+		num.count_value AS num_count_value,
+		denom.count_value AS denom_count_value
+	FROM (
+		SELECT *
+		FROM @resultsSchema.dbo.heracles_results
+		WHERE analysis_id = 404
```
chrisknoll: The @resultsSchema.dbo should just be @resultsSchema.  I think we need to look through all the scripts to ensure that there's no .dbo in any of the scripts.  The getOhdsiSchema() handles getting the proper schema.

src/main/resources/resources/cohortresults/sql/condition/sqlConditionTreemap.sql
```diff
@@ -0,0 +1,101 @@
...
+		select snomed.concept_id, 
+			snomed.concept_name as snomed_concept_name,
+			pt_to_hlt.pt_concept_name,
+			hlt_to_hlgt.hlt_concept_name,
+			hlgt_to_soc.hlgt_concept_name,
+			soc.concept_name as soc_concept_name
+		from	
+		(
+		select concept_id, concept_name
+		from @cdmSchema.dbo.concept
```
chrisknoll: like @results_schema, @cdmSchema.dbo should just be @cdmSchema.

Also, I'm working on my own fork to normalize on these variable names. I've been asked to make the cdm schema: @cdm_database_schema. and results is @ohdsi_database_schema.

cahilton: Okay - I can work on that.

src/main/java/org/ohdsi/webapi/service/CohortAnalysisService.java
```diff
@@ -192,13 +210,14 @@ public String getRunCohortAnalysisSql(CohortAnalysisTask task) {
...
-        String[] values = new String[] { this.getCdmSchema(), this.getOhdsiSchema(), this.getSourceName(), 
+                "measurement_concept_ids", "cohort_period_only" };
+        String[] values = new String[] { utils.getCdmSchema(), utils.getOhdsiSchema(), utils.getSourceName(), 
         		String.valueOf(task.getSmallCellCount()),
-                String.valueOf(task.runHeraclesHeel()).toUpperCase(), this.getCdmVersion(), cohortDefinitionIds,
-                analysisIds, conditionIds, drugIds, procedureIds, observationIds, measurementIds };
+                String.valueOf(task.runHeraclesHeel()).toUpperCase(), utils.getCdmVersion(), cohortDefinitionIds,
+                analysisIds, conditionIds, drugIds, procedureIds, observationIds, measurementIds,
+                String.valueOf(task.isCohortPeriodOnly()) };
         sql = SqlRender.renderSql(sql, params, values);
```
chrisknoll: This is intesting: i don't know why this would work on oracle if there is tokens in the script with .dbo:  @cdmSchema.dbo replaced in sqlRender with {oracle_schema} would be {oracle_schema}.dbo.SOME_TABLE.  That can't work, can it?  I know it does work for you, but wondering if there's something I don't understand about sqlrender that is happening here...

@schuemie , any comments?

chrisknoll:
Let's just get the schema tokens in the scripts ironed out and we'll merge this in.

src/main/java/org/ohdsi/webapi/service/CohortAnalysisService.java
cahilton: It definitely works, but I can remove the .dbo

pbr6cornell: SqlRender is replacing '.dbo' if the target RDBMS is not SqlRender.  A
historical artifact that we need to clean up in earlier apps (ACHILLES,
HERACLES SQL scripts) was use of @CDM_schema.dbo.<table name>,  instead of
the new convention that we agreed to of:  @CDM_database_schema.<table
name>.    Here's that thread:
http://forums.ohdsi.org/t/getting-rid-of-dbo-in-our-parameterized-sql/267

This will allow SQL Server users who want to call a different schema name
to do so.

cahilton:
I think I have all the dbo references removed, and I tested it. Going to look at using the same names.

cahilton:
I switched over to use the normalized names for cdm and ohdsi schema, and I have tested i.t

PR #36, opened 2015-04-15
Cohort expression sql optimize

chrisknoll:
This pull request adds some optimization tricks with the cohort expression translation. We were running into issues doing large-scale intersects and it was giving us performance problems. We now leverage a group-by-count technique to find members that exist across all sets.


chrisknoll:
I would appreciate it if any of our non-sql partners could check out this branch and try to run a cohort creation (just access the webapi /cohortdefinition/{id}/generate and you should see in the console log that the job completed successfully.  I really should put together a test case for this, but I'm not sure our build server has the necessary CDM schema installed on it for me to write a test against.

chrisknoll:
I'll merge this request at the end of the day if there are no objections.

PR #37, opened 2015-04-18
WebAPI laertes services

leeevans:
This request incorporates the Laertes WebAPI services that Rich created.  I created the flyway migration scripts for all 3 DBMSs to create the new Laertes tables. The Postgresql migration was tested from my local machine.  I have manually run the Oracle and SQL Server SQL scripts on those DBMSs.  The Laertes tables are now created in the ohdsi schema instead of the cdm schema.  

Some minor changes were necessary to the original Laertes table DDL to enable them to work across DBMSs. In particular replacing postgresql serial with sequence object, replacing postgresql boolean with varchar(1) column and replacing postgresql text with varchar(4000) (Oracle limitation).


chrisknoll:
Lee, check out how I defined the text column in cohort definiton details:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/oracle/V1.0.0.3__cohort_definition_persistence.sql

You can declare the column as a CLOB, but you'll have to annotate the class field as so:
```java
 @Lob
  @Type(type = "org.hibernate.type.TextType")  
  private String expression;
```

That should get you around any limitations you see. Also, i'm worried about a postgresql bolean being used as a varchar(1).  Can an int suffice, or is there a datatype of 'bit'?

chrisknoll:
I should add: i believe CLOB is the oracle database table type.  text is the postgres type, varchar(max) is for MSSQL.  You apply the @LOB and the @Type attributes on the field in Java so that the persistence framework handles the serialization properly.  

I'd like to also understand the nature of the long text field: is it basically free text, like you'd be storing very large text files in the row?

-Chris

leeevans:
Chris, thanks for the info on CLOB support - that's good to know.
In this case, none of the postgresql text fields need to hold long text fields or files.  

I have already loaded the LAERTES data tables in the LAERTES postgresql database so you can see for yourself that the data in these tables/columns is not very wide (way less than the 4000 chars that I allowed for). I'm not aware of any requirement to convert these text columns into CLOBs.  The Oracle limitation I was referring to in my request is that Oracle doesn't have the same flexibility as postgresql to allow a varying character column without specifying a max length -  it wasn't to say that 4000 characters is a limitation. I should have been more specific.

There is a single postgresql boolean column called MODALITY.  Oracle is limited because it doesn't have a boolean or bit column type. Rich defined MODALITY as string in the evidence services and casts the column to a string resulting in the service returning a value of 't' or 'f'.  Therefore I went with an Oracle implementation of VARCHAR(1) to implement the same behavior in Oracle deployments.

Note. I have tried as much as possible to keep any LAERTES schema changes to an absolute minimum for this pull request but these minor changes were necessary.  My focus was on flyway migration of the LAERTES tables in the OHDSI schema in order to help Rich get the LAERTES services incorporated into the WebAPI and deployed in the OHDSI public server.

chrisknoll:
Ok, good to know.  If you've verified it on the differnt RDBMS then we can proceed with the merge.

fdefalco:
Thanks @leeevans - All looks good as I'm now able to access the evidence APIs.

PR #38, opened 2015-04-30
Services for cohort specific drilldowns

cahilton:
3 new queries for Heracles


cahilton:
Merging so we can wrap up Olympus build - changes isolated to Heracles.

PR #40, opened 2015-05-14
ported schema changes to the Leartes evidence webapi

rkboyce:
The Laertes evidence component is now current with the table schema as present in the OHDSI schema. Flyway was used for migration. All my testing indicates that the code runs fine and is ready merging upstream.


anthonysena:
Rich, we were taking a look at this branch and came across a few flyway scripts that need adjustment for SQL Server.

The following script needs to be renamed to include two underscores: V1.0.0.6.1__schema-create_laertes.sql

Then in V1.0.0.6.2__schema-create_laertes.sql the following needs a syntax change:

ALTER TABLE LAERTES_SUMMARY DROP COLUMN eb05;
ALTER TABLE LAERTES_SUMMARY DROP COLUMN ebgm;

I made the adjustments to a clone of your code locally and once I did that I was able to verify that the evidence service is in good working order. If you could make these adjustments, we'll get this merged into the main branch. Thanks!

rkboyce:
Ok. I implemented those changes and synced my fork with OHDSI. Please let me know if there is anything further I need to do. Thanks! -R

rkboyce:
Sorry about that guys. I think it is fixed now. Would you please recheck? thanks

anthonysena:
Looks like the old file is still hanging around in the source tree which is causing the flyway script deployment to fail in my testing. Can you please remove this file: V1.0.0.6.1_schema-create__laertes.sql and then we should be set. Thanks!

chrisknoll:
@rkboyce , you can delete the old file directly in github UI. Just navigate to the folder, click the file, and then click the trashcan icon on the top right to delete the file. I think it will automatically commit the delete (you may need to provide a message).  I'm not sure why it didn't identify the new file as a rename unless you made a copy of the file and gave it a different name.

anthonysena:
@chrisknoll let's merge in Rich's changes and delete the flyway script that is no longer needed. I've tested and all of the LAERTES services are working well.

PR #43, opened 2015-06-09
Add create index to flyway

cahilton:
Originally, I ran the Heracles indexes manually. 

This will allow to be ran with flyway, also will probably help with https://github.com/OHDSI/Heracles/issues/12


cahilton:
Thanks. Alright, I guess Jon took the use out of the Heracles scripts, but didn't finish the fully qualifying of the tables, so I'm changing/testing that now. Will probably add to this pull request.

PR #78, opened 2015-11-03
OHDSI-60 => fix Oracle errors in penelope scripts.

chen-regen:


fdefalco:
I will follow up with @anthonysena and then look to merge this branch.

anthonysena:
Chen and I worked through these issues over email and I'm comfortable with merging these changes. Just let me know if we need to talk through any details. Thanks!

PR #91, opened 2016-04-04
LAERTES update changes some evidence table schemas and queries

rkboyce:
Hi, This pull request is  a bit urgent as we need to test the update before next Tuesday's all team call.  This pull request is to bring the changes needed to run the web api evidence services against the updated LAERTES database tables. A flyway script has been added to address schema changes and slight modifications were made in the EvidenceService and related so that the WebAPI will function with the updated database. Changes were tested in my development environment against a postgres DB. 

thanks,
-R


anthonysena:
Rich, I tested the changes locally and found a few issues with the SQL Server flyway scripts. I've revised them and attached them for review. Once I made these changes, the database was updated properly to reflect the changes.

I also modified the getEvidenceSummaryBySource.sql file to reflect the change in the column value type. In the version from the PR, this query was still using "supports = 1" and I've revised it to "supports = 't'" to make it consistent with the other queries and the code in the EvidenceService.java. Furthermore, I had to modify the check to the "supports" column in the EvidenceService.java. The "== "t"" operator wasn't working properly so I've replaced it with .equalsIgnoreCase("t") instead.

I've tested the services from my side and with these changes, all is working well. Can you make these updates and then update the PR? We should then be able to then merge this into master.

FYI - I had to add .TXT to the attachments in this comment.

[EvidenceService.java.txt](https://github.com/OHDSI/WebAPI/files/204906/EvidenceService.java.txt)
[getEvidenceSummaryBySource.sql.txt](https://github.com/OHDSI/WebAPI/files/204903/getEvidenceSummaryBySource.sql.txt)
[V1.0.0.6.3__schema-create_laertes.sql.txt](https://github.com/OHDSI/WebAPI/files/204904/V1.0.0.6.3__schema-create_laertes.sql.txt)
[V1.0.0.6.4__schema-create_laertes.sql.txt](https://github.com/OHDSI/WebAPI/files/204905/V1.0.0.6.4__schema-create_laertes.sql.txt)

PR #93, opened 2016-04-13
Concept Set Existance; Penelope Tables

anthonysena:
@fdefalco - looking for some feedback before merging into master

### DIFF
--- a/src/main/java/org/ohdsi/webapi/conceptset/ConceptSetRepository.java
+++ b/src/main/java/org/ohdsi/webapi/conceptset/ConceptSetRepository.java
@@ -15,7 +15,10 @@
  */
 package org.ohdsi.webapi.conceptset;
 
+import java.util.Collection;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.query.Param;
 
 /**
  *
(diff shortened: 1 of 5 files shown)


leeevans:
@anthonysena  -  here is some general info on creating postgreseql flyway scripts that may be helpful:

The following file is a good example of a postgresql flyway script: 

https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/postgresql/V1.0.0.6.1__schema-create_laertes.sql

It is idempotent by using 'drop if exists' and it also contains examples of creating postgresql sequence objects and indexes.  The flyway script will need to create postgresql sequence objects for any table that has  an identifier column and also create any needed indexes.

FYI. You can click on an existing LAERTES table in pgadmin and see the postgresql DDL. that's helpful for getting the index DDL for any existing tables that are moving between schemas.

anthonysena:
@leeevans thanks for the information and for the code pointer - this is certainly a good script. 

I had considered utilizing the "drop if exists' for these tables but decided not to since I didn't want to potentially drop these tables in the public environment since they are populated with data. For other tables though, I'll be sure to do that.

leeevans:
Thanks @anthonysena I hadn't considered how to handle the existing data in tables migrating from the public (cdm) schema to the ohdsi schema.

My understanding is that these tables are new in the ohdsi schema so there shouldn't be any existing data in them in the ohdsi schema in the OHDSI cloud (public environment).  The original public schema versions of the LAERTES/penelope tables were not created/managed by flyway. I believe Rich has a LAERTES SQL script that created and populated them.

If a flyway migration creates a table in the ohdsi schema that will replace an existing public schema table then we will need to populate the data in the new ohdsi schema table and manually drop the old public schema table.  This needs to be coordinated so it happens when the flyway migration is pushed to the release branch and auto-deployed to the OHDSI cloud.

I'll follow up with @rkboyce on Friday to point me to the LAERTES script to run and a list of tables to drop.

Issue #94, opened 2016-04-14
Embedded Tomcat


vnikolaenko:
Very usable feature to use embedded tomcat in spring boot.
This is my pull request. I added embedded tomcat support.
We only need to run command "java -jar WebAPI.jar"
https://github.com/OHDSI/WebAPI/pull/95

PR #95, opened 2016-04-14
Embedded Tomcat

vnikolaenko:
https://github.com/OHDSI/WebAPI/issues/94
I added embedded tomcat support. We can run it without other tomcat.
We only need to run command "java -jar WebAPI.jar"


fdefalco:
Can you explain the intent of this PR a little more? Would this change the default build target from a WAR to a JAR?

vnikolaenko:
We can use both war and jar targets. It doesn't matter. The application server and all configuration files will be located inside this archive. Anyone can run the command "java -jar" and work with the application without additional container settings. [Here is description](https://en.wikipedia.org/wiki/Spring_Framework#Spring_Boot)

chrisknoll:
@vnikolaenko , this PR is very stale at this pint, and it's till not clear what this PR is solving. So if you could reissue this off of the latest branch and provide some context (such as how we're not able to run this app as stand alone?  I believe I'm able to do this as a stand-alone app whenever I start a debug session).

Thank you.

Issue #141, opened 2016-11-07
Cohort Definition Primary Events optimization


chrisknoll:
It's a new pattern for some cohorts to look for visits and then select visits based on other correlated criteria.  This usually starts by picking primary events of 'Any Visit' but this can be a very expensive operation (it puts the results into the primary event query into a temp table).  Instead, if this primary event query could be combined with the qualifying criteria, the initial temp table would be reduced.

Additionally, for MPP environments, the correlated criteria is joined based on an event_id, and it would be better optimized if the events were partitioned by person_id, and then subsequent joins back to the event table are joined on both event_id and person_id.

Issue #139, opened 2016-11-02
'first occurrence' is not deterministic when first occurrence is on the same date as another row


chrisknoll:
We're seeing an issue on our MPP platform that when using row_number() that may return the first row as a different value between executions because there are multiple records with the same date. This query for example (querying procedure_occurrence):

````
select count(*) from (
	select C.person_id, C.procedure_date as start_date, DATEADD(d,1,C.procedure_date) as END_DATE, C.procedure_concept_id as TARGET_CONCEPT_ID
	from 
	(
		select po.*, ROW_NUMBER() over (PARTITION BY po.person_id ORDER BY po.procedure_date) as ordinal
		FROM CDM_TRUVEN_CCAE_v466.dbo.PROCEDURE_OCCURRENCE po
		where po.procedure_concept_id in (SELECT concept_id from  #Codesets where codeset_id = 22)
	) C
	JOIN CDM_TRUVEN_CCAE_v466.dbo.PERSON P on C.person_id = P.person_id
	JOIN CDM_TRUVEN_CCAE_v466.dbo.VISIT_OCCURRENCE V on C.visit_occurrence_id = V.visit_occurrence_id and C.person_id = V.person_id
	WHERE  C.ordinal = 1
		AND (C.procedure_date >= DATEFROMPARTS(2014, 01, 01) and C.procedure_date <= DATEFROMPARTS(2014, 12, 31))
		AND YEAR(C.procedure_date) - P.year_of_birth >= 50
		AND V.visit_concept_id in (9201)
) C
````

Returns varying results because there are multiple rows that have the same procedure_date.  The fix is to ensure that the row has a 'tie-breaker' so that there is deterministicly only one 'earliest row':

```
select count(*) from (
	select C.person_id, C.procedure_date as start_date, DATEADD(d,1,C.procedure_date) as END_DATE, C.procedure_concept_id as TARGET_CONCEPT_ID
	from 
	(
		select po.*, ROW_NUMBER() over (PARTITION BY po.person_id ORDER BY po.procedure_date, po.procedure_occurrence_id) as ordinal
		FROM CDM_TRUVEN_CCAE_v466.dbo.PROCEDURE_OCCURRENCE po
		where po.procedure_concept_id in (SELECT concept_id from  #Codesets where codeset_id = 22)
	) C
	JOIN CDM_TRUVEN_CCAE_v466.dbo.PERSON P on C.person_id = P.person_id
[...]

PR #142, opened 2016-11-08
Cohort def optimizations

chrisknoll:
This PR addresses two issues: #139 and #141.

The major enhancement is cohort queries should operate faster in a MPP environment since we are now partitioning on person_id for the generated event_ids, and it should show improvement in SMP platforms with clustered indexes on person_id.

The fix for #139 was to include the _id in the order to tiebreak to fix the non-deterministic behavior of MPP platforms.


chrisknoll:
Please hang on before approving this I might have found an unexpected issue.

chrisknoll:
Ok, this is ready for review now.  Thank you.

ericaVoss:
Nice job adding them everyone.  :)

chrisknoll:
@pbr6cornell , @anthonysena @fdefalco this has been open for 10 days without a review.  Please review and either approve or notify me of concerns.

chrisknoll:
Yes I ran internal tests using a version of Joe's cohort definition where the primary events were 'Any Visit' qualified with death events within 180 days after.  you can see this definition here:
http://hix.jnj.com/atlas/#/cohortdefinition/3333

From a review perspective, I think it's just a matter of reviewing the files and making a local build and generating a few cohorts of your own to see if it looks like everything is square.  I've done this myself, but looking for a PR reviewer.

anthonysena:
OK - I'll pull this down and test it locally. Please let me know if there are any other cohorts that are good candidates for testing and I'll review those as well.

chrisknoll:
@anthonysena , Thank you very much for taking this on.  I'd say if you have any current cohort defs that you test with in hixbeta, you can copy them to a new definition (so to keep the old results) and then execute the copy under the new codebase and check that the numbers line up. That is what I did, but a second check would be good.

PR #143, opened 2016-11-17
Updated sql to match SqlRender

tnovo:
SqlRender does not support 'TOP' instruction, changed as suggested by @schuemie in OHDSI/SqlRender#38


anthonysena:
Thanks @tnovo for the contribution!

PR #151, opened 2016-12-12
Apache Shiro integration revised

ganisimov:
Purpose of this PR is finalizing Apache Shiro integration. 

All commits, including @anthonysena 's one, were squelched down and one commit was appended to reflect changes of HTTP verbs.


chrisknoll:
Ok, this looks good.  Thanks for squelching this down, it'll merge into master easier now.  

Here's my plan:
I'm going to get this branch locally and do a quick smoketest, I don't think there's going to be any problems since we've (me and @anthonysena ) have been looking into this for a while and I think everything is worked out.  I'll try to complete that work by EOD.

Once everything checks out, i'm going to put this into a staging branch in the OHSDI repo.  The reason is that we need to push out a final 'pre JDK1.8' release that has the latest bug-fixes and enhancements that should be generally available before the JDK update.  

Then once we get that final release of JDK 1.7 out, we'll start merging the JDK1.8 features into master and start moving towards that Jan '17 release.

chrisknoll:
Ok, this was pushed into a shiro-final branch which we'll hold onto until we're ready for the JRE 1.8 pushes into master.  But, I confirmed that this is working locally, and so this PR is closed, and I'll prune all the obsolete branches.

Issue #160, opened 2017-01-19
mistake in getConceptRecordCount.sql?


Sigfried:
Trying to figure out the best way to get record counts, I came on this and think it might be wrong:

https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cdmresults/sql/getConceptRecordCount.sql#L11

Here's the query:

```
with concepts as (
	select CAST(ancestor_concept_id as VARCHAR) ancestor_id, CAST(descendant_concept_id as VARCHAR) descendant_id 
	from @vocabularyTableQualifier.concept_ancestor ca 
	where ancestor_concept_id in (@conceptIdentifiers)
), counts as (
	select stratum_1 concept_id, max(count_value) agg_count_value
	from @resultTableQualifier.achilles_results
	where analysis_id in (2, 4, 5, 201, 301, 401, 501, 505, 601, 701, 801, 901, 1001,1201,1801)
	group by stratum_1
	union
	select stratum_2 as concept_id, sum(count_value) as agg_count_value
	from @resultTableQualifier.achilles_results
	where analysis_id in (405, 605, 705, 805, 807, 1805, 1807)
	group by stratum_2
)
select concepts.ancestor_id concept_id, isnull(max(c1.agg_count_value),0) record_count, isnull(sum(c2.agg_count_value),0) descendant_record_count
from concepts
left join counts c1 on concepts.ancestor_id = c1.concept_id
left join counts c2 on concepts.descendant_id = c2.concept_id
group by concepts.ancestor_id
```

But look at the analysis descriptions:

```
select * from achilles_analysis where analysis_id in (405, 605, 705, 805, 807, 1805, 1807);
 analysis_id |                                           analysis_name                                            |     stratum_1_name     |       stratum_2_name        | stratum_3_name | stratum_4_name | stratum_5_name
-------------+----------------------------------------------------------------------------------------------------+------------------------+-----------------------------+----------------+----------------+----------------
         405 | Number of condition occurrence records, by condition_concept_id by condition_type_concept_id       | condition_concept_id   | condition_type_concept_id   |                |                |
         605 | Number of procedure occurrence records, by procedure_concept_id by procedure_type_concept_id       | procedure_concept_id   | procedure_type_concept_id   |                |                |
         705 | Number of drug exposure records, by drug_concept_id by drug_type_concept_id                        | drug_concept_id        | drug_type_concept_id        |                |                |
[...]

PR #162, opened 2017-01-20
add comments to record count query

Sigfried:


Sigfried:
fixes my confusion in #160

PR #174, opened 2017-02-07
Route cleanup.

### LINKED ISSUE #173: Missed table qualifier correction?
Should 

    String resultsTableQualifier = source.getTableQualifier(SourceDaimon.DaimonType.CDM);

[here](https://github.com/OHDSI/WebAPI/blob/0118f2897b5b1ecfd28a1887996211c9b764d17f/src/main/java/org/ohdsi/webapi/service/CohortResultsService.java#L1421) be changed to below?

    String resultsTableQualifier = source.getTableQualifier(SourceDaimon.DaimonType.Results);

chrisknoll:
Made service routes for CDMResutlsService, CohortResultsService, EvidenceService and Vocabulary service to stem from the url of the service (such as /cdmresults) vs. stemming from a source key.

Added a default handler for vocabulary searching as an example of where it's advantageous to use a route of /vocabulary/search/{query} directly instead of coding to a specific source key (which may not always be available).

Fixes #173.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,8 +138,8 @@
         <version>2.5.1</version>
         <inherited>true</inherited>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
         </configuration>
       </plugin>
       <plugin>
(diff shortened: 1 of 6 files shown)


chrisknoll:
Also, i should add that this PR modifies the pom.xml to allow source version 1.8. I'm using a 1.8 feature called 'streams' which makes filtering collections very elegant.

PR #192, opened 2017-03-16
Update Spring Libraries.

### LINKED ISSUE #186: Refresh WebAPI Spring Libraries
The following libraries require refresh to the latest GA release:
Spring Boot: 1.2.1.RELEASE -> 1.5.2.RELEASE
Spring Batch: 3.0.3.RELEASE -> 3.0.7.RELEASE

This will update Hibernate to 5.0.12, which will require some cleanup of our Entity definitions with respect to GeneratedValues(type=AUTO) annotations (will file separate issues to track this.)

chrisknoll:
Spring Boot -> 1.5.2
Spring Batch -> 3.0.7
Hibernate -> 5.0.12

Fixes #186.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,8 +13,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jersey.version>2.14</jersey.version>
     <!-- Spring Boot manages spring.version as well -->
-    <spring.boot.version>1.2.1.RELEASE</spring.boot.version>
-    <spring.batch.version>3.0.3.RELEASE</spring.batch.version>
+    <spring.boot.version>1.5.2.RELEASE</spring.boot.version>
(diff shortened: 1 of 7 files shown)


fdefalco:
Noticed the new gs-collections dependency, where is it being used?  Noticed that the library was migrated to eclipse foundation and rebranded, would we want to use that instead?

https://www.eclipse.org/collections/

chrisknoll:
The reference was it in line 3 of UserService.java:

```java
import com.gs.collections.impl.block.factory.Comparators;
```

When I updated the library I couldn't find how it was being referenced, so I searched online and saw it was part fo the GS library. If you know which dependency should be used, please make a commit to this branch that adjusts the POM.xml.

chrisknoll:
I've made the requested change, and the only impact I found was in UserService.java.

PR #205, opened 2017-05-03
Enquote ORDER BY columns in cdm results data density queries

### LINKED ISSUE #183: CDM results - data density query error
The data density web service call is failing on OHDSI.org

http://api.ohdsi.org/WebAPI/cdmresults/1PCT/datadensity

Tracing the error a bit reveals: 

_ERROR: column \"series_name\" does not exist\n  Position: 1616_

Tagging @t-abdul-basser for awareness. This error does not happen on our SQL Server instances internally so it is likely a PostgreSQL related issue with the queries utilized in the report.

mark-velez:
fixes #183

### DIFF
--- a/src/main/resources/resources/cdmresults/sql/report/datadensity/recordsperperson.sql
+++ b/src/main/resources/resources/cdmresults/sql/report/datadensity/recordsperperson.sql
@@ -1,7 +1,7 @@
 SELECT
-  t1.table_name                                      AS "series_Name",
-  t1.stratum_1                                       AS "x_Calendar_Month",
-  round(1.0 * t1.count_value / denom.count_value, 5) AS "y_Record_Count"
+  t1.table_name                                      AS series_Name,
+  t1.stratum_1                                       AS x_Calendar_Month,
+  round(1.0 * t1.count_value / denom.count_value, 5) AS y_Record_Count
 FROM
   (
     SELECT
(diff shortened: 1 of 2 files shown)


chrisknoll:
Hey, @mark-velez ,
  I'm wondering if quoting the order by is the less desirable approach vs. removing the quoted identifiers out of the select.  Was there a reason of choosing one over the other?  I only hesitate because I'm not sure we see the same semantics of what the quoted columns means across db platforms (as is the case for this specific issue between postgresql and mssql), and I think it is more universally accepted that non-quoted identifiers is case-insensitive when referrering to them in expressions. 

So what's your thoughts about solving this by removing the quotes from the select vs. adding them to the order by?

leeevans:
@chrisknoll in support of your point about quoted column semantic differences across DBMSs, Redshift doesn't treat quoted column names as case-sensitive:

http://docs.aws.amazon.com/redshift/latest/dg/r_names.html
**"ASCII letters in delimited identifiers are case-insensitive and are folded to lower case. "**

mark-velez:
@chrisknoll (and @leeevans) thanks for checking this out. No deep thinking on my part, only tried [your suggestion](https://github.com/OHDSI/WebAPI/issues/183#issuecomment-282133063) and saw it pass on postgres/mssql. I can remove the quotes instead if you think that's more cross-compatible.

chrisknoll:
Yep, my suggestion was more of a work-around to clear the error, but I later said that we probably can't rely on quoted identifiers being equally respected, so I think the proper fix is to remove the quoted identifiers.  So, go ahead and remove the quotes and if it passes postgres/sql, report back and we'll get it merged.

mark-velez:
@chrisknoll I removed quoted identifiers and it passed postgres/mssql

PR #213, opened 2017-07-06
Fixes to get (some of) ATLAS running against Impala

tomwhite:
https://github.com/OHDSI/Atlas/issues/418

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
+++ b/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
@@ -90,9 +90,10 @@ public class GenerateCohortTasklet implements Tasklet {
       options.targetTable = jobParams.get("target_database_schema").toString() + "." + jobParams.get("target_table").toString();
       options.resultSchema = jobParams.get("results_database_schema").toString();
       options.generateStats = Boolean.valueOf(jobParams.get("generate_stats").toString());
-      
-      String deleteSql = String.format("DELETE FROM %s.cohort_inclusion WHERE cohort_definition_id = %d", options.resultSchema, options.cohortId);
-      this.jdbcTemplate.update(deleteSql);
+
+      String deleteSql = String.format("DELETE FROM %s.cohort_inclusion WHERE cohort_definition_id = %d;", options.resultSchema, options.cohortId);
+      deleteSql = SqlTranslate.translateSql(deleteSql,"sql server", jobParams.get("target_dialect").toString(), sessionId, null);
+      this.jdbcTemplate.batchUpdate(deleteSql.split(";")); // use batch update since SQL translation may produce multiple statements
 
       String insertSql = StringUtils.replace("INSERT INTO @results_schema.cohort_inclusion (cohort_definition_id, rule_sequence, name, description) VALUES (?,?,?,?)", "@results_schema", options.resultSchema);
       insertSql = SqlTranslate.translateSql(insertSql,"sql server", jobParams.get("target_dialect").toString(), sessionId, null);
--- a/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
+++ b/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
@@ -63,7 +63,7 @@ public class DataAccessConfig {
         //note autocommit defaults vary across vendors. use provided @Autowired TransactionTemplate
 
         String[] supportedDrivers;
-        supportedDrivers = new String[]{"org.postgresql.Driver", "com.microsoft.sqlserver.jdbc.SQLServerDriver", "oracle.jdbc.driver.OracleDriver", "com.amazon.redshift.jdbc41.Driver"};
+        supportedDrivers = new String[]{"org.postgresql.Driver", "com.microsoft.sqlserver.jdbc.SQLServerDriver", "oracle.jdbc.driver.OracleDriver", "com.amazon.redshift.jdbc41.Driver", "com.cloudera.impala.jdbc4.Driver"};
 
         for (String driverName : supportedDrivers) {
             try {
--- a/src/main/resources/resources/cdmresults/sql/getConditionOccurrenceTreemap.sql
+++ b/src/main/resources/resources/cdmresults/sql/getConditionOccurrenceTreemap.sql
@@ -1,8 +1,11 @@
 SELECT
   concept_hierarchy.concept_id,
-  isNull(concept_hierarchy.soc_concept_name, 'NA') + '||' + isNull(concept_hierarchy.hlgt_concept_name, 'NA') + '||' +
-  isNull(concept_hierarchy.hlt_concept_name, 'NA') + '||' + isNull(concept_hierarchy.pt_concept_name, 'NA') + '||' +
-  isNull(concept_hierarchy.snomed_concept_name, 'NA')    concept_path,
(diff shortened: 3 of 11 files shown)


src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
```diff
@@ -90,9 +90,12 @@ public GenerateCohortTasklet(
       options.targetTable = jobParams.get("target_database_schema").toString() + "." + jobParams.get("target_table").toString();
       options.resultSchema = jobParams.get("results_database_schema").toString();
       options.generateStats = Boolean.valueOf(jobParams.get("generate_stats").toString());
-      
-      String deleteSql = String.format("DELETE FROM %s.cohort_inclusion WHERE cohort_definition_id = %d", options.resultSchema, options.cohortId);
-      this.jdbcTemplate.update(deleteSql);
+
+      if (!jobParams.get("target_dialect").toString().equals("impala")) {
```
chrisknoll: Hi, I don't think we should have dialect specific checks here.  This is an example where impala needs to add delete table support in order for our tools to function.  Removing this delete means that you'll just continue adding records into the table which will cause subsequent executions to fail.

I thought we found a certain table option in Impala that would allow table deletes?  My understanding was that we could create the results schema tables (the ones that will be inserted/dropped) using a particular table type, which would let these queries execute properly?

tomwhite: @chrisknoll I agree that dialect-specific checks are not a good idea. Impala can't handle deletes for Parquet-based tables, and we haven't got it working against Kudu yet.

Would you be OK merging the other two commits in this PR so that some parts of Atlas are working with Impala at least?

chrisknoll: This puts me in a difficult spot:  On one hand I want the impala capabilities to be available, on the other hand anyone who gets it and expects it to work will see it run one time properly and then wonder why it isn't working right after they run it again.  I'm not trying to claim that everything in WebAPI must be bug-free but in this case this code only works exactly once.  

So is there any chance we could keep this off into a branch until we get the required capabilities out of the Impala platform (maybe after Kudu is hooked up)?

tomwhite: @chrisknoll These are very reasonable points. The Kudu work is some way off, so I have a new suggestion which is to get DELETE FROM working properly in Impala on Parquet using an approach suggested at the OHDSI Hadoop Hackathon a couple of weeks back, by using a temporary table. See https://github.com/OHDSI/SqlRender/pull/81.

I've updated this PR to take advantage of this. How does this look to you now?

chrisknoll:
Looks good, I get that it's using the 'delete means re-insert everything except what you delete'.  That'll work more than once!  Looks like there's some sort of test case that's failing on the sql render side, so once that is squared, I'll merge this in.

Thanks!

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
chrisknoll: Looks good, I get that it's using the 'delete means re-insert everything except what you delete'. That'll work more than once! Looks like there's some sort of test case that's failing on the sql render side, so once that is squared, I'll merge this in.

Also, you fixed the issue where the delete statement wasn't being translated. Thanks for that.

chrisknoll:
@tomwhite , There's a another  place that delete happen but in the resource files so you might not have seen this:

Remove prior cohort results:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql#L59

Remove the inclusion rule match stats:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql#L73

Removes the cohort summary stats:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql#L105

The good news is that with your SQL render change that handles the deletes, these should just continue to work, but I wanted to call these out in case you wanted to step through the code and verify that the delete/inserts are working as expected.

tomwhite:
@chrisknoll Thanks for pointing out the other deletes. I have generated a cohort successfully now, so it looks like it works.

chrisknoll:
Ok, cool. I'll get this merged in.

It might be worth pointing out to the people using Impala in this mode that it's possible to run into a race condition where if 2 cohorts are being generated at the same time (which happens in our environment every day) that the delete mode of 'insert everything except deleted' will break if there's 2 people performing that op at the same time (I guess the last one to complete wins).  But at least we get to 'it runs more than 1 time in a row' with this change, so that'll get it in.  

-Chris

fdefalco:
This is awesome @tomwhite - let me know if there is anything I can do to help with the Kudu integration.

tomwhite:
@chrisknoll Yes, that's a very important limitation to bear in mind. Thanks for merging this. There will be more changes to get more ATLAS features working with Impala, but this is a good start.

@fdefalco Kudu is going to take a while, but in the meantime if you could look at https://github.com/OHDSI/Atlas/pull/419 and https://github.com/OHDSI/CommonDataModel/pull/68 that would be greatly appreciated.

chrisknoll:
@tomwhite:
Not to be a pest, but since we're on the topic of Kudu, I recall back in November '16 we spoke about the need for deletes, (http://forums.ohdsi.org/t/first-achilles-queries-working-with-hadoop/1876/2) and at that point it seemed a long way off.  But we're revisiting the need again 6 months later, and it's still a long ways off.  Do you know what the bottleneck of getting Kudu implemented?  Is anyone actively working on this?  When something takes this long, I feel that it's either 1) requires a lot of manual effort (like a task has to be repeated a million times) so in that case many hand make light work (but there are no hands working) or 2) No one is really working on it since people feel that there's an adequate workaround.

If it's 2), then I feel like I've just bolstered that position by merging this PR with a fragile workaround but now people might feel that they just kicked the can on Kudu.  If that's the case, then I feel that future PRs are going to be met with much more opposition before merging in order to force the issue.  

Would you be able to give any estimate when Kudu support would be made available?

tomwhite:
@chrisknoll the Kudu blocker is support for timestamps. This work has involved changes to both Kudu and Impala, and has only recently been completed (https://issues.apache.org/jira/browse/IMPALA-5137). So I don't think it's either 1) or 2) of your reasons, it's just that the support has not been there in Kudu and Impala.

Coincidentally, CDH 5.12 was released today with the latest versions of Kudu and Impala, which means it's much easier to try this out now. I hope to do this soon.

chrisknoll:
@tomwhite , cool, glad I'm wrong about that.  I don't think that there's any timestamp columns in the atlas side of the results schema...just closing my eyes and thinking hard...can't remember anything like that.  Most of the results is just counts and record IDs.  Achilles might be a different story because I think in the case of observation period it uses timestamps.  But maybe this means that it can be attempted even without the timestamp support on the Atlas tables.  But this new release looks like it makes thing better. Let us know one way or the other if the latest release moves this issue forward.

Thanks!

-Chris

tomwhite:
Atlas does a `DELETE FROM cohort`, which does contain a timestamp unfortunately... I'll report back when I get a chance to try out the new release.

Tom

chrisknoll:
Oh yes, cohort start and cohort end.  That's right, I forgot.

Issue #217, opened 2017-07-27
WebAPI Release 2.2.0 Planning


chrisknoll:
Update 9/21/2017:
We've extended the deadline for submitting pull requests to 9/29/2017.

Update 9/7/2017:
Below is an enhancement matrix of completed requests:
- [x] Cohort Characterization @alondhe
- [ ] Cohort Comparison @pbr6cornell 
- [ ] Vulnerability Fixes @gklebanov 
- [ ] Cohort Analysis code sharing @gklebanov 
- [x] Cohort Definition external library @gklebanov , @chrisknoll
- [x] Atlas charting library externalized and upgraded to d3v4 @gklebanov , @chrisknoll
- [x] New graph in datasources @PRijnbeek 
- [x] Performance enhancements in data sources @PRijnbeek 
- [x] Data completeness @chen-regen 
- [x] Observation Entropy @chen-regen 
- [x] Timeliness @chen-regen 
- [x] WebAPI PLP specification and editor @jreps 
- [x] Cohort collapse option (era-fy) @gowthamrao 
- [x] Same-visit cohort criteria @gowthamrao 
- [ ] Cost table option in cohort criteria @gowthamrao 
- [ ] Use payer plan period in cohort criteria @gowthamrao 
- [ ] Combine cohrots into new cohorts @gowthamrao 
- [ ] Additional cohort characterization options @gowthamrao 
- [x] Concept set quickview in Atlas @jhardin29 
- [x] Big Query support @t-abdul-basser 
- [ ] Assorted Achilles Heel enhacements @t-abdul-basser 
- [ ] Embeded DB into WebAPI @t-abdul-basser, @alondhe 
- [ ] Estimation enhancements @anthonysena 

Please comment in this thread if any of the above has been merged into master and should be marked completed.

Original Message:

The OHDSI Web Architecture group is planning a 2.2.0 release prior to the 2017 OHDSI Symposium.  This thread is to call for contributions for  features that we'd like to include in this release scheduled for 10/1/2017.

Contributions are due 9/22/2017 for inclusion in the release, to be provided as a pull request .

In order to submit an idea for a feature, please respond to this thread with the following information by 8/22/2017 so that we can anticipate any potential challenges for incorporating the features:

- A short description of the feature
- The organization or developers that are involved in the implementation
- Any anticipated impact to current WebAPI/Atlas functionality (if applicable)

After you have responded to this thread, please create a new issue so that the specific feature can be tracked in the V2.2.0 milestone.

[...]

PR #234, opened 2017-08-11
Cohort era-fy option

gowthamrao:
Added Event End Date Strategy template
- Get date field, get group for event end date strategy

Added Get Strategy SQL for Event End Date Strategy

OHDSI/WebAPI#217
OHDSI/Atlas#414
OHDSI/Atlas#375


gowthamrao:
OHDSI/WebAPI#217
OHDSI/Atlas#414
OHDSI/Atlas#375

chrisknoll:
Hi, @gowthamrao , before starting a a review for this, I just wanted to make sure I understand what the Event End Date Strategy functionality is performing.  

From the ieeuse OHDSI/Atlas#414, you are requesting a way to create cohorts where the cohort end date is based on the visit end date.  I've attempted to do that with the current functionality where the primary event is a visit, the diagnosis of interest is 'essential hypertension' and I want to cohort end date to be the end date of the visit. You can see this here: http://www.ohdsi.org/web/atlas/#/cohortdefinition/919659

The primary event will be the visit start date and end date, as seen in this SQL:

``` sql
-- Begin Visit Occurrence Criteria
select C.person_id, C.visit_occurrence_id as event_id, C.visit_start_date as start_date, C.visit_end_date as end_date, C.visit_concept_id as TARGET_CONCEPT_ID
```

By default the cohort end date would be the observation period end date unless a end date strategy i specified.  In this case, I'm using the 'Cohort exit criteria based on a fixed time period relative to initial event start or end date', and I specify that i want to use the primary event (the visit) end date, with an offset of 0 days.  This will make the cohort end date be the visit end date.

So, for this first topic, I'd like to understand how the Event End Date strategy differs from specifying a 0 day offset from the primary event's end date.

Second topic is the query change related to the issue about duplicate rows: OHDSI/Atlas#375: I believe the removal of duplicate rows is not dependent on the selected end date strategy.  For example, if we are searching for drug X, and we see 2 exposures start on the same day, but one end date is a single day, and another is 7 days, I'd thought that we'd want to just collapse those exposures into a single exposure with a 7 day suration fur purposes of defining the primary event's start and end date.  So, the suggestion there was to just group by person, start_date, max(end_date) per the suggestion in OHSI/Atlas#375.  But not based on a specified end date strategy, I think this was just a broader problem where cohorts could have duplicate start/end dates in them and that is something that should be fixed in general.

Let me know if I am misunderstanding anything.

gowthamrao:
Thank @chrisknoll  Regarding topic#1, the generated SQLcode is as follows for end-date strategy in the Atlas cohort you build
```

-- Apply end date stratagies
-- by default, all events extend to the op_end_date.
CREATE TEMP TABLE cohort_ends
AS
SELECT
event_id, person_id, op_end_date as end_date
FROM
included_events;

-- Date Offset Strategy
INSERT INTO cohort_ends (event_id,  person_id, end_date)
select event_id, person_id, 
  case when (end_date + 0*INTERVAL'1 day') > start_date then (end_date + 0*INTERVAL'1 day') else start_date end as end_date
from included_events
;
```
Does that mean the temp table cohort_ends has two inserts?

chrisknoll:
Hey, @gowthamrao.  You got it:  the cohort_ends temp table is used to select the appropriate end date for a given included event_id.  So, by default, every event is followed up to the observation period end date that contains the event.  Then, we add the other end dates for the strategy for each event.  It's true that it's a safe assumption that visit ends will be contained within the observation period, but what if you said that you wanted to specify the cohort end date as 365d after the visit end date?  And also let's say that in this case, the person's observation period ended 30d after the visit.  So what you get in the cohort end date is:

event id | person id | end_date | 
|---------|-----------|------------|
E1 | P1 | 4/1/2014
E1 | P1 | 1/1/2015

This assumes that the visit end date was 1/1/2014, the observation period end date was 4/1/2014, and adding 365d to the visit end date will produce the 1/1/2015 end date in the temp table.  In this case, we're going to select the OP end date as the cohort exit because adding 365d to the visit end date pushes it past this person's observation period end date, and we should use the op end date as the cohort end date in this case.  Of course, if we only added 0d to the visit end date, we wouldn't select the OP end date because, as you say, the visits should be contained completely in an observation period.

As an asside: other records get inserted into the cohort_ends table based on censoring events (you should check out that function too, lets you do things like 'person exits cohort after switch').

The final step of using the cohort_ends is to find the earliest end_date per primary event.  This is the final cohort start-end pair that gets put into the cohort table.    

But, per your issue  OHDSI/Atlas#375 about duplicate rows, I would really love it if this PR did address that issue, but without using a new end date strategy.  I can provide guidance on how this might be implemented, but I think you're getting some excellent exposure to the inner workings of the cohort generation sql, so I'd love it if you could work through this issue and propose a PR to fix it.   Doing the group by start_date, max(end_date) is the right approach, but remember that the records being put into the cohort table are based off of the included events...these events have unique event_ids, but non-unique start dates.   So I think the place to do the group by start_date, max(end_date) is either when identifying the qualified_events temp table, or you can make it a final step that after each event's end_date has been identified, you then do the group by start_date, max(end_date) to remove duplicates from the cohort records.

-Chris

chrisknoll:
@gowthamrao , I just game to a realization that you probably shouldn't try to remove duplicate start dates at the beginning of the cohort generation process (at the time of creating the included events) because you want to apply inclusion criteria to all the events that are selected...if you collapse early, you could lose events that you didn't mean to.

Instead, if you just de-dupe at the end when the start-end dates have been established, then you are sure that all the inclusion criteria were applied properly to all the events and all you are doing is removing duplicated start dates, and using the longest cohort end date that shared the same start.

gowthamrao:
Thanks Chris - how about just making this default
```

DELETE FROM @target_database_schema.@target_cohort_table where cohort_definition_id = @target_cohort_id;
INSERT INTO @target_database_schema.@target_cohort_table (cohort_definition_id, subject_id, cohort_start_date, cohort_end_date)
select @target_cohort_id as cohort_definition_id, F.person_id, F.start_date**, max(F.end_date) as end_date**
FROM (
  select I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
  from included_events I
  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
) F
WHERE F.ordinal = 1
**group by F.person_id, F.start_date**
;
```

instead of current
```

DELETE FROM @target_database_schema.@target_cohort_table where cohort_definition_id = @target_cohort_id;
INSERT INTO @target_database_schema.@target_cohort_table (cohort_definition_id, subject_id, cohort_start_date, cohort_end_date)
select @target_cohort_id as cohort_definition_id, F.person_id, F.start_date, F.end_date
FROM (
  select I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
  from included_events I
  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
) F
WHERE F.ordinal = 1
;
```

gowthamrao:
For OHDSI/Atlas#375, there are the following issues for same subjects
1. records with same start date and end date
2. records with same start date but different end date

There may be use-cases to support both of them, although I dont see value for 1 above. I do see a value for 2 above - where a person may have double exposure during some overlapping period. 

max(end_date) deduplicates 1 and forces 2 to always be the record with max(end-date). I think max(end_date) as default would meet 80% of all use-cases, but if we want to support the rest of 20% we need end-date strategy.

How about we make max(end_date) as default strategy -- and then rest as optional overides?


```
DELETE FROM @target_database_schema.@target_cohort_table where cohort_definition_id = @target_cohort_id;
INSERT INTO @target_database_schema.@target_cohort_table (cohort_definition_id, subject_id, cohort_start_date, cohort_end_date)
select @target_cohort_id as cohort_definition_id, F.person_id, F.start_date, max(F.end_date) as end_date
FROM (
  select I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
  from included_events I
  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
) F
WHERE F.ordinal = 1
group by F.person_id, F.start_date
;
```

from original


```
DELETE FROM @target_database_schema.@target_cohort_table where cohort_definition_id = @target_cohort_id;
INSERT INTO @target_database_schema.@target_cohort_table (cohort_definition_id, subject_id, cohort_start_date, cohort_end_date)
select @target_cohort_id as cohort_definition_id, F.person_id, F.start_date, F.end_date
FROM (
  select I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
  from included_events I
  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
) F
WHERE F.ordinal = 1
;
```

gowthamrao:
How about we create a new 'Cohort output strategy' (to be applied after cohort exit strategy).

Instead of this being hardcoded in generateCohort.sql
```
DELETE FROM @target_database_schema.@target_cohort_table where cohort_definition_id = @target_cohort_id;
INSERT INTO @target_database_schema.@target_cohort_table (cohort_definition_id, subject_id, cohort_start_date, cohort_end_date)
select @target_cohort_id as cohort_definition_id, F.person_id, F.start_date, F.end_date
FROM (
  select I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
  from included_events I
  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
) F
WHERE F.ordinal = 1
;
```
we create a new sql - say cohortOutputStrategy.sql and manage that with Atlas interface +java to do the following:
1. (Default)  - return records with unique combination of subject_id, start_date - by using max(end_date)
2. (Option) - return records with unique combination of subject_id, start_date - by using min(end_date)
3. (Option) - return all records
4. (Option) - return distinct records

gowthamrao:
Chris - I agree with your comments, and I also think that using some form of 'collapsing' strategy after the end-date strategy has been selected, is the way to go. If we define cohort as non-overlapping period records of subjects with a start-date and end-date, then the era-fying solution works. I agree with you, and would like to team up to make era-fying default. i.e. make the Option 5 from above OHDSI/Atlas#407 default.

However, I don't think the community unanimously defines cohort as non-overlapping period records of subjects with a start-date and end-date. If we look at the [cohort](https://github.com/OHDSI/CommonDataModel/blob/master/Documentation/CommonDataModel_Wiki_Files/StandardizedDerivedElements/COHORT.md) documentation - no where (that i know of) has it been explicitly mentioned that it has to be non-overlapping period records (but i have heard this in some discussions). 

There are few use-cases (I can't think a specific one right now) where
> people could have duplicate presence within the same period of time, thus resulting in double counting.
is actually fine. 
so, if we make the era-fying as default 'collapse' strategy - but then allow for other options.

gowthamrao:
Era-fy code: The generic era-constructor code is here OHDSI/Atlas#407 and is originally from https://github.com/OHDSI/Era-Constructor

Should we have a generic @eraConstructor code that takes table in the form id, start_date, end_date and builds another table with id, start_date, end_date ? That way the era-constructor may be reused anywhere in WebApi - such as in Cohort characterization for example OHDSI/WebAPI#235

pbr6cornell:
A cohort is defined as a set of persons who satisfy one or more inclusion criteria for a duration of time.  By this definition, a person can belong or not belong to a cohort at any moment in time, but it is not possible for a person to contain multiple records indicating cohort membership that are overlapping in time,

pbr6cornell:
One implementation solution within ATLAS would be to era-fy overlapping records.

but if there is a need for a construct to count records that occur within a given time period, i would not use the Cohort table to perform that task, but would instead include that as a function within the new proposed cohort characterization capability.

chrisknoll:
Thanks @pbr6cornell for the clarification.

@gowthamrao , I think it might be a bit complicated to try to make a generic erify query builder function, but if you feel up to the task, go for it.  maybe for this PR, we could just inline the era-ify query just to get this much needed feature added to the builder, and later we can refactor that query into some other builder that could be reused elsewhere.  Up to you.

One of the parameters to the era-ify logic is the gap that is allowed between timespans that would allow you to merge 2 separate eras together.  I'm thinking to default to 0 but provide a UI element to let them specify a number of days..

gowthamrao:
@chrisknoll  thanks. Yes - let us give it a try. We will make a PR in a few days. Will also incorporate an idea for gaps.

gowthamrao:
@chrisknoll  what do you think?

chrisknoll:
I think it's much cleaner.  Let me try to run it locally to do a few tests.  I think later we can optimize out one of the temp table creations, but for now this will work.

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
```diff
@@ -0,0 +1,74 @@
...
+		, dense_rank() over(order by @eraGroup) as groupid
+	FROM #era_constructor_input as so
+)
+,
+--------------------------------------------------------------------------------------------------------------
+cteEndDates (groupid, end_date) AS -- the magic
+(	
+	SELECT
+		groupid
+		, DATEADD(day,@eraconstructorpad,end_date) as end_date
```
chrisknoll: This is supposed to un-pad the end date from the inner query that padded the end day out by @eraconstructorpad.  This should be changed to:  

```
DATEADD(day,-1 * @eraconstructorpad,end_date) as end_date
```

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
```diff
@@ -0,0 +1,74 @@
+-- era constructor
+WITH cteSource (@eraGroup, start_date, end_date, groupid) AS
```
chrisknoll: I'm looking at the reference era-builder logic, and I don't see where groupID comes into play in the logic.  dense_rank() may lead to duplicate group IDs, so not sure what it is doing here...

chrisknoll: Ok, I think I understand:  groupID becomes a surrogate key for whatever the @eraGroup defines as the partitioning key for the data to be era-fyed.  In the case from webapi, @eraGroup is 'person_id' and so each person will be assigned their own ID, dense_rank() is required to give the same person_id the same number, and to eliminate gaps we'll just dense_rank() it.  later we can use the groupId to partition when we find the era end dates.  Makes sense, I think it will work...

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
gowthamrao: @chrisknoll  you are correct - fixed.

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
gowthamrao: Correct - dense_rank() will be one column that corresponds to an arbitrary number of columns that we are building an era around. This makes your original era-fy code genralizable and encapsulated, so may be reused anywhere else.

src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
```diff
@@ -56,18 +56,36 @@ from #included_events;
...
 @censoringInserts
 
+
+
+with era_constructor_input (person_id, start_date, end_date) as
+(
+	select F.person_id, F.start_date, F.end_date
+	FROM (
+	  select I.event_id, I.person_id, I.start_date, E.end_date, row_number() over (partition by I.person_id, I.event_id order by E.end_date) as ordinal 
+	  from included_events I
```
chrisknoll: When I ran this, included_events was not found, it shoudl be #included_events.

src/main/resources/resources/cohortdefinition/sql/generateCohort.sql
```diff
@@ -56,18 +56,36 @@ from #included_events;
...
+	  join cohort_ends E on I.event_id = E.event_id and I.person_id = E.person_id and E.end_date >= I.start_date
```
chrisknoll: Also, cohort_ends was not found, should be #cohort_ends.

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
```diff
@@ -0,0 +1,74 @@
...
+from
+(
+	select distinct @eraGroup , min(b.start_date) as start_date, b.end_date
+	from
+		(select distinct @eraGroup, groupid from cteSource) as a
+	inner join
+		cteEnds as b
+	on a.groupid = b.groupid
+	group by @eraGroup, end_date
+)
```
chrisknoll: This is invalid SQL synatx (at least on MSSQL), the subequery needs to be aliased ie:
select ... FROM ( subquery) Q -- Q is an alias

src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql
```diff
@@ -0,0 +1,74 @@
...
+		, dense_rank() over(order by @eraGroup) as groupid
+	FROM #era_constructor_input as so
+)
+,
+--------------------------------------------------------------------------------------------------------------
+cteEndDates (groupid, end_date) AS -- the magic
+(	
+	SELECT
+		groupid
+		, DATEADD(day,-1 * @eraconstructorpad,end_date) as end_date
```
chrisknoll: error during execution: end_date is not defined

This DATEADD() is unpadding the event_date that results in an era end (WHERE (2 * e.start_ordinal) - e.overall_ord = 0), so this should be:

```
DATEADD(day, -1 * @eraconstructorpad, event_date) as end_date
```

chrisknoll:
Hi, @gowthamrao , the temp table (adding # to the name) and padding fixes looks good. only thing that remanes is that we need to always alias the subquery (found in: src/main/resources/resources/cohortdefinition/sql/eraConstructor.sql line 73).  Just add any alias name to it like Q or something: FROM ( subquery) Q

chrisknoll:
Ok, @gowthamrao , latest push fixes the subquery alias, last fix I could find was related to the erify query.

This:

```
cteEndDates (groupid, end_date) AS -- the magic
(	
	SELECT
		groupid
		, DATEADD(day,-1 * @eraconstructorpad,end_date)  as end_date
	FROM
```

should be changed to:

```
	SELECT
		groupid
		, DATEADD(day,-1 * @eraconstructorpad, event_date)  as end_date
	FROM
```

After this, the query will be correct, and then we can turn our attention to the EraConstructor class:  This class (if we want to rename it to EraConstructorSettings, that might be more appropriate) should contain the parameters that the era construction logic needs. Note: this isn't the eraGroup values, but the padding value that the user can select.  Currenctly, EraConstructor contains a property 'constructor' of type 'CollapseType'. I don't think this is necessary.  Insead, EraConttructor (or EraConstructorSettings) should have a property of type int called 'pad' (or something pad-like named) can it can default to 0.  

The UI will need to be altered to prsent the UI elements to let the user set the padding.  but we need to define the place to store the user selected value.

gowthamrao:
@chrisknoll  thanks. I would like to retain the property 'constructor' of type 'CollapseType' to allow for any future (or Forked - custom) 'CollapseType' that a local implementation may want. e.g. 

[Default] era-fy collapse strategy
**[Option] - no collapse strategy. Presents records as is.
[Option] - collapse by de-duplicating records. de-duplicate records with the same subject-id, start date and end date
[Option] - max record.**

chrisknoll:
Ok, since there's only one option for now, we can just have the UI prompt for the padding and when additional options come up, we'll extend the UI to give the users a choice (but no point in making the users see choices where they have no choice at the moment). 

The one thing missing from your latest commit is to use the padding value that is in the EraConstuctorSettings padding value to set the era padding in the constructor's query:

Line 275 of CohortExpressionQueryBuilder should change from:

```
	query = StringUtils.replace(query, "@eraconstructorpad", Integer.toString(0));
```

To:

```
	query = StringUtils.replace(query, "@eraconstructorpad", eraConstructorSettings.eraPad );
```

chrisknoll:
@gowthamrao , Sorry for the back and forth on the class names and property names:

I misunderstood that we were going to try to support different collapse strategies, I thought (from the examples above with all the options) that the conclusion was 'always erify' because the other options could lead to invalidly constructed cohort records, such as:
[Option] - no collapse strategy. Presents records as is.

Which I said:
> Option 3 & 4: 
> (Option) - return all records - no changes (its the current version)
> (Option) - return distinct records - add distinct
> 
> This will create invalid cohort eras as described above:  people could have duplicate presence within the same period of time, thus resulting in double counting.
> 

But I don't want to quash any enthusiasm towards making this much needed feature, so if we want to leave the door open to different collapse'modes', then The class shouldn't be called 'EraConstructorSettings' (because there's different types of collapsing) maybe it shoudl be called 'CollapseSettings', it has a property called 'collapseType' (defaults to 'Era') and then the CollapseSettings will be a generic container for all the other types of collapsing (so it contains a padding property which is used when the type is 'Era' but if we have another type 'SomeOtherType', if it needs other fields, we'll just add it to the CollapseSettings object to store the user selections.

But, just one last pushback on the different types of collapsing:  I feel like era-ifying is the only valid thing to do with the cohort events.  It guarantees that the person will only appear at most 1 time in a give timewindow.  The other options may lead to this invalid state and I'm suggesting we avoid that.

gowthamrao:
You are correct about the push back - and I agree with the cohorts being invalid.

We allow for some non recommended stuff..like non standard vocabulary, and highlight that with red color. Maybe in UI we make other non recommended options RED

gowthamrao:
@chrisknoll changed naming to collapse instead of era - to allow generalization

chrisknoll:
Ok, @gowthamrao , this latest build looks good. No errors when running, and the inserted records look good.  Change to a 'collapse' mode instead of 'era' mode makes sense.  I was thinking about another mode that would work which is a 'min/max' mode that takes the minimum of all starts and the max of all ends, and makes that the cohort window for a person like this:

```
|-----|
              |-----|
should become:
|-------------------|
```

To demonstrate the collapsing: I made a 'Acute Influenza' cohort here: http://www.ohdsi.org/web/atlas/#/cohortdefinition/921492 and what it does is finds diagnosed influenza and used a fixed cohort end date as 30d after diagnosis (people are in the 'influenza' cohort from diagnosis to 30d after diagnosis.  The collapse logic works great, there's some people with just 1 influenza cohort window, others with multiples, and as soon as the UI supports it, I could play with the gap window to make it so that influenza that is re-diagnosed within 30d is considered the same era.

So, the next step is to merge it into master.  Just so you understand: when we merge, we're going to take all the individual commits made in this PR on your master branch, squash them into a single commit that describes the changes, and then merge into master.  You're going to have an issue on your fork because you commited your changes into your master branch, and the commits that will be merged into WebAPI's master are going to be squashed.  Normally this isn't an issue because we usually take pull requests off of a fork's branch (some branch other than master), but in your case, you have your changes on master.  What this means for you is: after we merge this into OHDSI/WebAPI, you may have to re-fork your repo off of webapi (delete and recreate) and then you should make future contributions off of a branch off of master.  Details of this process can be found here: http://www.ohdsi.org/web/wiki/doku.php?id=development:ohdsi_github_projects_v2.  

The other thing you should know is that we're working on refactoring the cohort definition code into a stand-alone library called 'Circe-be' (be for 'back-end') which is in the process of being incorporated into webAPI, but I'll take the steps to make sure your code changes get migrated into the stand-alone library.  But in the future, you should look to OHDSI/circe-be for the future development of the cohort definition expressions.

gowthamrao:
Awesome!! @chrisknoll  thank you!!!

PR #238, opened 2017-08-22
TxPath-lite backend code added.

choochoo7:
Adds the backend code necessary for a functional TxPath-lite application.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/TxPathwaysService.java
+++ b/src/main/java/org/ohdsi/webapi/service/TxPathwaysService.java
@@ -1,0 +1,76 @@
+package org.ohdsi.webapi.service;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import org.ohdsi.sql.SqlRender;
+import org.ohdsi.sql.SqlTranslate;
+import org.ohdsi.webapi.helper.ResourceHelper;
+import org.ohdsi.webapi.person.CohortPerson;
+import org.ohdsi.webapi.person.PersonProfile;
+import org.ohdsi.webapi.source.Source;
+import org.ohdsi.webapi.txPathways.Pathway;
+import org.ohdsi.webapi.txPathways.TxPathways;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.stereotype.Component;
+
+
+
+@Path("{sourceKey}/txPathways/")
+@Component
+public class TxPathwaysService extends AbstractDaoService {
+  
+  @Path("{cohortId}/{conceptSetId}/")
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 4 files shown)


anthonysena:
Hi @choochoo7 - I believe this PR is related to the OHDSI forum thread: [here](http://forums.ohdsi.org/t/txpath-slow-execution/3222/2). 

I wanted to get your input on how to test these additions. I believe the testing path is: using the WebAPI/txpath-lite branch with the web application: https://github.com/OHDSI/txPath-lite. Do I have this right? Are there any tests you would want me to perform as part of this review?

src/main/java/org/ohdsi/webapi/service/TxPathwaysService.java
```diff
@@ -0,0 +1,76 @@
...
+public class TxPathwaysService extends AbstractDaoService {
+  
+  @Path("{cohortId}/{conceptSetId}/")
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  public TxPathways getTxPathways(@PathParam("sourceKey") String sourceKey, @PathParam("cohortId") String cohortId, @PathParam("conceptSetId") String conceptSetId) 
+  {
+	  final TxPathways txpath = new TxPathways();
+	  Source source = getSourceRepository().findBySourceKey(sourceKey);
+	  String tableQualifier1="mimic_v5";
```
anthonysena: In your code, tableQualifier1 and tableQualifier2 are hard-coded to values that appear to be specific to your environment. I'd recommend you refactor this code in the following ways:

1. Rename the variables in your Java function _tableQualifier1_ to _cdmSchema_ and _tableQualifier2_ to _resultsSchema_
2. Follow the pattern shown in this  [function](https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/service/CohortResultsService.java#L131) and assign the value returned from _source.getTableQualifier(SourceDaimon.DaimonType.Vocabulary)_ to the _cdmSchema_ variable. Do the same for resultsSchema using _source.getTableQualifier(SourceDaimon.DaimonType.Results)_.
3. Make the same name replacement for tableQualifier1 & 2 in your SQL file for consistency.

In the UI side, you should expose a drop down for the sources that are configured via WebAPI by inspecting this URL: http://localhost:8080/WebAPI/source/sources. You can then add only those sources that have both a Vocabulary & Results daimon configured as this is a dependency for your code to work. Furthermore, you can use the value from this drop down when calling the REST end-point as I see that data source is also hard-coded into the [UI code](https://github.com/OHDSI/txPath-lite/blob/master/assets/js/TxPathController.js#L84)

anthonysena: [Here](https://github.com/OHDSI/Atlas/blob/master/js/main.js#L158) is a reference to traversing sources and daimons from Atlas.

anthonysena:
Hi @choochoo7 or @jduke99 - just wanted to see if you had any questions about the feedback I left on this pull request? I'm happy to discuss if there are any questions on how to move this forward. Thanks!

anthonysena:
Closing this PR due to inactivity but will leave the branch in the repository in case we need to revisit this. Please note that #361 is aiming to bring in the Treatment Pathways functionality as well.

PR #245, opened 2017-09-07
Correlated subquery fix

gowthamrao:
Some DBMS especially MPP systems do not allow correlated subqueries in aggregates with group by. These queries need to be rewritten. Also, using correlated sub-query may not be considered best practice.  SqlRender cannot translate correlated subquery into a query without subquery.


chrisknoll:
Thanks for the fix. It looks like you also want to introduce a new formatting style for the selects.

chrisknoll:
Thanks for the update, we'll check it locally and report back.

chrisknoll:
@gowthamrao , we're pulling this in for the next release, so I grabbed this branch from the PR and pushed it into a local repo where I can test and fix that subquery alias. We will open up a new PR soon.

PR #247, opened 2017-09-21
Vulnerability fixes

pavgra:
Fixes for https://github.com/OHDSI/WebAPI/issues/244.

`whitelist` method is still in process of implementation, it will be finished while community is reviewing rest of PR.


fdefalco:
Thanks for all the work - I'll try updating / commenting in line for most things but these are comments that are broadly applicable. 

- You are currently using the deprecated version of the SqlTranslate function.  It would be worth updating to the new method signature which no longer requires the source dialect be specified.

pavgra:
@fdefalco, we've updated the PR

fdefalco:
I've pulled the fork from the Odysseus repo into a new branch WebAPI/odysseusinc-vulnerability-fixes.  I've finished a good amount of testing and things are working.  Will continue to review based on that branch and then we can PR when everything checks out.

Issue #236, opened 2017-08-14
Population Level Estimation enhancements


anthonysena:
Per #217, adding a new issue to dive deeper into this new feature

**A short description of the feature**
Extend the Atlas estimation specification feature to allow for the specification of multiple target/comparator/outcomes. Revise the JSON representation of the estimation specification produced in Atlas to match ones that can be consumed directly by CohortMethod (e.g. analysis specification and t/c/o specification). We'd also like to follow this approach for the PLP feature proposed in #224.

**The organization or developers that are involved in the implementation**
@fdefalco @schuemie

**Any anticipated impact to current WebAPI/Atlas functionality (if applicable)**
This will be a breaking change as estimation specification storage will likely change as part of this enhancement. We'll attempt to preserve backwards compatibility or provide a migration path for existing specifications.

anthonysena:
Closed via #620

PR #250, opened 2017-09-29
Restructure CCA to have 1:M relationships

anthonysena:
Per #217 and #236, refactoring the estimation components to handle specification more than 1 of any of the following: target/comparator pairs, outcomes and analyses. This PR does not completely satisfy the requirements of #236 so I plan to contribute additional commits to this branch before we merge into master. This is here to publish what has been done so far and solicit feedback.


anthonysena:
I'm going to remove this PR and feature branch since there are other features that need to be incorporated. Those will be detailed later.

PR #251, opened 2017-10-02
Datasources Expansion and Improved Speed

lhalvors:
Implements features requested in issue #227 
- adds new Frequency Distribution graph in DataSources
- improvement of DataSources performance, by using a concept_hierarchy table generated by Achilles

PR #252, opened 2017-10-03
Datasources Expansion and Improved Speed (#251)

anthonysena:
* Add frequency_distribution queries + fix Data Density
- Add frequencyDistribution queries for drug, measurement, observation, and procedure
- Correct two queries for Data Density to work with PostgreSQL
* Update cdmresults > sql > report files to improve performance
- update all treemap.sql files to use concept hierarchy tables rather than embedded queries/joins
- update all drilldown reports sql files to cast stratum_x to INT rather than concept ID's to VARCHAR
* Use single concept_hierarchy table for treemap hierarchy info
* Update frequency summary plots to match updated Achilles implementation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *application.properties
 .idea/
+.metadata/
 *.iml
 sandbox/
 /target/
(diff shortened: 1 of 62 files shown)


anthonysena:
Depends on changes that are under review in OHDSI/Achilles#219

Issue #259, opened 2017-10-11
Oracle script V1.0.7.0__alter_cohort_generation_info failed 


chen-regen:
Flyway is throwing errors on the oracle script V1.0.7.0__alter_cohort_generation_info.sql in our Oracle environment:
```
-----------------------------------------------------------
SQL State  : 42000
Error Code : 903
Message    : ORA-00903: invalid table name

Location   : db/migration/oracle/V1.0.7.0__alter_cohort_generation_info.sql (/var/lib/tomcat/webapps/WebAPI/WEB-INF/classes/db/migration/oracle/V1.0.7.0__alter_cohort_generation_info.sql)
```

chen-regen:
@anthonysena Hey Anthony, I have created PR: https://github.com/OHDSI/WebAPI/pull/260 for this.

PR #260, opened 2017-10-11
Fix issue #259

chen-regen:
@anthonysena Hi Anthony, I found this problem in our Oracle today. Please review. Thanks!


chrisknoll:
Thanks, @chen-regen , testing on oracle is a current gap we're trying to close, thank you for finding this and submitting a pr!

PR #264, opened 2017-11-10
Fixes OHDSI/Atlas#502

anthonysena:
Fixes OHDSI/Atlas#502 by changing the signature of the REST endpoint to take the concept name as a query string parameter.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
+++ b/src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
@@ -25,6 +25,7 @@ import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.DELETE;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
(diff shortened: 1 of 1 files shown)


chrisknoll:
Hey, @anthonysena , since we're chaging the rest endpoint anyways, is the id part of the rest endpoint necessary?  Wouldn't it be enough to just go to /conceptset/exists?name={somename}.

Also, on the client side, we just need to make sure we properly escape the querystring.

chrisknoll:
I see it, the id is meant to ignore a concept set by ID when checking for existing.  It's a little wierd because when you first create a concept set, you don't have an ID, but I guess you could pass up a -1 or something.  Also it's a little wierd because the REST path makes it look like you're navigating to a conceptSetId, and then performing an action in it, but in this case, the ID is not exactly that (especially in the case of the new concept set)....If I could change more things about this check, I think i'd check for a name already existing in the following cases:  
- New Concept Set
- Existing concept set but name has been changed.

chrisknoll:
@anthonysena, one more idea:  for purposes of keeping backwards compatablity, would it be possble to leave the current endpoint in place (maybe we can mark it depricated), and instead we add a new rest endpoint looking for the query string and we can change all our front end code to use the new reset endpoint, and in our next (major) release we can remove the depricated endpoint...

anthonysena:
Hey @chrisknoll - thanks for taking a look! I was torn on this so let me share my thought process and maybe there is a different approach we can take. 

Originally, the endpoint was: {id}/{name}/exists and the problem with this approach was attempting to properly escape special characters in the {name} portion since it is part of the URL path. Specifically, if the {name} contained a '/' character, it would not reach the end point. So, this forced me to rethink the end-point all together to resolve this issue. I had to break the backwards compatibility but in this case, it seemed liked the only path (terrible pun intended) forward. 

We can add a header to mark the old path as depreciated as described [here](https://stackoverflow.com/questions/13884141/convention-for-http-response-header-to-notify-clients-of-deprecated-api). I also agree that the convention of including the {id} in the path is also confusing. Perhaps we should consider making both of these parameters to the query string so that the end point was of the form:

/conceptset/exists?id={id}&name={name}

Let me know what you think of that proposal.

chrisknoll:
I like the idea of putting both id and name as query params.  This way, the id can be made optional (in the case where we're creating a new concept set and there's no existing ID.

PR #273, opened 2017-12-01
Vulnerability fixes

mpozhidaeva:


src/main/java/org/ohdsi/webapi/service/VocabularyService.java
```diff
@@ -109,13 +114,19 @@ private String getDefaultVocabularySourceKey()
   @Produces(MediaType.APPLICATION_JSON)
   @Consumes(MediaType.APPLICATION_JSON)
   public Collection<Concept> executeIdentifierLookup(@PathParam("sourceKey") String sourceKey, long[] identifiers) {
+    Collection<Concept> concepts = new ArrayList<>();
     if (identifiers.length == 0) {
-      return new ArrayList<>();
+      return concepts;
+    }
+    else if (driver.equals(SQLServerDriver.class.getCanonicalName()) || identifiers.length > 2000){
```
anthonysena: Is it possible to move this logic into the PreparedSqlRender class instead? Otherwise the caller needs to be responsible for splitting the parameters which should be a concern of the PreparedSqlRender in my opinion.

pavgra: @anthonysena, the only way to move the logic into PreparedStatementRenderer, which I see, is to generate SQL statement, which consists of several similar queries joined by UNION ALL. E.g.

SELECT * FROM table WHERE id IN (1, ..., 2000)
UNION ALL
SELECT * FROM table WHERE id IN (2000, ..., 4000)

Do you have anything against such option?

anthonysena: Thanks @pavgra and @chrisknoll for the discussion earlier offline. To recap: the limit on the number of parameters a prepared statements can utilize is something that calling functions will need to consider when creating a SQL statement using the new Prepared Statement approach. In the case of the vocabulary service, we'll want to see if we can refactor the SQL calls to eliminate the 'IN' clause or provide alternate REST endpoints to resolve identifiers, etc.

anthonysena:
Hi - I'm going to set up a new branch in WebAPI for testing these changes along with a few other open PRs before we merge these into master.

Issue #282, opened 2017-12-18
New Authentication providers


pavgra:
Authentication providers for LDAP, Active Directory, Kerberos, OpenID, Custom DB.

PR will be created as soon as https://github.com/OHDSI/WebAPI/issues/244 will be resolved (waiting for approval and merge).

anthonysena:
@pavgra as mentioned in #244, I've merged in the vulnerability changes into the rc_2.3 branch. For this issue and the others (#283, #284, #285), please rebase your changes off the rc_2.3 branch so that we can review and merge them with the rest of the work. Thanks!

anthonysena:
Closed

PR #287, opened 2017-12-19
New Authentication providers #282

mpozhidaeva:
For https://github.com/OHDSI/WebAPI/issues/282


src/main/resources/application.properties
```diff
@@ -26,16 +27,16 @@ r.serviceHost=${r.serviceHost}
 #Flyway database change management.
 #DataSource for Change Managment / Migration
 flyway.enabled=true
-flyway.datasource.driverClassName=${flyway.datasource.driverClassName}
-flyway.datasource.url=${flyway.datasource.url}
-flyway.datasource.username=${flyway.datasource.username}
-flyway.datasource.password=${flyway.datasource.password}
+flyway.datasource.driverClassName=${datasource.driverClassName}
```
chrisknoll: It makes sense that the flyway's driver and url match the datasource's URL (so those changes are good on line 29,30), the other change to re-use the datasource's username and password means that your default account runing the app has DDL administrative rights on the datasource, which is what the separation of the flyway user/password was intending to do.  We should keep the normal WebAPI datasource user seaparate from the flyway user to keep this protection in place.

src/main/resources/application.properties
```diff
@@ -26,16 +27,16 @@ r.serviceHost=${r.serviceHost}
...
+flyway.datasource.url=${datasource.url}
+flyway.datasource.username=${datasource.username}
+flyway.datasource.password=${datasource.password}
 # check that migration scripts location exists
 flyway.check-location=true
 flyway.locations=${flyway.locations}
 # locations of migrations scripts
 # schemas to manage/update (e.g. ohdsi/results schema) -NOTE: CASE SENSITIVE!
-flyway.schemas=${flyway.schemas}
+flyway.schemas=${datasource.ohdsi.schema}
```
chrisknoll: The flyway.schemas is a list of schemas (in the event we want to maintain different schemas for the webapi database.  The datasource.ohsi.schema is always a single schema.  So, you could change it so flyway.schemas is the datasource.ohdsi.schema, but these things represent different values.

chrisknoll:
Hey, @pavgra and @mpozhidaeva,
  This PR looks like it has a lot of valuable changes, but some of them I thought we already had the capability:  specifically windows authentication.  Could you briefly summarize how the previous implementation of windows authentication was swapped for the implementation in this PR?

Separately, I noticed that an option for a local DB storage was added, and in the past I've said that I didn't think we (the ohdsi devs) wanted to be on the hook for any sorts of security leaks or password hacks, so I recommended we only support those authentication mechanisms that depend on a third-party authentication provider (such as Ladap, OAUTH, kerberos, windows auth)....anything that would keep us from having to store passwords.  But, looks like we now have an option of storing passwords, and I've got a strong feeling that we should try to avoid that liability.

pavgra:
@chrisknoll,

First of all, thank you for your review.

Speaking of windows auth, the implementation, which has already existed in WebAPI, used the server domain (so the domain of the box where the WebAPI is deployed) to auth users across it and therefore was not working, when WebAPI was deployed at non-Windows OS, same as was not configurable in any way. Our implementations works over LDAP (and accepts connection details, which can be configured via params), so no problems occur when WebAPI is deployed at Unix box. Second `AD` provider was added, because Windows Active Directory LDAP protocol differs a bit from plain LDAP and although it is possible to use plain LDAP to login across Windows AD, it is not convenient - credentials should be passed in a form non-typical for win users.

As of auth across custom DB, I would point following:
- As you see, if user chooses this auth type, it is up to him, whether passwords are stored in WebAPI's DB or some other DB. We are not pushing anyone to store creds in the same WebAPI's DB. No migrations were added.
- The auth type accepts only encrypted passwords, it will not even work if passwords are stored in DB in a form of plain text.
- And anyway passwords are stored somewhere and such place will always be a place of interest for fraudsters. Most of public web services store passwords in their DBs.
What do you think of this?

chrisknoll:
Hey, Pavgra,
  Thanks for the comments.  Good call on the Windows AD / LDAP points.  Hadn't considered Windows auth on non-Windows OS.  

Re: custom DB: as long as the implementation requires some sort of encryption/salt to keep the password hard to brute-force, then that is better.  I'd still prefer 3rd party providers, but if we want to have a 'buyer beware' disclaimer that people take a risk when hosting their own credential store, people can take the responsibility for it.

-Chris

anthonysena:
@mpozhidaeva @pavgra - I'm unable to get this branch working locally and I suspect it is because of some of the new properties that are introduced in the pom.xml. There is a new profile called 'develop' which has new settings and I think that is causing my problem. Can you provide some guidance on how to configure WebAPI using the settings.xml for testing these changes?

mpozhidaeva:
@anthonysena, I've updated this PR and https://github.com/OHDSI/Atlas/pull/553.
In internal dev process we use the following auth/security configuration (for example let's choose DB auth provider):
 - Atlas: config/config.json -> `"userAuthenticationEnabled": true` and leave only "DB" in list of authProviders
 - WebApi: pom.xml -> "security.db.datasource" properties
 
```
<security.db.datasource.url>jdbc:postgresql://localhost:5432/db_name</security.db.datasource.url>
<security.db.datasource.driverClassName>org.postgresql.Driver</security.db.datasource.driverClassName>
<security.db.datasource.username>username</security.db.datasource.username>
<security.db.datasource.password>password</security.db.datasource.password>
<security.db.datasource.authenticationQuery>
<security.db.datasource.authenticationQuery>select password from ${security.db.datasource.schema}.users where email = ?</security.db.datasource.authenticationQuery>
```

After the first login don't forget to add the user to admin role `insert into ohdsi.sec_user_role (user_id, role_id) values(1000, 2);` to have access to the configuration of permissions and roles.

Also there are separate branches with auth-providers and permissions PRs' changes and I would like you to use them for testing (I think it will be more convenient because of the logical dependency of these features):
https://github.com/odysseusinc/WebAPI/tree/auth_and_permissions_features
https://github.com/odysseusinc/Atlas/tree/auth_and_permiss_features

anthonysena:
Hi @mpozhidaeva thanks for the changes and guidance around this PR - it is greatly appreciated. I've done some testing and one problem I'm experiencing is when attempting to use the Windows authentication. At least on my local machine, I am unable to get this to work with the code on this branch. However, the code on the rc_2.3 branch does work properly. Could you investigate this potential regression?

mpozhidaeva:
@anthonysena - in fact it is the bug in odysseusinc repository. Thank you a lot.
Fixed here https://github.com/OHDSI/Atlas/pull/561/files

anthonysena:
Closing as this work is now being tracked under OHDSI/WebAPI#312

Issue #283, opened 2017-12-18
Permission system improvements


pavgra:
Added filters for all RESTs, new roles.

PR will be created as soon as #244 will be resolved (waiting for approval and merge).

anthonysena:
Closed via #316

PR #288, opened 2017-12-19
Permission system improvements #283

mpozhidaeva:


mpozhidaeva:
Issue https://github.com/OHDSI/WebAPI/issues/283

src/main/resources/db/migration/sqlserver/V1.0.9.0__data-permissions.sql
```diff
@@ -0,0 +1,76 @@
+insert into sec_permission(id, value, description) values(34, 'conceptset:get', 'List of Concept Sets');
```
chrisknoll: These inserts should have the ${ohdsiSchema} placeholders to fully qualify the table being inserted into.

anthonysena:
Hi - I had to make a few small corrections to the migration scripts that are included in this PR. I've only done this for PostgreSQL as shown in this commit: https://github.com/OHDSI/WebAPI/commit/5a3f3051522f5fcf705358ff0ed6feaa5141d690.

A few notes on the changes I made:
- The migration script V1.0.0.9.2__shiro_security-permissions.sql appears to be redundant with the first few SQL lines in V1.0.10.0__data-atlas-user.sql since both are attempting to create a permission record for the `cohortdefinition:*:put` permission.
- I modified V1.0.10.0__data-atlas-user.sql to associate the newly created permission to the 'cohort creator' role. 
- I had to remove the `ON CONFLICT DO NOTHING` clauses as they were causing the script to fail on my local PostgreSQL install (PG v9.4). These upsert operations might not be fully supported across the community so I thought it best to remove them. 

A few other notes for others on what is changing here:
- More fine-grained permissions for the various features in Atlas/WebAPI including the IR, Estimation, Prediction modules. The full details can be found in this script: V1.0.9.0__data-permissions.sql.
- The role 'public' is removed and replaced with the role 'Atlas user'. 

@mpozhidaeva - if you are agreeable to the changes in the commit I referenced, would you be able to carry them through to Oracle & SQL Server scripts and add them to this PR? Then I think we'll be ready to merge this into the 2.3 branch. Thanks!

chrisknoll:
I've a little concern about a role being replaced.  I'm not sure it makes sense to have a 'Atlas user' role vs. a 'public' role.  It brings into question how separate WebAPI and Atlas should be (considering the likelyhood that other applications coudl be written to interact with WebAPI that is not 'Atlas')

anthonysena:
So the 'public' role as it stands today has no permissions assigned by default so I'm not sure why the role was removed as part of this PR. Maybe we should consider keeping the 'Atlas user' role and take out the portion that removes the 'public' role? 

I agree though that we need to think about some default permissions related to consumers of WebAPI as part of a broader community discussion. In the meantime though, perhaps @pavgra and @mpozhidaeva  can describe why these changes were needed from their perspective? There are a few security-related PRs between WebAPI and Atlas and I'm not entirely clear on their aims beyond expanding the base security model to secure the other modules that were not part of the initial security release.

chrisknoll:
Yeah I'm drawing a line between 'role' and permission.   We need to define the roles that exist in WebAPI so that we have some sort of expectation about some 'system level' roles.  External providers are free to add their own roles and assign permissions, but we just need to be careful about the 'system range' of roles and permissions (which i think start at 1 and go up to 100? Maybe 1000).  

As far as default permissions, makes sense to me to leave 'public' blank and let people apply their own permission schema based on their needs (by default to be safe, public had no perms).  Maybe we should create a 'permission pack' that sets permissions to a specific profile. but i don't think there's a one size fits all here.

I think there's  UI to manage permissions somewhere, right?  So I think you have to resort to ddl to add roles, but associating roles to permissions I think can be UI driven.

mpozhidaeva:
@anthonysena - thank you very much for helping and suggested changes!
I checked your diffs locally and also added some constraints of the admin role (now the admin permissions should be only for the Configuration tab).

mpozhidaeva:
Sorry, I need to update this pr for corresponding to Atlas permission pr. Please check last commit

src/main/java/org/ohdsi/webapi/shiro/SendTokenInHeaderFilter.java
```diff
@@ -1,25 +1,55 @@
 package org.ohdsi.webapi.shiro;
 
+import static org.ohdsi.webapi.shiro.management.AtlasSecurity.PERMISSIONS_ATTRIBUTE;
+import static org.ohdsi.webapi.shiro.management.AtlasSecurity.TOKEN_ATTRIBUTE;
+
+import com.google.gson.Gson;
```
anthonysena: This needs to be added to the pom.xml; the code will not compile otherwise.

mpozhidaeva: @anthonysena - thanks you. Fixed.

anthonysena:
Closing as this work is being tracked under OHDSI/WebAPI#312

Issue #285, opened 2017-12-18
Add REST endpoint, which generates DDL for Result schema


pavgra:
PR will be created as soon as #244 will be resolved (waiting for approval and merge).

anthonysena:
Closed via c92863dbdd5cec90c5e17c29368b91f9b703c6b9

PR #289, opened 2017-12-19
Add REST endpoint, which generates DDL for Result schema #285

mpozhidaeva:


mpozhidaeva:
Issue https://github.com/OHDSI/WebAPI/issues/285

src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
```diff
@@ -108,7 +111,17 @@ public JdbcTemplate getJdbcTemplate() {
...
 
   public JdbcTemplate getSourceJdbcTemplate(Source source) {
-    DriverManagerDataSource dataSource = new DriverManagerDataSource(source.getSourceConnection());
+    DriverManagerDataSource dataSource = new DriverManagerDataSource(source.getSourceConnection()){
+      @Override
+      protected Connection getConnectionFromDriverManager(String url, Properties props) throws SQLException {
+
+        Connection c = super.getConnectionFromDriverManager(url, props);
+        if (url.startsWith("jdbc:redshift")) {
+          c.setAutoCommit(false);
```
anthonysena: Can you explain why you are using the setAutoCommit option in the case of RedShift? Was there a specific issue that this fixed? This doesn't appear to be related to the overall content of this PR so was curious.

src/main/java/org/ohdsi/webapi/service/SqlRenderService.java
```diff
@@ -72,5 +85,34 @@ public TranslatedStatement translateSQL(SourceStatement sourceStatement) {
...
+        return sql.replaceAll(";", ";\n");
+    }
+
+    private String translateSqlFile(String resource, String target, String schema) throws IOException {
+
+        Resource cohortFeasibilitySql = applicationContext.getResource(resource);
+        SourceStatement statement = new SourceStatement();
+        statement.targetDialect = target;
+        try(InputStream in = new FileInputStream(cohortFeasibilitySql.getFile())) {
+            statement.sql = IOUtils.toString(in);
```
anthonysena: The .getFile() call fails in my environment. The calling function sets the parameter **resource** to a value such as:

`"classpath://db/cohort_results/cohort_feasiblity.sql"`
 
This then fails to resolve to the file path. Elsewhere in the code base, we use the ResourceHelper class to obtain the SQL from a .sql file. When testing this locally, I made the following change to this function:


```
    private String translateSqlFile(String resource, String target, String schema) throws IOException {

        //Resource cohortFeasibilitySql = applicationContext.getResource(resource);
        SourceStatement statement = new SourceStatement();
        statement.targetDialect = target;
	statement.sql = ResourceHelper.GetResourceAsString(resource);
        //try(InputStream in = new FileInputStream(cohortFeasibilitySql.getFile())) {
        //    statement.sql = IOUtils.toString(in);
        //}
        HashMap<String, String> parameters = new HashMap<>();
        parameters.put("results_schema", schema);
        statement.parameters = parameters;
        TranslatedStatement translatedStatement = translateSQL(statement);
        return translatedStatement.targetSQL;
    }

```

I then changed the calling function as shown:

```
    public String generateResultSQL(@PathParam("target") String target, @PathParam("schema") String schema) throws IOException {

        String sql = translateSqlFile("/db/cohort_results/cohort_feasiblity.sql", target, schema); //translateSqlFile("classpath://db/cohort_results/cohort_feasiblity.sql", target, schema);
        sql += "\n" + translateSqlFile("/db/cohort_results/cohort_features_results.sql", target, schema); //translateSqlFile("classpath://db/cohort_results/cohort_features_results.sql", target, schema);
        sql += "\n" + translateSqlFile("/db/cohort_results/ir_analysis.sql", target, schema); //translateSqlFile("classpath://db/cohort_results/ir_analysis.sql", target, schema);
        sql += "\n" + translateSqlFile("/db/cohort_results/createHeraclesTables.sql", target, schema); //translateSqlFile("classpath://db/cohort_results/createHeraclesTables.sql", target, schema);
        if (!REDSHIFT.equalsIgnoreCase(target)) {
            sql += "\n" + translateSqlFile("/db/cohort_results/heracles_indexes.sql", target, schema); //translateSqlFile("classpath://db/cohort_results/heracles_indexes.sql", target, schema);
        }
        return sql.replaceAll(";", ";\n");
    }

```
Then everything was working as expected. I've only left the comments in these functions to show you what has changed; feel free to remove them.  If you agree, would you make these changes in the code and push a new commit? Thanks!

src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
mpozhidaeva: There was some trouble with old version of driver - but anyway it's unnecessary code now (after updating driver version).
Thank you for question!

src/main/java/org/ohdsi/webapi/service/SqlRenderService.java
mpozhidaeva: In fact your proposed code looks better - thanks!
Updated.

anthonysena:
Hi @mpozhidaeva - Thanks for addressing the comments - I'm now able to get this branch running properly. In digging into more details, I had a few more comments:

- We'll need to add into the results schema the following tables:
   - feas_study_inclusion_stats
   - feas_study_index_stats
   - feas_study_result

I'd suggest adding these tables to src/main/resources/db/cohort_results/cohort_feasiblity.sql. The structure of these tables can be found here:

https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/postgresql/V1.0.0.5__feasability_tables.sql#L41

I'd like to also suggest that we create a new path for this REST endpoint; having it under the SqlRender service doesn't feel appropriate. I'd suggest creating a new class org.ohdsi.webapi.service.DDLService.java and have that support the following calls:

`<server>/WebAPI/ddl/results` ==> This returns the token-ized OHDSI-SQL 
`<server>/WebAPI/ddl/results?dialect=<dialect>&schema=<schema>` ==> This returns the SQL translated to the `<dialect>` specified with the `<schema>` substitution. Note that this is using the QueryString instead of the URL path in order to ensure that we can support any special characters.

Please reach out if you have any questions or concerns.

mpozhidaeva:
@anthonysena - updated

chrisknoll:
This is a nitpick, and feel free to ignore:  the resource fodler name is 'db', maybe that could be 'ddl'.  Also, it seems like the folder structure is on a per-feature basis with only the tables related to that feature inside.  For simplicity (and perhaps a way to avoid table name conflicts), could we set up the folder structure as:

/ddl
   /results
      table1.sql
      table2.sql
   /someOtherSchema
      otherSchemaTable1.sql
      otherSchemaTable2.sql

So the ddl is organized by schema, allowing us to see all the tables defined in a schema in a single folder view.  

Again, I can understand the grouping by feature, that lets you know which tables use which features, which is also a fine idea, I was just throwing out an alternative.  You guys are doing the work, so it's your prerogative on how you want to ultimately implement.

pavgra:
@chrisknoll, we will be happy to continue improving the feature, just wanted to keep consistent and go step-by-step: merge existing things first and then work on their improvements, especially keeping in mind how much code from our team was not merged yet. Will this work for you?

chrisknoll:
Fine with me. We'll revisit this PR once the other items are closed out.  Thanks!

Issue #284, opened 2017-12-18
Fix for Cohort Characterization results export


pavgra:
E.g. http://api.ohdsi.org/WebAPI/cohortresults/CS1/3288/export.zip gives exception.

PR will be created as soon as #244 will be resolved (waiting for approval and merge).

anthonysena:
Closed via #290

PR #290, opened 2017-12-19
Fix for Cohort Characterization results export #284

mpozhidaeva:


mpozhidaeva:
Issue https://github.com/OHDSI/WebAPI/issues/284

anthonysena:
@mpozhidaeva and @pavgra - unfortunately this change breaks compatibility for SQL Server as it does not support the LIMIT statement. 

Upon further review, the new feature_integration branch on Atlas (the branch that work with the WebAPI rc_2.3 branch) no longer features the export button so I'd propose that we remove this end-point or possibly throw an exception to tell people that the service is no longer available.

Tagging @fdefalco, @chrisknoll for awareness and input.

anthonysena:
Hi - I'm going to close this as @chrisknoll provided a fix to address this per https://github.com/OHDSI/WebAPI/commit/846383ed110f8dae160a6b77ad372dd1a6b07b88. Thanks!

Issue #305, opened 2018-01-25
Allow change of source_daimon priorities via UI


mpozhidaeva:
PRs:
https://github.com/OHDSI/Atlas/pull/541
https://github.com/OHDSI/WebAPI/pull/306

anthonysena:
Closed via #306

PR #306, opened 2018-01-25
Allow change of source_daimon priorities via UI

mpozhidaeva:
Issue #305


src/main/resources/db/migration/postgresql/V2.2.0.20180118105000__change-daimons-priority-permission.sql
```diff
@@ -0,0 +1,14 @@
+INSERT INTO ${ohdsiSchema}.sec_permission (id, value, description)
+  SELECT
+    nextval('${ohdsiSchema}.sec_permission_id_seq'),
```
anthonysena: sec_permission_id_seq does not exist for PostgreSQL - this should be:

`nextval('${ohdsiSchema}.sec_permission_sequence'),`

mpozhidaeva: I guess sec_permission_id_seq was added https://github.com/OHDSI/WebAPI/pull/288/files#diff-5f6ed586505c60e568f4d976bc903913

chrisknoll: Looks like this PR has a dependency on #288.

anthonysena: Thanks for clarifying the dependency - I'll take a look at #288 and then maybe that will address this comment.

anthonysena:
So, in order for this PR to move forward, we need to decide what we want to do on OHDSI/Atlas#541. As it stands, this PR would enable WebAPI to change a daimon priority for the whole system when security is disabled which is not desired. This PR does support locking down this functionality when security is enabled and only allows someone with the admin role to change it so I support that functionality. 

Perhaps the fast path forward is to simply disable this functionality if security is disabled?

pavgra:
@anthonysena,

If such approach is OK for you, then we will adopt the PR in such way
> Perhaps the fast path forward is to simply disable this functionality if security is disabled?

anthonysena:
@pavgra - yes, I think the easiest path would be to ensure that if a user is not logged in, we allow them to change the daimon for their session as is done today.

As I've been reviewing this work and some of the other PRs from @mpozhidaeva, I'm having some challenges in reviewing these changes because they all seem to have dependencies. Furthermore, when I've attempted to test these individual PRs, I've run into issues (compile and run-time) and when looking at the total sum of these PRs, it is probably easier to consider all the changes together. I'd like to propose a path forward and would appreciate your feedback:

- I am planning to do a merge of the rc_2.3 branch in WebAPI into master and the feature-integration branch in Atlas. We need to do this since those branches have a lot of functionality that have been tested at Janssen and other collaborators should be able to branch off of these new features from master when making changes.
- I think we should combine the following WebAPI PRs into a single PR to make review easier: OHDSI/WebAPI#287, OHDSI/WebAPI#288, OHDSI/WebAPI#306. The same goes for Atlas: OHDSI/Atlas#541, OHDSI/Atlas#545, OHDSI/Atlas#552, OHDSI/Atlas#553.
- If you combine those in your fork and rebase off master (once the other changes are merged) and then resubmit the consolidated PRs, I'll then do another review.

Appreciate your time and patience here with these PRs - I'm looking forward to getting these contributions into the code base.

anthonysena:
@mpozhidaeva - can you merge the contents of this PR into OHDSI/WebAPI#312 for review? Since this has dependencies on the security subsystem, I'd like to test this together.

pavgra:
@anthonysena,

> change a daimon priority for the whole system when security is disabled which is not desired
this issue we've fixed, but currently are a bit out of resources to work on implementing synchronous ability for admin to change portal-wide setting and regular user to change his local ones

anthonysena:
I'm wondering if it would be easier to rebase these changes off the most recent version of WebAPI's master branch and submit that along with the changes that are forthcoming for OHDSI/Atlas#57? Please let me know your thoughts and perhaps we can close this one.

pavgra:
Was closed as part of https://github.com/OHDSI/WebAPI/pull/374

PR #312, opened 2018-02-06
Auth providers and permissions

mpozhidaeva:
Merged changes of the issues:
https://github.com/OHDSI/WebAPI/issues/282
https://github.com/OHDSI/WebAPI/issues/283.


mpozhidaeva:
I've squashed the features' commits but  there are several not affecting commits from odysseusinc fork - I hope that's all right.
Please could you use 'squash and merge' too after testing/reviewing changes?
Thank a lot

pom.xml
```diff
@@ -511,13 +569,50 @@
...
         <hibernate.dialect>org.hibernate.dialect.SQLServer2012Dialect</hibernate.dialect>
+        <security.db.datasource.url>${datasource.url}</security.db.datasource.url>
+        <security.db.datasource.driverClassName>${datasource.driverClassName}</security.db.datasource.driverClassName>
+        <security.db.datasource.username>${datasource.username}</security.db.datasource.username>
+        <security.db.datasource.password>${datasource.password}</security.db.datasource.password>
+        <security.db.datasource.authenticationQuery>select password from ${security.db.datasource.schema}.user where \
+          login = ?</security.db.datasource.authenticationQuery>
       </properties>
     </profile>
+    <profile>
```
chrisknoll: Can we remove this develop profile?  If you have any local settings, can you override them in settings.xml?

pom.xml
```diff
@@ -494,12 +547,17 @@
         <datasource.password>PASS</datasource.password>
         <datasource.dialect>postgresql</datasource.dialect>
         <datasource.ohdsi.schema>ohdsi</datasource.ohdsi.schema>
-        <flyway.datasource.driverClassName>${datasource.driverClassName}</flyway.datasource.driverClassName>
```
chrisknoll: Could you explain why these settings (flyway.*) have been removed from the pom?

mpozhidaeva: It seems we tried to avoid duplicating the "flyway.datasource" properties, but for greater clarity I returned them.

anthonysena:
@mpozhidaeva thanks for combining these changes into a single PR. I've been able to get through the review now. Thank you! Here are a few items that we'll need to address:

1. src/main/resources/db/migration/postgresql/V1.0.9.0__data-permissions.sql - please remove the "ON CONFLICT DO NOTHING;" clause. 
2. Let's keep the public role for now. So, please remove the following Flyway script from each DB provider: src/main/resources/db/migration/<db_provider>/V1.0.10.1__data-public-permissions.sql

One question: 

- I noticed the introduction of the HikariCP library and it looks like it is used when managing user credentials in the database. There are also new connection pool settings added to the src/main/resources/application.properties to configure this component. I'm not familiar with this library and was curious to the need here? Here is a reference I found that others might be interested in: http://juliuskrah.com/tutorial/2017/02/16/getting-started-with-hikaricp-hibernate-and-jpa/

And a note for myself as we approach the release:

- This new set of changes allows you to control CORS enablement through configuration - this needs to be added to the documentation. Here is the setting:

` <security.cors.enabled>true</security.cors.enabled>`

pavgra:
@anthonysena, the case, which pushed us to add Hikari, was following:

- we implemented auth via custom DB
- this separate DB can go down and then be restored (e.g. in our situation it happened every time when Jenkins redeployed our Arachne, while Arachne's users table was used by Atlas as users source)
- default connection pool cannot re-initialize the connection and fires exception, but Hikari supports auto re-init of a connection

anthonysena:
Thanks @pavgra - this is helpful to understand. Appreciate it. Along these lines: can you or @mpozhidaeva provide some guidance how to test authentication against a custom DB? And how passwords are encrypted as part of this process?

pavgra:
@anthonysena,  we've implemented support of passwords encrypted with:

```
org.springframework.security.crypto.bcrypt BCryptPasswordEncoder passwordEncoder = new BCryptPasswordEncoder();
String encodedPass = passwordEncoder.encode(password);
```

mpozhidaeva:
@anthonysena - please pay attention to https://github.com/OHDSI/WebAPI/pull/287#issuecomment-363085666. Is this comment enough or required new clarifications?
BR

chrisknoll:
The BCryptPasswordEncoder is a good choice (read about it here: https://docs.spring.io/spring-security/site/docs/4.0.4.RELEASE/apidocs/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.html).  The default 'strength' of 10 seems reasonable to prevent brute force. However, I have a caution: if we use the default as of the 4.0.4 release, and in later releases the strength is changed to a different parameter, I believe this will break password encryption (asuming you get a different hashed password based on the different levels of 'strength').

In short: should strength be explicitly specified in the constructor?

anthonysena:
Hi - closing this out so we can track the work under a branch that is now shared between our teams. Please see #316 for all work to date. I've also attempted to link back to any open questions/issues from the previous PRs on this topic. Thanks!

PR #314, opened 2018-02-11
Refactor DDL files for DDLService

chrisknoll:
This PR refactors the files introduced in #289.  This layout has some advantages:

1. Looking at the file system, people can see all the tables defined in the schema without having to browse through the files.
2. Now that the tables are in separate files, we can see how the individual table changes over time

Also fixed up the DDL service so that all the assets are in lists.


mpozhidaeva:
@chrisknoll  - thank you very much for refactoring! 
I would like to note that the insertion commands are not ddl - but init_heracles_analysis.sql is in "ddl" folder.
Also I have a question about indents in the repository - spaces or tabs are more preferable?

chrisknoll:
RE: tabs vs. spaces: tabs won, so we use tabs.

I understand about the init() being in the DDL folder.  You're right it's not data-definition-language that is creating tables, but I wasn't sure where that should go.  Should we support another query param that includes the init code?  Do you think another endpoint is required, one for ddl one for initialization? I can put the init code under a sub folder so that it's located in ddl/schema/init...I believe there is a tight coupling between the ddl that creates the tables and the init() code that populates it.

Thank you for talking this through, it's a good point to consider.

mpozhidaeva:
@chrisknoll - ok, let's leave the existing changes

chrisknoll:
Thanks for the quick turnaround!

chrisknoll:
@mpozhidaeva ,

In the future, could you use the 'rebase and merge' option when possible?  If you see multiple commits that are little typo corrections, you can use 'squash and merge', but if the commits are clean and straightfoward, 'rebase and merge' is preferred.

Thanks!

PR #316, opened 2018-02-12
Odysseusinc security improvements

anthonysena:
Continuation of work from #312, #287: Merge of changes from the issues #282, #283. In short, adding additional authentication providers to WebAPI and extending the security permissions and roles to encompass the features added to Atlas after the Shiro security was introduced to WebAPI.

Open items: 

- [x] DB for authentication needs to be tested. Per @mpozhidaeva, need to configure WebAPI based on the guidance here: https://github.com/OHDSI/WebAPI/pull/287#issuecomment-363085666. 
- [x] Per @chrisknoll: https://github.com/OHDSI/WebAPI/pull/312#issuecomment-364511689 - should we explictly set the encryption strength?

### DIFF
--- a/src/main/java/org/ohdsi/webapi/shiro/JdbcAuthRealm.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/JdbcAuthRealm.java
@@ -1,0 +1,115 @@
+/*
+ *
+ * Copyright 2017 Observational Health Data Sciences and Informatics
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Authors: Mikhail Mironov, Pavel Grafkin
+ *
+ */
+package org.ohdsi.webapi.shiro;
+
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.shiro.authc.AccountException;
+import org.apache.shiro.authc.AuthenticationException;
+import org.apache.shiro.authc.AuthenticationInfo;
+import org.apache.shiro.authc.AuthenticationToken;
+import org.apache.shiro.authc.SimpleAuthenticationInfo;
+import org.apache.shiro.authc.UsernamePasswordToken;
+import org.apache.shiro.authz.AuthorizationInfo;
+import org.apache.shiro.authz.SimpleAuthorizationInfo;
+import org.apache.shiro.realm.jdbc.JdbcRealm;
+import org.apache.shiro.subject.PrincipalCollection;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public class JdbcAuthRealm extends JdbcRealm {
+
+    private static final Log log = LogFactory.getLog(JdbcAuthRealm.class);
+    private BCryptPasswordEncoder bCryptPasswordEncoder = new BCryptPasswordEncoder();
+
+    public JdbcAuthRealm(DataSource dataSource, String authenticationQuery) {
+
+        setAuthenticationTokenClass(UsernamePasswordToken.class);
+
+        this.setDataSource(dataSource);
+        this.setAuthenticationQuery(authenticationQuery);
+    }
+
+    @Override
+    protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken token) throws AuthenticationException {
+        SimpleAuthenticationInfo info;
+
+        UsernamePasswordToken upToken = (UsernamePasswordToken) token;
+        String username = upToken.getUsername();
+
+        if (username == null) {
+            throw new AccountException("Null usernames are not allowed by this realm.");
+        } else {
+
+            String password = this.getPasswordForUser(username);
+
+            if (password == null ||
+                    !bCryptPasswordEncoder.matches(new String(upToken.getPassword()), password)) {
+                throw new AuthenticationException("Incorrect username or password");
+            } else {
+                info = new SimpleAuthenticationInfo(username, upToken.getPassword(), this.getName());
+            }
+        }
+        return info;
+    }
+
+    @Override
+    protected AuthorizationInfo doGetAuthorizationInfo(PrincipalCollection principals) {
+
+        // Prevent from querying of permissions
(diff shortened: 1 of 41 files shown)


src/main/java/org/ohdsi/webapi/shiro/JdbcAuthRealm.java
```diff
@@ -0,0 +1,115 @@
...
+        String username = upToken.getUsername();
+
+        if (username == null) {
+            throw new AccountException("Null usernames are not allowed by this realm.");
+        } else {
+
+            String password = this.getPasswordForUser(username);
+
+            if (password == null ||
+                    !bCryptPasswordEncoder.matches(new String(upToken.getPassword()), password)) {
```
anthonysena: @mpozhidaeva - I configured my local WebAPI per https://github.com/OHDSI/WebAPI/pull/287#issuecomment-363085666. In reviewing the pom.xml, I made some adjustments here: https://github.com/OHDSI/WebAPI/pull/316/commits/63c1471ebc23134354889e1a6b1bb3e20fcb0b65. This way we're using the default OHDSI schema and sec_user table found in WebAPI.

The code I've noted above is causing me an issue in that: I am unsure how to encrypt my password to allow this functionality to work. What do you propose for people to use if they'd like to configure their DB for authentication? Is there a command-line interface or some other utility that you have for creating users?

chrisknoll:
General comment about the settings in pom.xml:  the defaults seem to point to the WebAPI tables for storing passwords, but my understanding about the database authenticaton mode was that you'd provide an external database with the user/password encrypted in it and pom.xml would contain the connection settings to connect to that external database and query for the hashed password.  I was surprised when @anthonysena pointed out that the sec_user table actually has a password field in it!  

So, should we put in the pom.xml some default settings that guide people to point to an external database and perhaps part of this PR drop the password fields form the WebAPI tables? I need to check with  @ganisimov about the origional implementation of the sec_user table...

@ganisimov: is there some place in the initial shiro implementation where we'd write to a password table in sec_user?

chrisknoll:
Hi, spoke with @ganisimov:

> I believe it’s safe to drop ‘password’ field form SEC_USER. It also contains ‘salt’ field, which is not used and can be dropped.
> 
> You’re right, with external auth mechanism we don’t need these columns. As far as I remember, they were introduced by GMalikov and I decided to keep them to be able to add basic or digest authentication, if needed.
> 

So, I propose that we remove password and salt columns, and also it would be a good idea to put together a short wiki article about how to set up the database for authentication, and included details on how passwords needed to be hashed into the data.

pavgra:
@chrisknoll, there is no place to explicitly set number of rounds used for password's hash generation.
```
bCryptPasswordEncoder.matches(new String(upToken.getPassword()), hashedPassword)
```

And there is no need for it due to following reasons. First, a hash of password stores number of rounds used for its generation itself. Second, we use `BCryptPasswordEncoder`, which automatically retrieves used number of rounds from a given hash
`org\springframework\security\spring-security-crypto\4.2.3.RELEASE\spring-security-crypto-4.2.3.RELEASE.jar!\org\springframework\security\crypto\bcrypt\BCrypt.class, line 577`
```
rounds = Integer.parseInt(salt.substring(off, off + 2));
```
and uses it for hashing a given plain text password
`line 590`
```
hashed = B.crypt_raw(passwordb, saltb, rounds);
```
for further comparison of two equally hashed passwords.

@anthonysena, give a try to following service, it should work: 
https://www.dailycred.com/article/bcrypt-calculator

chrisknoll:
Hi, @pavgra , that was a good find.  There is a place to set the strength: https://docs.spring.io/spring-security/site/docs/4.0.4.RELEASE/apidocs/org/springframework/security/crypto/bcrypt/BCryptPasswordEncoder.html

like this:
```
bCryptPasswordEncoder = new BCryptPasswordEncoder(10) ;
```

However, you raise a good point (and good work finding it int he source code) that the 'strength' is hashed into the password encoding so that if a different strength is used later, legacy passwords will work! great!  Tho, I think it's a weakness of the hash to say exactly which strength was used in the password hash, but I understand the tradeoff.

Thank you for the clarification.  That looks good.

anthonysena:
Thanks @pavgra for the guidance here! I'll give that a shot on my side and verify that it works.

anthonysena:
Tested this all on my side and happy to report that everything is working well. Here are some notes that I'm capturing that we'll put into the overall release notes:

- Adds support for authentication providers for LDAP, Active Directory, Kerberos, OpenID, Custom DB. (#282)
   - The Custom DB option adds support for a separate, external database to house username/passwords.  Use the `security.db.datasource` configuration properties to configure the DB connection and query to retrieve the password for a user. The table which holds the credentials must encrypt the password using BCrypt. This website provides a mechanism to generate an encrypted password: https://www.dailycred.com/article/bcrypt-calculator.
  - We have dropped the passsword/salt columns that were part of the SEC_USER table in WebAPI. As a rule, we do not want to store credentials in the WebAPI database. This table will now hold the list of authenticated users and will be used to associate them to other entities in the WebAPI database.
- Extends roles to include new functionality added to Atlas since the introduction of the security module (#283).

SBylemans:
Why was the decision taken to remove permissions from the token? Is this not more secure as it is base 64 encoded?

The reason I ask is because when integrating with CAS SSO along with Atlas, we pass through _/welcome/:token:_. This will fill in the token but we lack the permissions at this point.

pavgra:
@SBylemans, each time you create an entity (concept set, cohort, etc), Atlas grants you permissions on the entity. That's why user's permissions list is growing quite fast and starts going beyond size limits allowed for headers by some web servers. And there is no such issue, when we pass the list in response body.

Speaking of security, first, permissions itself doesn't look like sensitive info (and there is anyway check at back-end side), second, if you worry about security, you may want to use SSL and then there will not be any issues

SBylemans:
@pavgra thanks for for the response, that indeed makes sense...

But this solution is not adequate for an SSO implementation or am I wrong? 

For an SSO environment you need to pass through the authentication before proceeding to the actual site. In this case we are not able to retrieve the permissions from the header because of several redirects... Are there any resolutions to this?

gowthamrao:
@chrisknoll @pavgra @anthonysens is this being worked on as part of 2.4 release? Or was removing the password from webapi already part of 2.3?

SBylemans:
@pavgra In an SSO environment, in my case CAS, you need to pass through the authentication mechanism before accessing the site behind the SSO. Now we configured the CAS client in WebAPI under _/user/oauth/cas_ so in order to redirect the user to OHDSI Atlas (from another site behind the SSO), we send the user to _/user/oauth/cas_  in order to authenticate him with CAS and now Atlas knows this as well. This will send him back to Atlas as an authenticated user. The problem is now the permissions will never be passed to Atlas or any other login method that does not use `ajax` calls. 

An easy way to solve this would be to create an endpoint to retrieve the permissions for such authentication methods.

chrisknoll:
Just a message from the future:   folding permissions into the header for authentication has caused major problems when the permission payload is large (which in cases exceeds 350KB).   If the problem was solved by retrieving permissions through an endpoint, we should separate authentication concerns (as in JWT tokens, session timeouts, SSO) from authorization concerns (now that i'm authenticated, what am I allowed to do?). 

It's a complicated lift because security is folded into every function, but these two things need to be isolated from each other, since it's now causing a very large impact on UI performance (taking 3-5 seconds to determine permissions **per request**).

Issue #315, opened 2018-02-12
Capture author for all entity types


pavgra:
Today WebAPI captures author only for Cohorts and Incidence Rates. The suggestion is log author for Concept sets, Estimations and Predictions too.

gowthamrao:
Thank you @pavgra 

This makes absolute sense. Infact, I think we should at one point plan to design and store user specific definitions and results. E.g. If user A has results, user B can still run incidence rates, but it will not overwrite user B work

anthonysena:
Closed via #320

PR #320, opened 2018-02-13
Capture author for all entity types

wivern:
for #315


anthonysena:
Note to test with https://github.com/OHDSI/Atlas/pull/566

src/main/resources/db/migration/postgresql/V2.2.5.20180212152023__concept-sets-author.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE ${ohdsiSchema}.concept_set
+  ADD COLUMN IF NOT EXISTS created_by VARCHAR(255),
```
anthonysena: While I like the check here, the `IF NOT EXISTS` is not compatible with PostgreSQL 9.4 or earlier. I'd recommend that we remove this clause from these flyway scripts so that we create the column. Also, can you add the migration scripts for SQL Server & Oracle as well?

Issue #322, opened 2018-02-14
Vocabulary schema is not supported


wivern:
Now it`s limited to utilize the same schema for cdm and vocabularies.

PR #323, opened 2018-02-14
Use Vocabulary schema in cohort generation

wivern:
for #322


chrisknoll:
Hi, @wivern , I have a PR open for circe-1.4.0 changes, I can take these changes and merge it into that one.

chrisknoll:
I'm adding null checks to your code, @wivern .  You could have a source set up with only a CDM daimon type, and your code assumes that you have both daimon and vocabulary.  So, we will check for the existance of a Vocabulary on the same source as the CDM, and if the vocab and CDM exist, we'll set both parameters. Otherwise, only cdm is set.

Issue #324, opened 2018-02-14
Sequences for entities


pavgra:
@chrisknoll, @anthonysena, could you help me to understand, why WebAPI uses a single sequence (Hibernate's default sequence) for all entities, not a sequence per entity. So, rephrasing, why @SequenceGenerator annotation is not set and configured for all identifiers columns (e.g. to use database's `concept_set_sequence` for `ConceptSet` entity)? I guess, there was a good reason for it, so would be grateful for your clarification.

pavgra:
Seems like the question was mentioned before - https://github.com/OHDSI/WebAPI/issues/187. So, @chrisknoll, would you mind, if we migrate all entities to their own sequences?

anthonysena:
I'm supportive of making the change to having a sequence per table versus using the hibernate default sequence.

chrisknoll:
I think all of the DBs we support for WebAPI supports the notion of sequences, so I'd rather go that route (it saves a db udpate call when we go to a sequencer vs. updating a table because the DB optimizes this type of operation).  There is a challenge:

> The challenge is creating the DDL for these sequencers that are initialized based on a value from a select max (id) from EntityTable. I'm sure this is possible on MSSQL, but more challenging on oracle.

If we do a migration for this, we need to create the migration script (one for each platform, mssql, oracle, postgresql) that can figure out the current max id for each table, and initialize a sequence to the  max(id) of the table in question.  Then the autonumber columns (in mssql case) needs to be dropped and replaced with just a simple int field...and there's primary keys and foreign keys invovled to make ti trick.  This is all technically possible, just a challenging task.  In my mind, the hardest part is getting this to work over in oracle world: on mssql i could declare a variable as the max ID of a table, and then create a sequence initializing it to the variable.  On oracle, can you do that? no idea.

Other option is to completely reset the clock with WebAPI 3.0 with brand new sequencers, and provide a way to migrate data from WebAPI 2.x to WebAPI 3.0 as a separate utility (ie: a CLI tool that takes a webAPI 2.0 path, a webapi 3.0 path, and given the identier, it loads from one webapiand saves as new to the other webapi).  Making it CLI means we could do it in a batch operation.

chrisknoll:
> So, rephrasing, why @sequencegenerator annotation is not set and configured for all identifiers columns (e.g. to use database's concept_set_sequence for ConceptSet entity)? 

[...]

Issue #301, opened 2018-01-16
WebAPI/Atlas Release v2.3.0 Planning


anthonysena:
Hi,

We are planning a v2.3.0 release of WebAPI/Atlas for 2/28/2018 and this thread is a call for contributions for features to include in the release.

In order to submit an idea for a feature, please respond to this thread with the following information by 1/23/2018 so that we can anticipate any potential challenges for incorporating the features:

- A short description of the feature
- The organization or developers that are involved in the implementation
- Any anticipated impact to current WebAPI/Atlas functionality (if applicable)
- After you have responded to this thread, please create a new issue so that the specific feature can be tracked in the V2.3.0 milestone.

Contributions are due 2/14/2018 for inclusion in the release, to be provided as a pull request. If you have work that is planned but cannot make it into this release timeline, please do post them up to raise community awareness and collaboration. 

Thank you!

pavgra:
- _Name:_ REST endpoint for configured and supported WebAPI features
- _Description:_ during last architecture call (11 Jan) there was a discussion, that it would be useful to have REST endpoint, which returns what features are supported by current instance of WebAPI. E.g. whether authentication is turned on / off, what authentication providers are configured and available, whether there are any Data sources with CDM v5.3 connected to the WebAPI and therefore whether it is reasonable to show payer_plan options in cohort builder (the ones which are only supported by v5.3 and at older models will just cause errors), etc. Because now we have to sync such things between back-end and front-end configuration, which is bad approach - no single source of truth.
- _Organization:_ Odysseus Data Services Inc.

anthonysena:
Thanks @pavgra! I'm interested in collaborating with you on this feature and have some ideas on how we might accomplish this via WebAPI. Would you mind creating a separate issue that we can use to work through the design? I know that others in our team would also be interested in contributing to the design.

anthonysena:
A new feature idea:

- Integrate WebAPI/Atlas with the new Common Evidence Model (CEM) data source for processing negative controls.

pavgra:
@anthonysena, as you requested, here is separate issue
https://github.com/OHDSI/WebAPI/issues/303

gowthamrao:
Interested in the following

1. Atlas support payer_plan_period https://github.com/OHDSI/circe-be/pull/11
2. Heracles to include cost analysis using FeatureExtraction 
3. Heracles to include record counts (e.g. number of visit_occurrence records, drug_exposure records)
[...]

PR #327, opened 2018-02-16
Sequences for entities

wivern:
for #324


chrisknoll:
This looks very impressive, and I like how you worked through all the 'initial values for sequences'.  The oracle code looks pretty complicated and I don't actually have any way of testing this out.

It appears that you put in the various checks for when max(id) might return null, which is good, this migration needs to work when the database is empty (such as in a completely new install).  Did you manage to test this migration on the different flavors of RDBMS?  We can test the migrations internally on MSSQL and Postgresql (@anthonysena : you have a local install of Postgresql with some cohort defs?) but oracle has always been a challenge. If you can confirm that this works for Oracle, we'll work out testing it on the other 2 platforms.

**Note to everyone: before running webapi with this PR in place, I strongly suggest making a backup of your database before launching.  The changes here are not easily rolled back, so your best option is a backup-restore if something doesn't work right.**

anthonysena:
I can test this out on PostgreSQL.

+1 for backing up your DB!

chrisknoll:
Ok, I'll check it on our internal SQL server.  @chen-regen : do you have an oracle instance you could test this with?

anthonysena:
@chrisknoll - This is my bug - sorry for that. I contributed that script as part of another PR. Let's change it since we should assume that a previous script created those columns and therefore the existence check is not required. 

Also, for all: I'm also finding some bugs with the PostgreSQL script for sequencing. I'm thinking it might be easier to merge this working into a branch on WebAPI so that we can all contribute fixes on it? I'll plan to do this unless I hear from folks not to do that.

chen-regen:
@chrisknoll @anthonysena  Hi Anthony and Chris, sorry I just noticed this as Anthony brought it to my attention. I am actually working on a solution for the sec_role_permission table by sequence. Regenstrief will be working on creating a second instance for our OHDSI suite. Now we only have one instance so it would be challenging and dangerous for me to test this big changes. When do we plan on getting this released? I will try to talk to our folks here and see if we can get a testing environment sooner...

anthonysena:
Hi @chen-regen - we aim to have the 2.3 release ready by 2/28 per #301. I'd agree that it is important to have a secondary environment for testing these changes. Just let us know the timing for your secondary environment and worst case, we can apply some patches as part of a 2.3.x release to address any issues found with Oracle.

chen-regen:
@anthonysena wow this is pretty intense in terms of timing. I doubt our infrustracture team will get the environment ready by then considering they are working on our new ARACHNE servers now. Well if I won't be able to test, we will have to apply patches as you said. Internally, we really would like to get the cohort analysis changes and security improvements deployed with 2.3 release...

PR #338, opened 2018-02-28
No lockout policy

wivern:
for #337


pavgra:
@chrisknoll, @anthonysena, fyi, diff of the PR will not be clear until https://github.com/OHDSI/WebAPI/pull/335 and https://github.com/OHDSI/WebAPI/pull/334 are not merged

chrisknoll:
Making PRs that depend on other PRs is making it difficult for us, @pavgra .  Can't these independent PRs be based off the same commit in master (the latest commit in master) so that when we look at the changes and the commits, it just looks like the changes that are for this topic?

pavgra:
@chrisknoll, we'll keep it in mind next times, but having permission to create branches directly in OHDSI/WebAPI and OHDSI/Atlas would automatically remove this org issue

chrisknoll:
Looking at the files changed, looks like it incorporated other PRs changes into it.  Also I see changes that might be intended as part of this PR (Jersey configuraiton?) but it's not clear to me how they are relevant.

 Please re-post this PR with just the changes related to the lockout policy and we'll be able to approve this faster.

Issue #302, opened 2018-01-17
Integrate WebAPI/Atlas with Common Evidence Model (CEM) for negative controls


anthonysena:
Create a process whereby candidate negative controls are identified using the new Common Evidence Model (CEM) data source. 

Process Overview:
- In Atlas, define a concept set with a set of standard drug concepts (RxNorm ingredient) or standard conditions. 
- Utilize the 'Evidence' tab on the concept set page to use the concepts in the concept set to identify candidate negative controls. In the case of drugs, the process will attempt to identify conditions that are candidate negative controls. For conditions, the process will attempt to identify candidate negative control drugs.
- Users will have the ability to specify drug/condition concepts to include/exclude in the processing.
- The results of the processing will provide filters to identify those candidate controls that CEM identifies as potential negative controls and others that are useful for navigating the result set.
- The results of the processing can be saved as a new concept set.

Tagging @ericaVoss for awareness and input on the process overview.

Related to #301

PR #345, opened 2018-03-14
Use the Common Evidence Model #302

anthonysena:
- Updated EvidenceService.java to use the Common Evidence Model. Deprecated some methods that no longer apply.
- Refactored functionality for negative controls into their own namespace: org.ohdsi.webapi.evidence.negativecontrols
- Updated comments for javadocs/miredot


pom.xml
```diff
@@ -212,6 +212,30 @@
         </executions>
         <configuration>
            <licence>cHJvamVjdHxvcmcub2hkc2kuV2ViQVBJfDIwMjAtMDMtMzF8dHJ1ZXwtMSNNQzBDRkNNQ2R2ZEhjbDg0Z2tONVI3bHlmQ2Y4SnNkYUFoVUFoMkdtOUVFdy9RelpLVzM1V2p6a2FsdzJZV009</licence>
+           <restModel>
+						<httpStatusCodes>
```
chrisknoll: Are these additions to change the way the service reports internal errors?

anthonysena: Nice catch! These additions are to globally configure miredot to automatically document the response of an endpoint based on the return type used per this article:

http://www.miredot.com/docs/manual/rest-model-configuration/status-codes/

By configuring this in the pom.xml, we eliminate the following error at compile time:

```
Miredot is unable to document the status codes of rest interfaces, unless a statusCodes-block is specified in pom.xml
```

Also by configuring this globally, it eliminates the need to document this for each end-point. I bundled this in with this PR since I spent some energy documenting the Evidence service end-points. Let me know if you have any questions or concerns.

chrisknoll: No, fair enough, just was curious. Thank you!

anthonysena:
Adding a patch for security to this branch; it is SQL Server specific and it encapsulated in https://github.com/OHDSI/WebAPI/pull/345/commits/1731170ca61dc55e2e4b099b810ce8530516e638

src/main/resources/db/migration/oracle/V2.3.0.20180302143300__negative_control_redo.sql
```diff
@@ -0,0 +1,45 @@
...
+    ancestor_pmid_count INTEGER,
+    indication INTEGER,
+    too_broad INTEGER,
+    drug_induced INTEGER,
+    pregnancy INTEGER,
+    splicer INTEGER,
+    faers INTEGER,
+    user_excluded INTEGER,
+    user_included INTEGER,
+    CONSTRAINT PK_CONCEPT_SET_NEGATIVE_CONTROLS PRIMARY KEY (id)
```
chrisknoll: This could be too long for oracle.

src/main/resources/db/migration/oracle/V2.3.0.20180302143300__negative_control_redo.sql
```diff
@@ -0,0 +1,45 @@
...
+BEGIN
+  EXECUTE IMMEDIATE 'DROP SEQUENCE ' || ${ohdsiSchema}.CONCEPT_SET_NEGATIVE_CONTROLS_SEQUENCE;
+EXCEPTION
+  WHEN OTHERS THEN
+    IF SQLCODE != -2289 THEN
+      RAISE;
+    END IF;
+END;
+
+CREATE SEQUENCE ${ohdsiSchema}.CONCEPT_SET_NEGATIVE_CONTROLS_SEQUENCE MAXVALUE 9223372036854775807 NO CYCLE;
```
chrisknoll: Check about assignign the start value of 1. I've seen some sequences start at 0...can you confirm (or just specify) that the sequencer will start at 1?  Some of our webapi calls use 0 to indicate 'new' status of record.

anthonysena:
I've moved the patch for security mentioned in this thread to a separate branch and opened as a PR (#387) since it was off-topic for these changes. I've addressed the notes regarding the Oracle sequence name length and initialization. We are still testing but I have re-based this branch off master and force pushed the update so if you have a branch locally, please remove it.

PR #351, opened 2018-03-20
Added EE results

agackovka:


anthonysena:
Frank has been evaluating this so adding him as a reviewer. @fdefalco just let me know if you need any help with this review.

src/main/java/org/ohdsi/webapi/executionengine/controller/ScriptExecutionController.java
```diff
@@ -0,0 +1,155 @@
...
+    @Path("run_script")
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public JobExecutionResource runScript(ExecutionRequestDTO dto) throws Exception {
+
+        logger.info("Received an execution script to run");
+
+        JobParametersBuilder parametersBuilder = new JobParametersBuilder();
+        parametersBuilder.addString("jobName", "Generate estimation (" + dto.sourceKey + ")");
```
fdefalco: Could we improve this message to accurately reflect the type of  execution that was requested - currently it logs "Generate Estimation" for all cases whereas the type of script execution might be other than an estimation study.

fdefalco:
Having an issue - the execution status is returning 401:

http://localhost:8080/WebAPI/executionservice/status

With this error all the generate buttons are disabled (which is good) but I can't test further.
I am running the latest updates to the ATLAS and WebAPI branch as well as latest Execution Engine docker container.

agackovka:
@fdefalco in this https://github.com/OHDSI/WebAPI/pull/351/commits/45c047ca89c0697a8f2e5d4a85cf7d7e78d8d135#diff-be3c52d5c7d9c197172d9e93ada7ca69R86 commit migration was updated. So you should bring these two changes to your db

chrisknoll:
@fdefalco : getting an error on Sql 2014:  CREATE TABLE IF NOT EXISTS ohdsi.cca_execution_ext (
IF NOT EXISTS does not  work on anything before Sql 2016. Is there some change in SQL Server requriements in this release?

chrisknoll:
Guys, The migration scripts don't make sense on sqlserver:  the script is dropping a constraint on a table before it even exists.  I'm renaming these migration files in the PR #392 so that they don't deploy, and we can determine how to address this post 2.4.

PR #369, opened 2018-03-26
Atlas 31 ancestors in concept tab

agackovka:
https://github.com/OHDSI/Atlas/issues/31

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/VocabularyService.java
+++ b/src/main/java/org/ohdsi/webapi/service/VocabularyService.java
@@ -2,8 +2,8 @@ package org.ohdsi.webapi.service;
 
 import static org.ohdsi.webapi.util.SecurityUtils.whitelist;
 
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
-import com.microsoft.sqlserver.jdbc.SQLServerDriver;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -25,7 +25,7 @@ import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
-import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.math.NumberUtils;
 import org.ohdsi.circe.helper.ResourceHelper;
 import org.ohdsi.circe.vocabulary.Concept;
@@ -105,7 +105,7 @@ public class VocabularyService extends AbstractDaoService {
   }
 
   /**
-   * @summary Calculates ancestors for every descendants
+   * @summary Calculates ancestors for the given descendants
    * 
    * @param ids concepts identifiers from concept set
    *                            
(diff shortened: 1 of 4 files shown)


src/main/java/org/ohdsi/webapi/service/VocabularyService.java
```diff
@@ -101,7 +103,66 @@ private String getDefaultVocabularySourceKey()
...
+    Source source = getSourceRepository().findBySourceKey(sourceKey);
+		
+		if (ids.ancestors == null || ids.ancestors.length == 0) { 
+			ids.ancestors = new long[] {0};
+		}
+		if (ids.descendants == null || ids.descendants.length == 0) {
+			ids.descendants = new long[] {0};
+		}
+
+    final PreparedStatementRenderer psr = prepareAscendantsCalculating(ids.ancestors, ids.descendants, source);
```
anthonysena: This code will fail when evaluating a concept that has more than 2000 concept_ids due to the parameter limit imposed when using the PreparedStatementRender. As an example, create a concept set with the RxNorm Ingredient "metformin" (concept_id = 1503297) and include the descendants (in our environment this contains 2470 descendants). 

Please see the "executeIdentifierLookup" function in the same class to see how to utilize the PreparedSqlRender.getParameterLimit(source) to chunk up the calls. I was going to attempt to revise this but given some of the other data manipulation happening in the remainder of the function, I thought it would be best for your team to evaluate this.

src/main/java/org/ohdsi/webapi/service/VocabularyService.java
```diff
@@ -101,7 +103,66 @@ private String getDefaultVocabularySourceKey()
...
+  public Map<Long, List<Long>> calculateAscendants(@PathParam("sourceKey") String sourceKey, Ids ids) {
+
+    Source source = getSourceRepository().findBySourceKey(sourceKey);
+		
+		if (ids.ancestors == null || ids.ancestors.length == 0) { 
+			ids.ancestors = new long[] {0};
+		}
+		if (ids.descendants == null || ids.descendants.length == 0) {
+			ids.descendants = new long[] {0};
+		}
```
anthonysena: Added the checks for ids.ancestors & ids.descendants in the event that someone attempts to look at the included concepts for an empty concept set or has defined a concept set that does not include descendants for a concept.

agackovka: Oh, thx. Ive simplified that check a bit

src/main/java/org/ohdsi/webapi/service/VocabularyService.java
agackovka: Done

src/main/java/org/ohdsi/webapi/service/VocabularyService.java
```diff
@@ -101,7 +103,78 @@ private String getDefaultVocabularySourceKey()
...
+    public List<Long> descendants;
+  }
+
+  protected PreparedStatementRenderer prepareAscendantsCalculating(Long[] identifiers, Long[] descendants, Source source) {
+
+    String sqlPath = "/resources/vocabulary/sql/calculateAscendants.sql";
+    String tqName = "CDM_schema";
+    String tqValue = source.getTableQualifier(SourceDaimon.DaimonType.Vocabulary);
+
+    return new PreparedStatementRenderer(source, sqlPath, tqName, tqValue,
```
anthonysena: So this code will fail in the case when we have a combination of identifiers and descendants that is greater than the number of parameters supported by the underlying RDBMS (see the PreparedSqlRender.getParameterLimit() for limits). Take a look at the executeIdentifierLookup function in the same class for a possible solution to this problem. I ran into this issue when testing a very large concept set all with the descendant flag selected.

agackovka: @anthonysena maybe I cant understand something, but in lines 131, 133 we split lists by parts of needed size and limit cannot be exceeded

agackovka: @anthonysena i believe we can have an arbitrary count of IN clauses, but each of them should not exceed the given limit

agackovka: image this query can be executed, but next query cannot (pay attention to error and count of elements)
image

anthonysena:
I pushed a commit that addresses the parameter limit. It worked well on my side but let me know if you have any concerns.

PR #374, opened 2018-03-30
User Interface for Daimon / Source Configuration

wivern:
closes OHDSI/Atlas#57


src/main/resources/db/migration/oracle/V2.3.0.20180330124512__add_source_permissions.sql
```diff
@@ -0,0 +1,8 @@
+INSERT INTO ${ohdsiSchema}.sec_permission(id, value, description) VALUES(300, 'source:post', 'Create source');
```
pavgra: Why don't you pull new id from sequence?

src/main/resources/db/migration/sqlserver/V2.2.0.20180118105000__change-daimons-priority-permission.sql
```diff
@@ -0,0 +1,11 @@
+INSERT INTO ${ohdsiSchema}.sec_permission (id, value, description)
```
pavgra: May be we can collapse migrations?

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -93,4 +116,82 @@ public SourceInfo getPriorityVocabularySourceInfo() {
...
+    } else {
+      throw new NotFoundException();
+    }
+  }
+
+  @Path("{key}")
+  @DELETE
+  @Transactional
+  public Response delete(@PathParam("key") String sourceKey) {
+    Source source = sourceRepository.findBySourceKey(sourceKey);
```
anthonysena: Deleting a source should also clean up the corresponding "Source User (<sourceKey>)" role.

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -93,4 +116,82 @@ public SourceInfo getPriorityVocabularySourceInfo() {
...
+  public SourceDetails getSourceDetails(@PathParam("key") String sourceKey) {
+    return new SourceDetails(sourceRepository.findBySourceKey(sourceKey));
+  }
+
+  @Path("")
+  @POST
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Produces(MediaType.APPLICATION_JSON)
+  public SourceInfo createSource(SourceRequest request) {
+    Source source = conversionService.convert(request, Source.class);
```
anthonysena: Creating a source should create a corresponding role named "Source user (<sourceKey>)"

anthonysena:
I'm going to leave some additional comments here that apply to both the WebAPI work and the OHDSI/Atlas#622:

- The configuration page in Atlas is hidden to unauthenticated users. Could we add in the "access-denied" component for that page?
- Source details are not secured properly - I removed all role permissions that had "source" in the name and I was still able to view the details for each source in my environment. 
- The data source name on the configuration screen should only have a link when the user has rights to view or edit the details.
- Can we tackle this as part of these changes? https://github.com/OHDSI/Atlas/issues/599?

src/main/resources/db/migration/sqlserver/V2.2.0.20180118105000__change-daimons-priority-permission.sql
anthonysena: I'd support collapsing the migrations as well and making sure the migration scripts are prefixed with "V2.3.0" (I think the other migration that is V2.2.0 was taken from the other PR). Furthermore, the new source permissions are being applied to the public role as part of these migration scripts. I'd recommend that these scripts create the permissions and associate them to the 'admin' role (role_id = 2) for these particular permissions. 

For someone setting up Atlas with security, they will need to run some SQL to grant themselves admin privileges and we should also point out that the data source management permissions can be removed if needed.

anthonysena:
Adding an additional protection for the REST path: `/source/details/<id>` and locally added a new corresponding permission "source:details:*" so that we can ensure that someone can't change the route in Atlas and view the source details. Please review and if agreeable, can you add the corresponding permission to the Flyway scripts? 

One test that I did that was problematic: I cleared the source configuration from my cache. I then signed out and attempted to refresh Atlas. It appears that the source initialization isn't called or the route is protected and the home screen hangs with the "loading" message. Can you see if you can replicate and if so address this bug?

Finally, any thoughts on how we can provide some user feedback when adding/deleting a source? This operation takes a while with > 1 data source configured and there is no visual feedback that the operation is working.

anthonysena:
I made a commit to WebAPI on this branch and to Atlas via OHDSI/Atlas#622 to button up a few bugs I found when attempting to use the application with security disabled. Please review and if you don't find any issues, we should be set to merge in these changes.

PR #378, opened 2018-04-05
Atlas profile - add calendar dates to x-axis in addition to days

wivern:
closes OHDSI/Atlas#556

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,8 +42,8 @@
     <!-- CDM properties -->
     <source.name>CDM_NAME</source.name>
     <cdm.version>5</cdm.version>
-    <!-- Profiles -->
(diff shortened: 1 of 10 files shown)


anthonysena:
I made some updates to this PR and the corresponding PR on Atlas: https://github.com/OHDSI/Atlas/pull/631/commits/5e9073a88390e769099416bac6688b2c8a231c8e. A quick summary for both sets of changes:

WebAPI:
- Refactored the permission settings and some of the variables
- Added an additional check to allow for showing dates when security is disabled

Atlas:
- Added a value to config.js so we can show dates on the profile page when security is disabled
- Updated the tooltip and x-axis display to handle the display when user authentication is disabled

Please review these and if you have any concerns, please let me know. From my side, these are both set to merge into master.

PR #379, opened 2018-04-05
Fixes #377

### LINKED ISSUE #377: set up bug
with newest WebAPI from master, I get this at some point:

```
Caused by: org.flywaydb.core.internal.dbsupport.FlywaySqlScriptException:
Migration V2.2.5.20180215105415__separate-sequences.sql failed
--------------------------------------------------------------
SQL State  : 42P01
Error Code : 0
Message    : ERROR: relation "webapi.negative_controls_sequence" does not exist
  Position: 15
Location   : db/migration/postgresql/V2.2.5.20180215105415__separate-sequences.sql (/usr/local/Cellar/tomcat/9.0.6/libexec/webapps/WebAPI/WEB-INF/classes/db/migration/postgresql/V2.2.5.20180215105415__separate-sequences.sql)
Line       : 14
Statement  : SELECT setval('webapi.negative_controls_sequence', coalesce(max(id), 1)) FROM webapi.concept_set_negative_controls
```

anthonysena:
Clean up the flyway script for PostgreSQL to address bug #377.

### DIFF
--- a/src/main/resources/db/migration/postgresql/V2.2.5.20180215105415__separate-sequences.sql
+++ b/src/main/resources/db/migration/postgresql/V2.2.5.20180215105415__separate-sequences.sql
@@ -4,15 +4,6 @@ SELECT setval('${ohdsiSchema}.cca_sequence', coalesce(max(cca_id), 1)) FROM ${oh
 CREATE SEQUENCE ${ohdsiSchema}.cohort_definition_sequence;
 SELECT setval('${ohdsiSchema}.cohort_definition_sequence', coalesce(max(id), 1)) FROM ${ohdsiSchema}.cohort_definition;
 
--- concept_set_sequence already exists - no need to create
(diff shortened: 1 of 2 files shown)


mpozhidaeva:
@anthonysena - I noticed a few things:
1)migrations for other db use "negative_controls_sequence" (not "CONCEPT_SET_NEGATIVE_CONTROLS_SEQUENCE")
2)NegativeControlRecord class uses negative_controls_sequence too
3)migration 1.0.1.2 for oracle doesn't have creation of the sequence (only in V2.2.5.20180215105415) and   "link" between id and sequence

So I suggest remove old postgres CONCEPT_SET_NEGATIVE_CONTROLS_SEQUENCE and create negative_controls_sequence with right initial start value for similarities with other databases.

What do you think about this points?

anthonysena:
@mpozhidaeva thanks for reviewing the other scripts and I would agree that we should use a consistent name across the different scripts. I think using "negative_control_sequence" works well since it is under the 30 character limit for Oracle. Let me know if you'd like to make those changes or I can take a shot at it.

PR #329, opened 2018-02-20
Entity sequence

anthonysena:
Continuation of work from @wivern as part of #324.

Add fix for Flyway migration for PostgreSQL. With this fix applied, the changes look good from my side for PostgreSQL. I've also added a commit to address a regression in another SQL Server flyway script for dropping the password/salt fields for the sec_user table per @chrisknoll's issue raised on: https://github.com/OHDSI/WebAPI/pull/327#pullrequestreview-97933012

Tagging @chen-regen for awareness and help testing on Oracle.

PR #380, opened 2018-04-10
Sequence patches

anthonysena:
Fixes to changes from #329

### DIFF
--- a/src/main/resources/db/migration/sqlserver/V2.2.5.20180215105415__separate-sequences.sql
+++ b/src/main/resources/db/migration/sqlserver/V2.2.5.20180215105415__separate-sequences.sql
@@ -102,7 +102,10 @@ GO
 UPDATE ${ohdsiSchema}.concept_set SET concept_set_id_tmp = concept_set_id;
 GO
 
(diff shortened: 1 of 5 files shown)


src/main/resources/db/migration/sqlserver/V2.2.5.20180215105415__separate-sequences.sql
```diff
@@ -102,7 +102,10 @@ GO
 UPDATE ${ohdsiSchema}.concept_set SET concept_set_id_tmp = concept_set_id;
 GO
 
-ALTER TABLE ${ohdsiSchema}.concept_set_generation_info DROP CONSTRAINT FK_concept_set_generation_info_concept_set;
```
chrisknoll: Shouldn't this always be here if the constraint was created via a prior migration?

anthonysena: Yes, it should. However, in our environment is was not and I'm unsure why that constraint was removed. So, I added an additional check to for existence before dropping.

anthonysena:
From testing, I found that several of the sequences were initialized to the current ID for the corresponding table which was causing a PK violation when creating new records. I've added another commit on this branch to address those issues.

PR #390, opened 2018-04-23
Store data sources' passwords in encrypted way

### LINKED ISSUE #381: Store data sources' passwords in encrypted way
### Expected behavior
Data sources credentials in database shouldn't be stored in plaintext.

### Solution proposal
Ask user to enter a master password during application startup / in application properties and encrypt data sources credentials entered via UI using the password. Also, double check that decrypted passwords are not printed in logs / UI / elsewhere.

wivern:
closes #381


src/main/java/org/ohdsi/webapi/source/SourceRepository.java
```diff
@@ -15,13 +15,13 @@
...
 
-import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.jpa.repository.JpaRepository;
 
 /**
  *
  * @author fdefalco
  */
-public interface SourceRepository extends CrudRepository<Source, Integer> {
+public interface SourceRepository extends JpaRepository<Source, Integer>, PersitableRepository<Source, Integer> {
```
chrisknoll: Is this change necessary? Up to this point, we've depended on Spring's CrudRepository to automatically generate a proxy instance without having to make an implementation ourselves (as you have done in SourceRepositoryImpl.java.  

Was there some probably that you needed to solve in this change?

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -30,14 +29,28 @@
...
 @Path("/source/")
 @Component
+@Transactional
 public class SourceService extends AbstractDaoService {
 
   public static final String SECURE_MODE_ERROR = "This feautre requires the administrator to enable security for the application";
 
+  @PostConstruct
+  public void ensureSourceEncrypted(){
+    getTransactionTemplateRequiresNew().execute(new TransactionCallbackWithoutResult() {
```
pavgra: We've discussed that such migration behavior should be optional, w/ ability to turn it off

konstjar:
I have some comments on this. There are environments where 10+ different data sources exist.
Administrators should to be able to see existing datasets and connection strings (host, dbname/schema) to verify connections.
If everything will be encrypted, it will be hard to define what is the current connection string and source. I suggest to add an option to view decrypted connection string (like passwords in Chrome) for admin role or at least keep host and dbname in additional fields in 'source' table and have option to see them in UI for admin role.

src/main/java/org/ohdsi/webapi/source/CheckedEncryptedStringType.java
```diff
@@ -0,0 +1,67 @@
...
+        if (rs.wasNull()){
+            return null;
+        } else {
+            final String message = rs.getString(names[0]);
+            Object result;
+            if (Objects.isNull(message)) {
+                return null;
+            }
+            if (message.startsWith(ENCODED_PREFIX) && Objects.nonNull(this.encryptor)) {
+                String value = message.substring(4, message.length() - ENCODED_SUFFIX.length());
```
chrisknoll: I believe substring(4...) is incorrect: shouldn't' you get the length of the prefix?

src/main/resources/application.properties
```diff
@@ -138,8 +138,12 @@ execution.status.period=${execution.status.period}
...
+
 person.viewDates=${person.viewDates}
 
 #Heracles settings
-heracles.smallcellcount=${heracles.smallcellcount}
\ No newline at end of file
+heracles.smallcellcount=${heracles.smallcellcount}
+
+jasypt.encryptor.enabled=${jasypt.encryptor.enabled}
+jasypt.encryptor.password=
```
chrisknoll: Is "password=" correct?  Or should it read from a property in the POM.xml via ${jasypt.encryptor.password}

pavgra:
@chrisknoll , I've extracted Data Source username and password into separate fields to encrypt them only. Also latest master was merged in.

PR #395, opened 2018-04-27
Mssql fixes

agackovka:


src/main/resources/db/migration/sqlserver/V1.0.6.0.1__schema-add-analysis_execution_password.sql
```diff
@@ -0,0 +1,57 @@
...
+WHERE TABLE_NAME = 'cca_execution_ext'
+AND TABLE_SCHEMA = '${ohdsiSchema}'))
+BEGIN
+  CREATE TABLE ${ohdsiSchema}.cca_execution_ext (
+    cca_execution_id INTEGER NOT NULL PRIMARY KEY ,
+    update_password  VARCHAR(250)
+  );
+END;
+
+IF (NOT EXISTS(SELECT * FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS
```
chrisknoll: I'm not sure this will work:  when you declare the table with a primary key (from line 6) i believe the constraint name will be randomly generated.  Later, when you try to check for the constraint name, it's not going to be the name you are expecting, and you'll try to add another primary key constraint to the table, and this might fail.

Also: I don't think we should be doing any 'if exists' check. The migration scripts should expect that the state of the database is in the state that the migration specified. By checking for 'if exists' you're opening the door to external changes to the database schema, which should not be allowed.

agackovka: Oh, i'm sorry, ive deleted that strange primary key

agackovka: removed

Issue #398, opened 2018-04-28
Incorrect values in Healthcare Visit reports


pavgra:
Result sets for Healthcare Visit reports are mapped to DTOs incorrectly:

```
s.paidPmPm = rs.getString("paid_pmpm") == null ? null : new BigDecimal(rs.getDouble("paid_pmpm")).setScale(2, BigDecimal.ROUND_HALF_UP);
s.allowedChargedRatio = rs.getString("allowed_charged") == null ? null : new BigDecimal(rs.getDouble("paid_pmpm")).setScale(2, BigDecimal.ROUND_HALF_UP);
s.paidAllowedRatio = rs.getString("paid_allowed") == null ? null : new BigDecimal(rs.getDouble("paid_pmpm")).setScale(2, BigDecimal.ROUND_HALF_UP);
```

PR #399, opened 2018-04-28
Fixed Healthcare Visit reports mapping

pavgra:
Closes https://github.com/OHDSI/WebAPI/issues/398


chrisknoll:
Pavel, thanks for finding this, that was a great find.  We've set up a release-2.4.0-fixes branch to contain all of our last-minute updates for the upcoming release. I'm merging this into that branch.

PR #405, opened 2018-05-03
Impala jdbc drivers support added

### LINKED ISSUE #404: Add Impala support
Add Impala support like that was done for Netzza with separate profile in pom.xml.

wivern:
closes #404


chrisknoll:
Hi,
  The support of these platforms (netezza and impala) is beginning to get a little out of hand when we have this large amount of profile/dependency settings that have to be carried along into our POM.xml.

Before this becomes the 'standard practice', couldn't there be a POM.xml for impala hosted somewhere that we can just set up a single dependency in the POM.xml for impala, and the referenced dependency will pull out all the necessary dependencies that Impala needs? This way you centralize the impala dependencies such that if there's other apps that need that suite of dependencies, you just change the central repository's POM.xml, (which can be version controlled) and when a new release of the impala stack comes out, we just increment the depedency in WebAPI's POM.xml and we will automaticily pull the new dependencies for Impala.

I feel like this is the approach spring boot takes (where 1.5.2 sets up dependiecis on hibernate,, JPA, logging, servlet api, etc) without us having to keep up to date on all the other dependencies that spring boot manages under the covers.

chrisknoll:
Spoke to @pavgra , we'll try to think about refactoring and externalizing chunks of our POM if it is to support a specific capability at a later time.

However: when building the WAR with the webapi-impala profile active, it failed to load the dependencies for example:

```
        <dependency>
          <groupId>hive</groupId>
          <artifactId>hive_metastore</artifactId>
          <version>1.0.0</version>
        </dependency>
```
Is there another maven repo that should be included in the main POM.xml? Example:

```
    <repository>
      <id>central</id>
      <name>Central Repository</name>
      <url>http://repo.maven.apache.org/maven2/ </url>
    </repository>
```

chrisknoll:
Ok, I followed the instructions, and put the jar files in the the /impala folder.  Some thoughts:

After putting the JAR files in the folder, when I went to git add, it identified those new files as being commited to the repo.  This opens up the chance that someone accidentally  commits these JARS to the repo.  We could add a rule to .gitignore for /impala/*.jar be ignored, but I have another idea:

Instead of copying these JARS to a folder under the root of the repo, could we introduce a new parameter to the POM.xml called something like 'impala.jdbc.classpath' that the users can specify an external directory to point to the drivers, and the maven build step can perform the filecopy from the external directory to the /target folder that we want to bundle with the build.  Would this work?

Next issue: the non-impala jars that are in the /impala folder:  commons-logging-1.1.1.jar, slf4j-api-1.5.11.jar, httpclient-4.1.3.jar etc.  I believe the core maven POM.xml has dependencies that reference these JARS (possibly a different version, however).  My worry is that these JARS are going to lead to a conflict with the other libs.  I'm looking at the build messages, tho, and it looks like only impala specfic classesa re being instaleld to the local M2:

```
pom.xml not found in ImpalaJDBC41.jar
Installing C:\Git\WebAPI\src\main\extras\impala\ImpalaJDBC41.jar to C:\Users\cknoll1\.m2\repository\com\cloudera\ImpalaJDBC41\2.5.41\ImpalaJDBC41-2.5.41.jar

--- maven-install-plugin:2.5.2:install-file (hive_metastore) @ WebAPI ---
pom.xml not found in hive_metastore.jar
Installing C:\Git\WebAPI\src\main\extras\impala\hive_metastore.jar to C:\Users\cknoll1\.m2\repository\hive\hive_metastore\1.0.0\hive_metastore-1.0.0.jar

--- maven-install-plugin:2.5.2:install-file (hive_service) @ WebAPI ---
pom.xml not found in hive_service.jar
Installing C:\Git\WebAPI\src\main\extras\impala\hive_service.jar to C:\Users\cknoll1\.m2\repository\hive\hive_service\1.0.0\hive_service-1.0.0.jar

--- maven-install-plugin:2.5.2:install-file (libfb) @ WebAPI ---
pom.xml not found in libfb303-0.9.0.jar
Installing C:\Git\WebAPI\src\main\extras\impala\libfb303-0.9.0.jar to C:\Users\cknoll1\.m2\repository\libfb\libfb303\0.9.0\libfb303-0.9.0.jar

--- maven-install-plugin:2.5.2:install-file (libthrift) @ WebAPI ---
pom.xml not found in libthrift-0.9.0.jar
Installing C:\Git\WebAPI\src\main\extras\impala\libthrift-0.9.0.jar to C:\Users\cknoll1\.m2\repository\libthrift\libthrift\0.9.0\libthrift-0.9.0.jar
```

So I think this is not a real issue, and I can probably remove those un-used .jar files from the /impala directory.

Next issue:  In the logs, we show messages indicating success/failure of loading the driver. When i ran, I got this output:

```
driver loaded: org.postgresql.Driver
driver loaded: com.microsoft.sqlserver.jdbc.SQLServerDriver
error loading oracle.jdbc.driver.OracleDriver driver.
driver loaded: com.amazon.redshift.jdbc.Driver
error loading com.cloudera.impala.jdbc4.Driver driver.
error loading net.starschema.clouddb.jdbc.BQDriver driver.
error loading org.netezza.Driver driver.
```

The impala driver should have loaded successfully, but returned an error. Is the name 'com.clodera.impala.jdbc4.Driver' correct?  I didn't find a driver in jdbc4.Driver, but there is a jdbc41.Driver.

-Chris

chrisknoll:
Re: the JDBC Driver name:
In WebAPI: https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/DataAccessConfig.java#L66

This is where we specify the list of driver names to load.  We tried to rely on automatic JDBC driver loading, but it did not work for us, so we are doing an Class.forName(driverName) for each driver to load it directly.  Not ideal, but neither is the JDBC41 implementation of most of these JDBC drivers.

As you can see, the impala jdbc driver name is set to: "com.cloudera.impala.jdbc4.Driver", and this should probably be "com.cloudera.impala.jdbc41.Driver".

Issue #413, opened 2018-05-15
Incidence rates generation failed on Impala


pavgra:
```
org.springframework.jdbc.UncategorizedSQLException: PreparedStatementCallback; uncategorized SQLException for SQL [DELETE FROM synpuf_50_test_results.ir_strata WHERE analysis_id = ?]; SQL state HY000; error code 500051; SimbaImpalaJDBCDriver(500051) ERROR processing query/statement. Error Code: 0, SQL state: TStatus(statusCode:ERROR_STATUS, sqlState:HY000, errorMessage:AnalysisException: Impala does not support modifying a non-Kudu table: synpuf_50_test_results.ir_strata
), Query: DELETE FROM `synpuf_50_test_results`.`ir_strata` WHERE (`ir_strata`.`analysis_id` = 3181).; nested exception is java.sql.SQLException: SimbaImpalaJDBCDriver(500051) ERROR processing query/statement. Error Code: 0, SQL state: TStatus(statusCode:ERROR_STATUS, sqlState:HY000, errorMessage:AnalysisException: Impala does not support modifying a non-Kudu table: synpuf_50_test_results.ir_strata
), Query: DELETE FROM `synpuf_50_test_results`.`ir_strata` WHERE (`ir_strata`.`analysis_id` = 3181).
```

PR #416, opened 2018-05-15
Fixing IR generation at Impala

pavgra:
Fixes https://github.com/OHDSI/WebAPI/issues/413

### DIFF
--- a/src/main/java/org/ohdsi/webapi/ircalc/PerformAnalysisTasklet.java
+++ b/src/main/java/org/ohdsi/webapi/ircalc/PerformAnalysisTasklet.java
@@ -92,7 +92,7 @@ public class PerformAnalysisTasklet implements Tasklet {
       options.resultsSchema = jobParams.get("results_database_schema").toString();
       options.vocabularySchema = jobParams.get("vocabulary_database_schema").toString();
 
-      String delete = "DELETE FROM @tableQualifier.ir_strata WHERE analysis_id = @analysis_id";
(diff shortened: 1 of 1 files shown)


wivern:
There are not enough changes to make Impala work.
The issue is that SqlRender is ignored since PreparedStatementRender got null instead of actual Source.

chrisknoll:
By the way, @wivern  and @pavgra , There is a method on PreparedStatementRenderer to set the targetDialect directly (via setTargetDialect(), and so if you didn't want to make a 'fake' source instance, you can pass null as the source, and later call setTargetDialect().

chrisknoll:
@wivern:  when you merge the PRs, can you please use 'rebase and merge' if the PR commits look meaningful, or just 'squash and merge' to combine all commits into a single commit into master. We'd like to keep the commit history on master as clean as possible. 

Thank you!

pavgra:
@chrisknoll , @wivern , the code block uses `getSql()`, which is just getter of the `sql` field of the PreparedStatementRenderer. The `sql` field is calculated in constructor, where `SqlTranslate` is used based on the `source` param. `setTargetDialect` doesn't affect the `sql` anyhow and the constructor is the only place where `SqlTranslate` is applied

chrisknoll:
Thanks @pavgra , as usual you are correct :)  But that's a flaw in the PSR that should be addressed:  there's no way to use the target dialect for anything. I'll work on a PR for this.

Issue #418, opened 2018-05-16
Incidence rates generation failed on Impala #2


pavgra:
During IR generation following exception is thrown:
```
org.springframework.jdbc.UncategorizedSQLException: StatementCallback; uncategorized SQLException for SQL [CREATE TABLE wt74k27aanalysis_events↵ AS↵SELECT↵row_number() over (partition by P.person_id order by P.start_date) as event_id, P.person_id, P.start_date, P.end_date, P.op_start_date, P.op_end_date, CAST(NULL as bigint) as TARGET_CONCEPT_ID, CAST(NULL as bigint) as visit_occurrence_id↵↵FROM↵(↵  select distinct T.subject_id as person_id, T.cohort_start_date as start_date, T.cohort_end_date as end_date, OP.observation_period_start_date as op_start_date, OP.observation_period_end_date as op_end_date↵  from wt74k27atime_at_risk T↵  JOIN synpuf_50_test.observation_period OP on T.subject_id = OP.person_id and T.cohort_start_date between OP.observation_period_start_date and OP.observation_period_end_date↵) P↵]; SQL state [HY000]; error code [500051]; [Simba][ImpalaJDBCDriver](500051) ERROR processing query/statement. Error Code: 0, SQL state: TStatus(statusCode:ERROR_STATUS, sqlState:HY000, errorMessage:AnalysisException: Unsupported type 'null_type' in column 'target_concept_id' of table 'wt74k27aanalysis_events'↵CAUSED BY: TableLoadingException: Unsupported type 'null_type' in column 'target_concept_id' of table 'wt74k27aanalysis_events'↵), Query: CREATE TABLE wt74k27aanalysis_events↵ AS↵SELECT↵row_number() over (partition by P.person_id order by P.start_date) as event_id, P.person_id, P.start_date, P.end_date, P.op_start_date, P.op_end_date, CAST(NULL as bigint) as TARGET_CONCEPT_ID, CAST(NULL as bigint) as visit_occurrence_id↵↵FROM↵(↵  select distinct T.subject_id as person_id, T.cohort_start_date as start_date, T.cohort_end_date as end_date, OP.observation_period_start_date as op_start_date, OP.observation_period_end_date as op_end_date↵  from wt74k27atime_at_risk T↵  JOIN synpuf_50_test.observation_period OP on T.subject_id = OP.person_id and T.cohort_start_date between OP.observation_period_start_date and OP.observation_period_end_date↵) P.; nested exception is
```

[...]

PR #419, opened 2018-05-17
Fixed IR analysis sql for Impala

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/418

### DIFF
--- a/src/main/resources/resources/incidencerate/sql/performAnalysis.sql
+++ b/src/main/resources/resources/incidencerate/sql/performAnalysis.sql
@@ -78,9 +78,20 @@ INTO #time_at_risk
 from cteRawData
 ;
 
--- from here, take all the people's person_id, start_date, end_date, create an 'events table' 
+-- from here, take all the people's person_id, start_date, end_date, create an 'events table'
+CREATE TABLE #analysis_events (
+  event_id BIGINT,
+  person_id BIGINT,
+  start_date DATE,
+  end_date DATE,
+  op_start_date DATE,
+  op_end_date DATE,
+  TARGET_CONCEPT_ID BIGINT,
+  visit_occurrence_id BIGINT
+);
+
+INSERT INTO #analysis_events
 select row_number() over (partition by P.person_id order by P.start_date) as event_id, P.person_id, P.start_date, P.end_date, P.op_start_date, P.op_end_date, CAST(NULL as bigint) as TARGET_CONCEPT_ID, CAST(NULL as bigint) as visit_occurrence_id
-INTO #analysis_events
 FROM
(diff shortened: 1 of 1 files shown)


src/main/resources/resources/incidencerate/sql/performAnalysis.sql
```diff
@@ -78,9 +78,20 @@ INTO #time_at_risk
 from cteRawData
 ;
 
--- from here, take all the people's person_id, start_date, end_date, create an 'events table' 
+-- from here, take all the people's person_id, start_date, end_date, create an 'events table'
+CREATE TABLE #analysis_events (
```
chrisknoll: This change switches from something that can be converted into a CTAS into a create-then-insert mode of functionality. The difference is that CTAS are orders of magnitude faster than create-then-inserts, and this is a pattern we use extensively around WebAPI.  We shouldn't be expected to convert all CTAS into create-then-insert.

pavgra: @chrisknoll , I understand that CTAS works a lot faster in PDW, but it cannot infer data types from NULLs, even the columns are CASTed, and it fails, when I run following:
```
CREATE TABLE temp_test AS SELECT CAST(null AS bigint) as nullable_int
```
with the error:
```
Unsupported type 'null_type' in column 'nullable_int' of table 'temp_test'
```

The reason for this, I assume, is following. In the INSERT statements data types of result set's columns are retrieved from the resulting values:
https://github.com/apache/impala/blob/789c5aac23480acc6e18c057b767b65fdd791c97/fe/src/main/java/org/apache/impala/analysis/InsertStmt.java#L275

Whereas in CTAS they are calculated via base table of the column (and there is no base table for NULL):
https://github.com/apache/impala/blob/ba84ad03cb83d7f7aed8524fcfbb0e2cdc9fdd53/fe/src/main/java/org/apache/impala/analysis/CreateTableAsSelectStmt.java#L171

I also think that this looks, at least, very strange, but I cannot do anything with it right now. I am going to double-check issues section in Impala and create a new issue, but this will be a solution only for future developments.

chrisknoll: Ok, so the path forward is:  we'll take this change to work on Implala, knowing that this impacts MPP systems negatively.  An issue is created with Impala to properly derive column types from the casting done in SELECTS, and as soon as it is resolved, we revert back to CTAS.

pavgra: https://issues.apache.org/jira/browse/IMPALA-7045

Issue #436, opened 2018-05-22
User session is invalidated due to concurrent token refresh


pavgra:
When multiple tabs are opened in parallel (and therefore concurrent token refresh is happening), user is logged out

PR #437, opened 2018-05-23
Concurrent token refresh fix

pavgra:
Fixes https://github.com/OHDSI/WebAPI/issues/436

Solution similar to used in https://github.com/tymondesigns/jwt-auth

### DIFF
--- a/src/main/java/org/ohdsi/webapi/shiro/TokenManager.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/TokenManager.java
@@ -1,35 +1,41 @@
 package org.ohdsi.webapi.shiro;
 
 import io.jsonwebtoken.Claims;
-import io.jsonwebtoken.JwsHeader;
 import io.jsonwebtoken.JwtException;
 import io.jsonwebtoken.Jwts;
-import io.jsonwebtoken.MissingClaimException;
 import io.jsonwebtoken.SignatureAlgorithm;
 import io.jsonwebtoken.SignatureException;
-import io.jsonwebtoken.SigningKeyResolverAdapter;
 import io.jsonwebtoken.impl.crypto.MacProvider;
 import java.security.Key;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import javax.servlet.ServletRequest;
 import javax.servlet.http.HttpServletRequest;
 import org.apache.shiro.web.util.WebUtils;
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/shiro/TokenManager.java
```diff
@@ -1,35 +1,41 @@
...
  */
 public class TokenManager {
-  
+
   private static final String AUTHORIZATION_HEADER = "Authorization";
-  
+
   private static final Map<String, Key> userToKeyMap = new HashMap<>();
-  
+  private static final ExpiringMultimap<String, Key> gracePeriodInvalidTokens = new ExpiringMultimap<>(30000);
```
chrisknoll: What does the 30000 ms here in the ExpiringMultimap constructor do?

src/main/java/org/ohdsi/webapi/shiro/TokenManager.java
```diff
@@ -40,32 +46,47 @@ public static String createJsonWebToken(String subject, Date expiration) {
...
+            })
+            .filter(Objects::nonNull)
+            .findFirst()
+            .orElseThrow(() -> new SignatureException("Signing key is not reqistred for the subject."));
   }
-  
+
+  protected static String getUntrustedSubject(String jws) {
+    int i = jws.lastIndexOf('.');
+    String untrustedJwtString = jws.substring(0, i+1);
```
wivern: What if string does not contain dot?

src/main/java/org/ohdsi/webapi/util/ExpiringMultimap.java
```diff
@@ -0,0 +1,58 @@
...
+        return map.containsKey(key)
+                ? map.get(key).entrySet().stream().map(Map.Entry::getKey).collect(Collectors.toList())
+                : new ArrayList<>();
+    }
+
+
+    public synchronized void put(K key, V value) {
+
+        PassiveExpiringMap<V, Object> storage = map.get(key);
+        if (storage == null) {
```
wivern: Map.putIfAbsent?

src/main/java/org/ohdsi/webapi/shiro/TokenManager.java
pavgra: @chrisknoll , defines how long invalidated keys will be stored. I've started moving the value into app's properties, but then payed attention that all methods in the class are static and the class itself is not a @Component, so left it hardcoded for now

PR #438, opened 2018-05-23
Not finished Cohorts / IRs should be invalidated on startup

### LINKED ISSUE #429: Not finished Cohorts / IRs should be invalidated on startup

wivern:
closes #429

### DIFF
--- a/src/main/java/org/ohdsi/webapi/IExecutionInfo.java
+++ b/src/main/java/org/ohdsi/webapi/IExecutionInfo.java
@@ -1,0 +1,16 @@
+package org.ohdsi.webapi;
+
+import java.util.Date;
+
+public interface IExecutionInfo<T extends IExecutionInfo> {
+    Date getStartTime();
+    Integer getExecutionDuration();
+    GenerationStatus getStatus();
+    boolean getIsValid();
+    String getMessage();
+    T setStartTime(Date startTime);
+    T setExecutionDuration(Integer executionDuration);
+    T setStatus(GenerationStatus status);
+    T setIsValid(boolean valid);
+    T setMessage(String message);
+}
--- a/src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationInfoRepository.java
+++ b/src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationInfoRepository.java
@@ -1,0 +1,12 @@
+package org.ohdsi.webapi.cohortdefinition;
+
+import org.ohdsi.webapi.GenerationStatus;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.List;
+
+public interface CohortGenerationInfoRepository extends CrudRepository<CohortGenerationInfo, CohortGenerationInfoId> {
+
+    List<CohortGenerationInfo> findByStatus(GenerationStatus status);
+    List<CohortGenerationInfo> findByStatusIn(List<GenerationStatus> statuses);
+}
--- a/src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
+++ b/src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
@@ -3,12 +3,12 @@ package org.ohdsi.webapi.service;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.ohdsi.webapi.GenerationStatus;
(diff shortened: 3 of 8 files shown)


src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
```diff
@@ -227,4 +231,17 @@ public TransactionTemplate getTransactionTemplateNoTransaction() {
   public String getOhdsiSchema() {
       return ohdsiSchema;
   }
+
+  protected IExecutionInfo invalidateExecution(IExecutionInfo executionInfo) {
+
+    return executionInfo.setIsValid(false)
```
chrisknoll: I spoke about this point under the Atlas PR, but I'll put it here too:
The GenerationStatus.ERROR enum value is not appropriate for storing the state of the generation:  the only valid options are 'PENDING/RUNNING/COMPLETE'.  So I think we should remove ERROR from the GenerationStatus, and in the invalidateExecution() call, just set the isValid to false, the generationStatus to COMPELTE and failMessage to 'Invalidated by System'.

src/main/java/org/ohdsi/webapi/IExecutionInfo.java
```diff
@@ -0,0 +1,16 @@
+package org.ohdsi.webapi;
+
+import java.util.Date;
+
+public interface IExecutionInfo<T extends IExecutionInfo> {
```
chrisknoll: I'm curious why this would need to be an interface?  ExecutionInfo is basically a data structure, but the purpose of defining an interface would be that there's different implementations of it?  When would we need different implementations of a simple data structure?

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationInfoRepository.java
```diff
@@ -0,0 +1,12 @@
+package org.ohdsi.webapi.cohortdefinition;
+
+import org.ohdsi.webapi.GenerationStatus;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.List;
+
+public interface CohortGenerationInfoRepository extends CrudRepository<CohortGenerationInfo, CohortGenerationInfoId> {
```
chrisknoll: By defining a XXXGenerationInfoRepository, it makes it appear that a generation info could live independently of the thing that this GenerationInfo represents: Ie you can have a CohortGenerationInfo live on independently of the existence of the Cohort that the generation is for.  So, I would have expected that CohortGenerationInfo is reachable only by accessing the Cohort that the generation belongs to.

wivern: But on the other hand to invalidate executions we should find CohortGenerationInfo with Pending or Running statuses. Having access to CohortGenerationInfo only via Cohort we have to enumerate all cohort generations on each cohort and check its status every time when WebAPI is starting.

src/main/java/org/ohdsi/webapi/IExecutionInfo.java
chrisknoll: Yes, I see now.  I was expecting the 'GenerationInfo' class to be an embedded collection where you use a common data object but specify the table using java annotations.  But, it's actually a full blown entity with it's own dedicated table so what you're describing makes sense.

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationInfoRepository.java
chrisknoll: Yes, you are right, I believe that there is a way to use a JPA query expression to fetch children of parent entities that satisfy some criteria, but it's a bit of an advanced technique, and I don't have specific guidance on how it's accomplished.  

At the root of this conversation thread, I proposed a simple update statement to reset all the status and set a fail message as a potentially simpler implementation for this.  But what you have implemented works, so there is no fault there, I'm just suggesting an alternative.

src/main/java/org/ohdsi/webapi/IExecutionInfo.java
chrisknoll: Another thing I missed:  the GenerationInfo class isn't shareable across things because cohort generation info contains asdditional summary statistics about the generation that is not universal across the concept set generation info.  So either having a base class that dictates what the 'shared' fields across them or an interface that they both must implement accomplishes the goal, and we get a bit more flexiblity off of the interface approach, so I'm right with you there. Thanks for the clarification.

chrisknoll:
Hi, looking at this further:
In postgresql, the is_valid field is defined as a boolean, while it is integer/bit in mssql server. This means that you can't do set is_valid = 0 in postgresql (it won't convert int to boolean).

So, I think that means we are left to letting JPA do the translation for us and your implementation looks good.

PR #440, opened 2018-05-24
Fixed calculation of values in summary table

pavgra:
Fixes https://github.com/OHDSI/Atlas/issues/697


gowthamrao:
@pavgra @chrisknoll only fixes visits, not drug. 
https://github.com/OHDSI/Atlas/issues/697#issuecomment-394707139

PR #441, opened 2018-05-24
Not finished Cohorts / IRs should be invalidated on startup

wivern:
merge #438 to rc-2.4.1


pavgra:
For https://github.com/OHDSI/WebAPI/issues/429

Issue #427, opened 2018-05-18
init_heracles_periods fails in Netezza


pavgra:
`INSERT INTO` fails in `init_heracles_periods.sql` at  Netezza Release 7.2.1.2-P1 [Build 48187]. While CTAS works

pavgra:
Should be closed by https://github.com/OHDSI/SqlRender/pull/125

PR #442, opened 2018-05-24
Fix results schema generation for Netezza

pavgra:
Fixes https://github.com/OHDSI/WebAPI/issues/427

### DIFF
--- a/src/main/resources/ddl/results/init_heracles_periods.sql
+++ b/src/main/resources/ddl/results/init_heracles_periods.sql
@@ -6,6 +6,7 @@ CREATE TABLE #temp_period (
 	period_end_date date
 );
 
+INSERT INTO #temp_period
 with digits as (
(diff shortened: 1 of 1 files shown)


src/main/resources/ddl/results/init_heracles_periods.sql
```diff
@@ -6,6 +6,7 @@ CREATE TABLE #temp_period (
 	period_end_date date
 );
 
+INSERT INTO #temp_period
```
chrisknoll: This statement: "INSERT INTO {table} with cte()..." doesn't work, does it?

pavgra: @chrisknoll , that's why I requested your review - the statement works at Postgres and Netezza, and alignes to SQL ANSI (http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt). But, you say, it fails at MS SQL, right?

PR #465, opened 2018-06-05
ATLP-373 Novartis: filter for AD auth

### LINKED ISSUE #464: Active Directory Authentication search filter
It is unable to get Atlas work with mailing list or group that require memberOf search.
Suggestion: add optional searchFilter that should run additional checks during login process with AD.

wivern:
closes #464


src/main/java/org/ohdsi/webapi/shiro/ADRealm.java
```diff
@@ -20,4 +50,36 @@ protected AuthorizationInfo doGetAuthorizationInfo(PrincipalCollection principal
...
+                    dnAttributesMapper);
+
+            if (result.size() == 1) {
+                String userDn = result.iterator().next();
+                List<String> filterResult = ldapTemplate.search("", String.format(searchFilter, userDn), SearchControls.SUBTREE_SCOPE,
+                        dnAttributesMapper);
+                if (!filterResult.isEmpty()) {
+                    return super.queryForAuthenticationInfo(token, ldapContextFactory);
+                }
+            } else {
```
chrisknoll: So do uses that have multiple results not able to authenticate?

chrisknoll: Ok, thanks for the info. In our internal domain, our IT has managed to associate a single principle with multiple windows logins (Ie I have my normal {login}, and there's also an associated account called admin_{login}). At one point in our internal env that has security enabled, I was actually getting both accounts ({login} and admin_{login}) added to sec_user.  I haven't had time to dig in enough to troubleshoot it, but when I saw the check for multiple results, I was wondering if that might bite me here at our internal env.  I'll find time to dig into it and see if it remains an issue.

chrisknoll:
@wivern , Please rebase your branch on top of current master and force push your branch. That will update the PR and you can corect any conflicts you find.

Thank you.

PR #470, opened 2018-06-09
AD search filter

wivern:
Adds AD search filter


chrisknoll:
@wivern: please take your local branch and rebase it to latest master. We incorporated many changes across 2.4.3 fixes and we'd like to see any open PRs rebased to master.

Note: don't rely on merge, we would just like to see these commits applied to the tip of current master.

Thank you.

PR #510, opened 2018-07-16
Fix Heracles execution in Impala

pavgra:
(cherry picked from commit c5fa8ea)


src/main/resources/resources/cdmresults/sql/report/observationperiod/observationlengthbyage.sql
```diff
@@ -1,6 +1,8 @@
 SELECT
-  cast(cast(ard1.stratum_1 AS INT) * 10 AS VARCHAR) + '-' +
-  cast((cast(ard1.stratum_1 AS INT) + 1) * 10 - 1 AS VARCHAR) AS category,
+  CONCAT(
+    cast(cast(ard1.stratum_1 AS INT) * 10 AS VARCHAR), '-',
```
chrisknoll: make this a fixed length varchar: ie cast(cast(ard1.stratum_1 AS INT) * 10 AS VARCHAR(38),

Or varchar(19), whatever is appropriate to hold the max size of INT * 10 as a string.

Issue #508, opened 2018-07-16
Add connection check button to Data sources management screen


pavgra:
Today an admin cannot verify whether an added DS can be accessed and the configuration was properly entered. A connection check button would make the task easier.

PR #519, opened 2018-07-19
ATLP-444. Added endpoint for checking ds connection

agackovka:
fixes https://github.com/OHDSI/WebAPI/issues/508

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/SourceService.java
+++ b/src/main/java/org/ohdsi/webapi/service/SourceService.java
@@ -19,6 +19,7 @@ import javax.ws.rs.core.Response;
 import org.apache.commons.lang3.StringUtils;
 import org.jasypt.encryption.pbe.PBEStringEncryptor;
 import org.jasypt.properties.PropertyValueEncryptionUtils;
+import org.ohdsi.sql.SqlTranslate;
 import org.ohdsi.webapi.shiro.management.Security;
 import org.ohdsi.webapi.source.*;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -246,6 +247,18 @@ public Response delete(@PathParam("sourceId") Integer sourceId) throws Exception
...
   }
 
+  @Path("connection/{key}")
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
+  public SourceInfo checkConnection(@PathParam("key") final String sourceKey) throws Exception {
+
+    final Source source = sourceRepository.findBySourceKey(sourceKey);
+    final JdbcTemplate jdbcTemplate = getSourceJdbcTemplate(source);
+    jdbcTemplate.execute(SqlTranslate.translateSql("select 1", source.getSourceDialect()));
```
pavgra: Will fail in Oracle. Use:
`SqlTranslate.translateSql("select 1;", dataSource.getType().getOhdsiDB()).replaceAll(";$", "");`

src/main/resources/db/migration/oracle/V2.5.0.20180720117120__source-connection-check-rule.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO ${ohdsiSchema}.sec_permission(id, value, description) VALUES(${ohdsiSchema}.sec_permission_id_seq, 'source:connection:*:get', 'Check source connection');
```
chrisknoll: Just confirm that this is the correct syntax for Oracle.

PR #526, opened 2018-07-27
Updated oracle migration DDL for analysis execution.

chrisknoll:
We need someone with Oracle environment to test this migration fix. Based on documentation, these columns should work, but I do not currently have access to an Oracle instance to test the deploy.


chrisknoll:
This file is being addressed in #535, so we will close this and have this change in that PR.

PR #527, opened 2018-07-30
Inclusion Rule reports at person-level.

chrisknoll:
Added new parameter: modeId to REST endpoint for cohort inclusion reports.
Added new column to cohort_inclusion_* tables to store mode_id.
Updated circe to 1.6.0-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -563,7 +563,7 @@
     <dependency>
       <groupId>org.ohdsi</groupId>
       <artifactId>circe</artifactId>
-      <version>1.5.1-SNAPSHOT</version>
+      <version>1.6.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.json</groupId>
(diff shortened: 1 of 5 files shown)


chrisknoll:
For those migrating existing tables, the following DDL is required to add the new columns:

```
ALTER TABLE @resultsSchema.cohort_inclusion_result ADD mode_id  int NOT NULL DEFAULT 0;
ALTER TABLE @resultsSchema.cohort_inclusion_stats ADD mode_id  int NOT NULL DEFAULT 0;
ALTER TABLE @resultsSchema.cohort_summary_stats ADD mode_id  int NOT NULL DEFAULT 0;
```

jduke99:
@chrisknoll Does this not automatically update now with flyway?  We hit a bunch of errors with a clean WebAPI install due to these columns not being present in these tables.

chrisknoll:
Flyway does not manage the results schema.   Those tables you are using from the flyway migration script were originally supposed to be for exemplar purposes, but only as a convenience for single cdm setups.  It's is now considered a bad practice to use the webapi schema as a CDM results schema, so I'd recommend you migrate off that setup.  We'd like to remove those CDM Results schema table from the Flyway migration (by applying DROP TABLE statements to a migration) so that we can avoid this confusion.

For your situation, you should run the above ALTER statements on your results schema directly.

PR #529, opened 2018-08-01
Added Users import from Active Directory

### LINKED ISSUE #506: Users and roles import from AD / LDAP
Today a user is added into Atlas database only after he / she logins there for the first time. Therefore there is no ability to assign roles to the user in advance. Moreover, the roles assignment can only be done manually even if the user belongs to a certain AD group which could define a set of roles which he / she transitively gets.

Therefore there is a suggestion to add a new page in Atlas: LDAP import screen with the ability to define DN-role mapping and "Import" button, which would scan the AD for not imported users or users with changed roles, provide the change list to admin for approval and do the import from AD. There is going to be "Import users from LDAP / AD" button available at the Configuration page if the LDAP / AD auth was setup, which will navigate an admin to the import page.

As a further improvement a scheduler for the automatic users import can be added.

wivern:
closes #506


pavgra:
@wivern, the change list for the feature looks quite loosely coupled to other WebAPI codebase, so I'd suggest to put it into separate package

chrisknoll:
Hi @pavgra  and @wivern,

Pavel and I spoke briefly about incorporating the idea of 'Controllers' in WebAPI, but after looking up information ont he 'Controller' concept, it was clearly a UI oriented construct (part of MVC) wheras WebAPI is specifically a service-oriented application... And any 'controller' would be defined on the client of a SPA (in other words: you don't get a UI from any endpoitns in WebAPI, it just acts as a service layer).

But, looking at these commits, I see the introduction of a 'controller' namespace Firstly: I don't think we need to make a namespace dedicated to a design pattern: ie, you wouldn't make a namespace for 'adapter' or a namespace for 'visitor' or a namespace for 'singleton'...if you have a 'controller' type of class in your codebase, put it in the same namespace that other classes share the same part of the application.

Namespace ideology aside: I'd like to understand what we're trying to accomplish with adding the notion of a 'controller' to WebAPI.  This feels like it goes against the grain of everything else that we've put into place so far, and I want to just keep the codebase from becoming a hodge-podge of different design principles that force devs to tilt their heads in various directions depending on which part of the codebase they are looking at.

wivern:
Hi @chrisknoll,
the main goal of suggested solution is to separate business logic from request handling that conforms best practices of [multi-tier architecture](https://www.ibm.com/developerworks/library/wa-aj-multitier/index.html).

From my point of view UserImportController is not controller from MVC pattern but mostly endpoint or request handler. The controller name was chosen because it is familiar to the people who knows pretty much popular Spring Framework. Anyway the 'controller' namespace presents the layer but not the subject of design pattern.

chrisknoll:
Thanks to the link to multi-tier architecture, I always wondered what that meant (sarcasm ;)).  

Kiddding aside:  the UserImportController ```@Compoonent``` you've defined set up new routes that can be handled by WebAPI. In the WebAPI structure, all those classes are put into the 'services' namespace.  if you wanted to separate the handing of those endpoints into separate classes, you'd still make a service (Perhaps this one is the UserService service, which provides mechanisms for importing).   If you don't want a monolithic class, or if you'd like to be able to provide a 'UserServiceProvider' to implement some of the behavior of the service, then you're free to do that.

But, to introduce a new notion of a 'Controller' when we already have the notion of a 'service' that basically does the same thing, it's just fragmenting the codebase into 2 separate ideas which could conform to the same style and form, while using terminology that is usually reserved for presentation-tier constrcuts.  

As to the article you've referenced, 2 things:
1- It's a bit out of date. That was from 2009, when SPAs didn't even exist.  MVC isn't even in the picture in WebAPI, it's assumed if you want to consume the service, you provide your own client UI in front of the service tier.
2- This article is another example where the term 'Controller' is referencing a Presentation-tier concept.  But there is no presentation tier in WebAPI. There is a 'presentation-tier facing element' in the form of Spring ```@component``` and reset endpoints, but Controller usually has a specific meaning.  The closest thing that's in the article you reference is a 'Request Handler', but in my opinion that's the business tier interface for the client, not something for 'presentation'.

So, I found an article describing a "front Controller" here: https://www.tutorialspoint.com/design_pattern/front_controller_pattern.htm 

In it, the idea that there's a single handler that receives requests and a dispatcher that handles the request.  I believe, with Spring ```@Component``` and ```@Path``` directives, we are leveraging the spring framework to declare the handlers that respond to dispatch requests.  So, you're not introducing any new multi-tier architecture here, we've always had it.  And, via Spring Repositories, we have separation between our data Tier and service tier.

What you _are_ introducing is new namespaces and code structure that breaks the current, established patterns.

So, please move your controller class to ```org.ohdsi.webapi.service```, and name it appropriately.  It appears that there is already a ```UserService``` defined, (however it does have an empty root ```@Path``` annotation, it should probably be set to ```user/```.  It looks like you want to introduce a 'composite' services where multiple classes come together to service one top-level endpoint (in this case /user).  This should be fine, we can have a UserService alongside a UserImportService. I don't recommend we make a separate namespace for each service under ```org.ohdsi.webapi.service``` since you'd probably have 1 class under each namespace, and that's not what namespaces are for.

As far as business logic be separated from the request handler, we can do that. Usually the business logic that is used to handle a service request is placed under a namespace specific to the domain of interest. For example,```CohortDefinitionService``` is under ```org.ohdsi.webapi.service```, but the business logic is found in ```org.ohdsi.webapi.cohortdefinition``` (for example the Tasklets used to generate cohort definitions).  In some cases, the business logic is handed off to a separate library.

Let me know if anything needs clarification.

src/main/java/org/ohdsi/webapi/userimport/entities/RoleGroupMappingEntity.java
```diff
@@ -0,0 +1,71 @@
+package org.ohdsi.webapi.userimport.entities;
```
chrisknoll: We don't have namespaces specfic for entitites.  if these classes are support user domain functions, we can put them in ```org.ohdsi.webapi.user``` and put ```@Entity``` Taslkets, and anything else belonging to this domain in that namespace.

chrisknoll:
@wivern : more digging on this led me here:
https://stackoverflow.com/questions/6827752/whats-the-difference-between-component-repository-service-annotations-in

So did you want to make it a broader effort to leverage ```@Component```, ```@Service``` and ```@Repository``` usage? I'd still follow the current structure we have for this PR, but if you want to talk about a longer term strategy to identify the types of ```@Components``` that we should annotate in WebAPI, we can do that.  

I'd also add: we're not using Spring REST APIs for endpoints, we're using JAX-RS which spring provides a jersey hook. So I don't see us migrating everything from ```@Path()``` to ```@RequestMapping```.

PR #535, opened 2018-08-05
Fixes for users import from ad

pavgra:


chrisknoll:
I've addressed the issue with SQL migration.  However, we still need verification for Oracle.

src/main/resources/db/migration/oracle/V1.0.6.0.1__schema-add-analysis_execution_password.sql
```diff
@@ -1,44 +1,42 @@
...
+  update_password  CHARACTER VARYING(255)
 );
 
-ALTER TABLE IF EXISTS ${ohdsiSchema}.input_files
-  DROP CONSTRAINT IF EXISTS fk_sof_cca_execution;
+ALTER TABLE ${ohdsiSchema}.input_files DROP CONSTRAINT fk_sof_cca_execution;
 
-ALTER TABLE IF EXISTS ${ohdsiSchema}.output_files
-  DROP CONSTRAINT IF EXISTS fk_sif_cca_execution;
+ALTER TABLE ${ohdsiSchema}.input_files DROP CONSTRAINT fk_sif_cca_execution;
```
chrisknoll: This table does not exist yet.

I believe this was because the 'alter table if exists' were replaced with straight alters, but we should assume that no tables exist yet (since they'd only be created by the migration process), and so we should not have any need for 'if exists' statements anywhere. If ```input_files``` already exists, then the migration should throw an error if it tries to create it.  This will catch environmental errors.

src/main/resources/db/migration/oracle/V2.5.0.20180725172844__add-ad-import-permissions.sql
```diff
@@ -1,64 +1,76 @@
...
   provider VARCHAR2(32) NOT NULL,
   group_dn VARCHAR2(2000) NOT NULL,
   group_name VARCHAR2(255),
   role_id INTEGER NOT NULL,
   CONSTRAINT role_group_prov_uniq UNIQUE(provider, group_dn, role_id)
-);
\ No newline at end of file
+);
+
+CREATE OR REPLACE TRIGGER new_sec_role_group
```
chrisknoll: Can you explain the need for a trigger?

chrisknoll:
I've made additional updates to fix oracle migrations:  there was references to 'bigint' column types, but Oracle does not have a 'BIGINT' column type.  The docs say that you should use NUMBER(19) to support that type of data.

PR #537, opened 2018-08-07
CommonEvidenceModel v1.0

anthonysena:
Adding functionality to support the CommonEvidenceModel (CEM) v1.0 release.  

@chrisknoll - I need to add a few more commits to this branch to remove some classes that are no longer needed and to address an Oracle Flyway issue. Just getting this up to GitHub for now.


src/main/java/org/ohdsi/webapi/evidence/negativecontrols/NegativeControlRepository.java
```diff
@@ -1,25 +1,25 @@
...
- */
-public interface NegativeControlRepository extends CrudRepository<NegativeControlRecord, Integer> {
-  @Query("SELECT n FROM NegativeControlRecord n WHERE n.sourceId = :sourceId AND n.conceptSetId = :conceptSetId")
-  Collection<NegativeControlRecord> findAllBySourceIdAndConceptId(@Param("sourceId") int sourceId, @Param("conceptSetId") int conceptSetId);
-  
-  @Modifying
-  @Query("DELETE FROM NegativeControlRecord n WHERE n.conceptSetId = :conceptSetId")
-  void deleteAllByConceptSetId(@Param("conceptSetId") int conceptSetId);
-}
+///*
```
chrisknoll: Maybe just delete this class instead of commenting it out? we can always getit back by checking out another release

src/main/resources/db/migration/oracle/V2.5.0.20180713123456__cem_v_1.sql
```diff
@@ -0,0 +1,28 @@
+DELETE FROM ${ohdsiSchema}.CONCEPT_SET_GENERATION_INFO;
```
chrisknoll: Unless there is a need for doing this in script blocks, just use plain SQL statements.

anthonysena: It is necessary to clear out any old generation information for concept sets as they will be invalid after applying these updates.

anthonysena:
I'm also planning to add a ddl results endpoint for the single result table needed for CEM so I'll push one more commit to this branch. Otherwise, this is ready for review. Thanks!

src/main/resources/db/migration/oracle/V2.5.0.20180713123456__cem_v_1.sql
```diff
@@ -0,0 +1,15 @@
...
+
+DROP SEQUENCE ${ohdsiSchema}.negative_controls_sequence;
+
+CREATE SEQUENCE ${ohdsiSchema}.negative_controls_sequence START WITH 1 MAXVALUE 9223372036854775807 NOCYCLE;
+
+DROP TABLE ${ohdsiSchema}.CONCEPT_SET_NEGATIVE_CONTROLS;
+
+CREATE TABLE ${ohdsiSchema}.CONCEPT_SET_NEGATIVE_CONTROLS (
+    id INTEGER NOT NULL DEFAULT NEXTVAL('negative_controls_sequence'),
```
chrisknoll: DEFAULT NEXVAL() doesn't exist in Oracle, not until 12c.  So, you should jus tleave that field NOT NULL, and anyone on oracle will have to pass in the id from the sequencer if they want to populate the table.

anthonysena:
Thanks @chrisknoll for the review. I believe I've addressed the items noted. Here are a few important notes related to these changes related to configuration and testing:

- This PR renames the 'evidence' daimon to 'CEM' ($ohdsi_schema.source_daimon.daimon_type == 3) to reflect the fact that these services are designed to work with the CommonEvidenceModel. Furthermore, I've introduced a 'CEMResults' daimon ($ohdsi_schema.source_daimon.daimon_type == 4) which will hold tables used for processing results against the CEM. The intent here is to have CEM follow the same daimon model for the CDM sources.
- As a consequence of the above changes, tables in the OHDSI schema that originally held results for negative controls are dropped and results are removed. As with any upgrade, proper backups should be taken to retain any data required.
- A new endpoint has been added to retrieve the DDL required to support setting up the 'CEMResults' daimon: WebAPI/ddl/cemresults?dialect=<your_db_dialect>&schema=<your_results_schema>.

src/main/resources/db/migration/sqlserver/V2.5.0.20180713123456__cem_v_1.sql
```diff
@@ -0,0 +1,29 @@
+DELETE FROM ${ohdsiSchema}.CONCEPT_SET_GENERATION_INFO;
+
+IF (EXISTS (SELECT *
```
chrisknoll: Let's remove the 'exists' checks on these scripts.  If it exists, it should be a migration error.

src/main/resources/resources/evidence/sql/negativecontrols/exportNegativeControls.sql
```diff
@@ -1,5 +1,5 @@
-INSERT INTO @evidenceSchema.NC_RESULTS (
-	job_id, 
+INSERT INTO @resultsSchema.NC_RESULTS (
```
chrisknoll: I thought I saw that the DDL service had a new schema for the evidenceSchema...should the evidence schema be called 'evidenceSchema' like you said? or 'cemresultsSchema'?  'resultsSchema' would be the WebAPI's Results Daimon schema, I believe?  

Unless you are thining that we refer to it as the 'resultsSchema' and the caller will specify the TableQualifier from the 'CEMResults' daimon to specify the 'resultsSchema' paramater....I just want to avoid any confusion between the CEMResults schema and the WebAPI's 'results schema'...

anthonysena: Let me update these schema qualifiers to reflect the new daimon types. I'm thinking they should be: @cemSchema and @cemResultsSchema respectively. I'll push a new commit to make this consistent.

anthonysena:
I've updated the Flyway script to remove the 'exists' condition and also revised the code to consistently refer to the 'cem_schema' and 'cem_results_schema'. For the DDL service, the endpoint are:

WebAPI/ddl/results?dialect=<db_dialect>&schema=<results_schema>
WebAPI/ddl/cemresults?dialect=<db_dialect>&schema=<results_schema>

chrisknoll:
I've removed the IF EXISTS checks from the PG migration.

@pavgra , would you mind pulling this branch and confirming the migration on PG. I will do this on oracle.

pavgra:
@anthonysena , @chrisknoll , I've pulled the branch, started WebAPI - migration `V2.5.0.20180713123456__cem_v_1.sql` appeared in `schema_version` and WebAPI works.

PR #544, opened 2018-08-10
Utilization - Removed roll-up using concept_ancestor

tagwhom:
For analysis_id between 4000 and 4025, removed roll-up using concept_ancestor. With roll-up takes 10 to 20 times more time to run, compared to without.

### DIFF
--- a/src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql
+++ b/src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql
@@ -26,8 +26,8 @@
 {DEFAULT @runHERACLESHeel = FALSE}   --runHERACLESHeel = @runHERACLESHeel
 {DEFAULT @CDM_version = '4'}  --we support 4 or 5,   CDM_version = @CDM_version
 {DEFAULT @cohort_period_only = FALSE}
-{DEFAULT @rollupUtilizationVisit = 0} -- by default, do not roll up Visit utilization using concept ancestor
-{DEFAULT @rollupUtilizationDrug = 0} -- by default, do not roll up Drug utilization using concept ancestor
+{DEFAULT @rollupUtilizationVisit = FALSE} -- do not roll up Visit utilization using concept ancestor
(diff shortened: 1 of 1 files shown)


chrisknoll:
Perhaps instead of removing it outright, we can give the option to the user to enable rollups?  It can be off by default (in case people aren't aware of how it works) but for those advanced users, they can enable it?

chrisknoll:
Ok, one thing that we should deal with for this PR is allow the user to pass those options in during cohort reporting generation. This will be an ATLAS ui change.  @pavgra , @gowthamrao will you have resources to spare to hook this up? 

It looks like the new options are:
```
{DEFAULT @rollupUtilizationVisit = FALSE} -- do not roll up Visit utilization using concept ancestor
{DEFAULT @rollupUtilizationDrug = FALSE} -- do not roll up Drug utilization using concept ancestor
{DEFAULT @periods = ''} -- default value for periods is blank. No calendar periods will be used. 
{DEFAULT @includeVisitTypeUtilization = 32025, 32028, 32022} -- default values of visit type concept id to perform utilization analysis
{DEFAULT @includeDrugTypeUtilization = 38000175,38000176} -- default values of visit type concept id to perform utilization analysis
{DEFAULT @includeCostConcepts = 31978, 31973, 31980} -- default values for cost concept_id (Allowed, Charged, Paid)
{DEFAULT @includeCurrency = 44818668} -- default values for currency (USD - United States dollar)
```

-Chris

pavgra:
@chrisknoll, I am going to add the required UI and back-end changes to support allow changing the parameters. But meanwhile, I'd merge the PR since it has built-in defaults and not going to break anything.

chrisknoll:
Ok, let me run a local test and if it checks out, we'll merge in.

chrisknoll:
@tagwhom : please remember that if you are going to make a change to master off of a pull request, you should fist create a branch in your local repository, and then submit the PR using that branch.  The reason is that when this merges back into master, your upstream fork of master branch may look slightly different than what you have now. It's always safer (and a recommended best practice) to create a branch before making any changes.

Issue #543, opened 2018-08-10
Related concepts retrieval fails in Impala


pavgra:
Exception:
`AnalysisException: Arithmetic operation requires numeric operands: 'Has relation to descendant of : ' + RELATIONSHIP_NAME`

Due to the `'Has relation to descendant of : ' + RELATIONSHIP_NAME RELATIONSHIP_NAME` in `/resources/vocabulary/sql/getRelatedConcepts.sql`.

The plus sign should be replaced with `CONCAT`.

PR #546, opened 2018-08-11
Fixed related concepts retrieval in Impala

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/543


chrisknoll:
@pavgra , we're seeign a lot of these PRs come in to address using CONCAT instead of ```+```, and was wonering, since other dialects have translation rules for +, could this be addressed by just providing a translation rule for ```+```?  That way, we won't have to find all the other places we are using the ```+``` operator to concat strings?

pavgra:
@chrisknoll, this is what I first thought of - create the rule for SqlRender, but was not able to come up with concrete regex for it which SqlRender would be able to process.

PR #559, opened 2018-08-16
Heracles reports casting fixes.

chrisknoll:
Fixed casting for concept stratums to BIGINT.
Switched cohort_defintion_id from IN to =.
Applied GROUP BY to subqueries to avoid MPP optimizations that fetch all rows and risk casting non-numerics to BIGINT.


pavgra:
Tested using Impala

PR #562, opened 2018-08-17
LDAP/AD import improvements

wivern:
1. Added connection test before import
2. Added help for user mapping columns


chrisknoll:
Please resolve conflicts.  Thank you!

chrisknoll:
There appears to only be a postgresql migration. Is there also one needed for oracle and sqlserver?

Issue #495, opened 2018-06-29
New Cohort Characterization


pavgra:
Today there are several mechanisms in Atlas which provide cohort characterization: Heracles (adoption of Achilles for cohorts), Feature extraction (Generate w/ Features) and Cost & Utilization reports. All of them try to accomplish the same goal, but at the same time introduce extra complexity both to business users and developers since being placed in different parts of the application and doing their job in different ways. This is not obvious which to choose and which to extend during development. The suggestion is to provide a single mechanism which will support both current needs and be simply extendable by business users (ad-hoc) and developers.

The core idea is to re-use https://github.com/OHDSI/FeatureExtraction package by providing a user options either to select existing FE analyses, create analysis on the fly based on Criterias or raw SQL:

![atlas cohort characterization use cases](https://user-images.githubusercontent.com/10010071/42067210-10f5b79c-7b13-11e8-8a38-f6e5a4c7cd6f.png)

First thing is to provide a library of FE analyses and ability to create new ones:

![atlas cohort characterization fe analysis sequence](https://user-images.githubusercontent.com/10010071/42067217-17f42646-7b13-11e8-8956-8fcdcbfd66bd.png)

When a user has available analyses, he/she can create a Cohort Characterization report by picking cohorts to analyze, analyses set and configuring the analyses parameters:

![atlas cohort characterization cc sequence](https://user-images.githubusercontent.com/10010071/42067256-3d5b441e-7b13-11e8-92b6-548018abfce0.png)

Database for such solution could look like:
![atlas cohort characterization database](https://user-images.githubusercontent.com/10010071/42067262-46e19b82-7b13-11e8-98d2-75e275b51754.png)

DAO layer:

![atlas cohort characterization data classes](https://user-images.githubusercontent.com/10010071/42067265-4e600420-7b13-11e8-81d0-82735aaeb216.png)

Service layer

![atlas cohort characterization service classes](https://user-images.githubusercontent.com/10010071/42067273-539bfc1e-7b13-11e8-9b6e-6ca2453bf93b.png)

Original discussion of the solution was held at Janssen R&D. Following contributors took participation

@pbr6cornell, @gowthamrao, Dave Kern, @chrisknoll, @pavgra.

gowthamrao:
Excellent work @pavgra , this is in-line with decision to remove and refactor the three cohort characterization mechanisms. 

**Diagram 1: User options**
[...]

PR #573, opened 2018-08-24
New cohort characterization

agackovka:
Related to https://github.com/OHDSI/WebAPI/issues/495


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcConst.java
```diff
@@ -0,0 +1,5 @@
+package org.ohdsi.webapi.cohortcharacterization;
+
+public interface CcConst {
```
chrisknoll: We don't need to prefix classes with their module subsystem. the java ```package``` is enough to do that.

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortDefinition.java
```diff
@@ -47,7 +47,7 @@
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.ohdsi.circe.cohortdefinition.CohortExpression;
-import org.ohdsi.standardized_analysis_api.cohortcharacterization.design.Cohort;
+import org.ohdsi.standardized_analysis_api.Cohort;
```
chrisknoll: Sorry, i didn't notice this before, but I don't think it's normal to have ```_``` in a package name?  Maybewe should adopt a 'codename' for this api (like shiro or fasterxml does)?

PR #588, opened 2018-09-13
Base Cohort Characterization functionality

pavgra:
This PR brings a subset of the base functionality described in https://github.com/OHDSI/WebAPI/issues/495. Improvements, such as generation of cohorts in isolated temp tables, Feature Analysis CRUD, Criteria analysis, etc, are going to be done iteratively in separate PRs.


pavgra:
@chrisknoll,

1) could you tell, how do you "get a response of the generation" which _doesn't exist_? Obviously, you send a request to _create_ a generation, which just returns some response as any other create / update method of RESTful API does. So that's POST. Otherwise, pls, point me to smth that describes why my understanding of RESTful API is wrong.

2) same as before - get a look at https://en.wikipedia.org/wiki/Representational_state_transfer

3) will do

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
```diff
@@ -84,6 +85,11 @@ public AtlasSecurity() {
 
     this.sourcePermissionTemplates.put("cohortdefinition:*:report:%s:get", "Get Inclusion Rule Report for Source with SourceKey = %s");
     this.sourcePermissionTemplates.put("cohortdefinition:*:generate:%s:get", "Generate Cohort on Source with SourceKey = %s");
+    this.sourcePermissionTemplates.put("cohort-characterizations:*:generate:%s:post", "Generate Cohort Characterization on Source with SourceKey = %s");
```
pavgra: @wivern , this overlaps with `checkSourceAccess` while `"cohort-characterizations:%s:generate:*:post"` remains to be unsecured

chrisknoll:
Ok, after our off-line discussion: a 'generation' manifests itself as a stand alone entity, where the user takes action that results in a new entity representing the generation.  Sounds like a POST.

In other areas of the app where we generate (which are numerous), the GET URI to a /generate endpoint was more RPC-ish, and also didn't result in a collection of generations being added to (we maintain 1 gen per source).  I'm not a REST-pureist, so I think it's acceptable for those subsystems to continue to work that way. but, I fully expect we will want to revisit the entire RPC vs. REST form of the WebAPI for the 3.0 design activities.

pavgra:
From the same discussion, since really there is no standard for REST to use only plural nouns in endpoints, and since currently, pretty all endpoints use singular form, we'll update the endpoints to be singular. @wivern, pls address it

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -0,0 +1,505 @@
...
+        sortInnerEntities(savedEntity);
+        
+        final String serialized = this.serializeCc(savedEntity);
+        savedEntity.setHashCode(serialized.hashCode());
+        
+        return repository.save(savedEntity);
+    }
+
+    public void deleteCc(Long ccId) {
+        repository.delete(ccId);
```
pavgra: I believe this should also clean up permissions

pavgra:
@wivern, in case of lacking permissions the `checkSourceAccess` throws 500, but that should be 403

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
pavgra: @wivern , `permissionManager.removePermissionsFromTemplate(atlasSecurity.getCohortCharacterizationCreatorPermissionTemplate(), id.toString());`

chrisknoll:
Problem during generation:
When I generate a CC, i am getting a 403 that led to this line: ```methodName":"checkSourceAccess","fileName":"GenerationUtils.java","lineNumber":13``` but secuirty is disabled.

List of sources on Generation Tab:
The list of sources to generate on should be restricted to those sources that have both a CDM and Results daimon set up. Please see cohort definition's generation view to see how to filter sources to those sources with the required daimons.  Ie: sources with only a Vocabulary daimon set up should not appear as an option to generate the characterization.

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.batch.repeat.RepeatStatus;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import static org.ohdsi.webapi.Constants.Params.*;
+
+public class GenerateLocalCohortTasklet implements StoppableTasklet {
```
pavgra: Why permanent table? FE seems to support temp tables:

```
* @param cohortTable
*            The name of the cohort table. Can be a temp table (e.g. "#cohort_temp"). If it is a permanent table provide the full path, e.g.
*            "cdm_synpuf.dbo.cohort".
```
https://github.com/OHDSI/FeatureExtraction/blob/master/java/org/ohdsi/featureExtraction/FeatureExtraction.java#L386

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
```diff
@@ -0,0 +1,105 @@
...
+  class GenerateTask implements Runnable {
+
+    private String targetTable;
+    private CohortCharacterizationEntity characterization;
+    private Source source;
+
+    public GenerateTask(ChunkContext chunkContext) {
+      Map<String, Object> jobParameters = chunkContext.getStepContext().getJobParameters();
+      targetTable = jobParameters.get(TARGET_TABLE).toString();
+      characterization = ccService.findById(Long.valueOf(jobParameters.get(COHORT_CHARACTERIZATION_ID).toString()));
```
pavgra: @wivern, pls, make abstract this tasklet from any specific use-case. I'd like to use the same code in PA. So I'd suggest to pass cohort definitions id list to the tasklet

pavgra:
@wivern , one more thing to fix: not finished jobs should be invalidated on application startup

pavgra:
@wivern, would be great to add column with error message occurred during job execution to `cc_generation` and appropriate DTO

chrisknoll:
I'm seeing this error when going to the generation list with security disabled:
image

I think the issue is that createdBy is null when security is disabled, so I believe we either need null-checks in the code, or created by should be initialized to some 'annonymous' state...but probably we'll need null checks.

chrisknoll:
> @wivern, would be great to add column with error message occurred during job execution to cc_generation and appropriate DTO (batch_job_execution.exit_message -> cc_generation.exit_message)

Also, would be useful to be able to delete generations (via a DELETE to /generations/{id}) so that we can clean up any old executions that we don't want.

chrisknoll:
Error during generation on MSSQL:
```
Parse error at line: 167, column: 280: Incorrect syntax near 'with'.; nested exception is java.sql.BatchUpdateException: Parse error at line: 167, column: 280: Incorrect syntax near 'with'.
```

chrisknoll:
I have a concern about what i'm seeing in the logs about what is being passed into the characterization job:
```
 cohort_characterization_id=1, source_id=2, generate_stats=true, design={"cohorts":[{"id":30,"name":"[411] TNFi excluding IFX","description":null,"expressionType":"SIMPLE_EXPRESSION","expression":{"Title":null,"PrimaryCriteria"
```
The actual design expression is one of the parameters, which will be excessively large (multiple criteria expressions with concept sets embedded, it will be quite a mess).  I believe that when we do a generation, we attach the executed design to some record in the database (unless when we click on the hash in the UI, is that read the paramater from the job to show the json?)   If it is stored in a generation-level field, could we just pass in the generationId as the paramater to the job instead of the entire design as json?

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -0,0 +1,400 @@
...
+    private void initTx() {
+        DefaultTransactionDefinition txDefinition = new DefaultTransactionDefinition();
+        txDefinition.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+        TransactionStatus initStatus = this.transactionTemplate.getTransactionManager().getTransaction(txDefinition);
+        this.transactionTemplate.getTransactionManager().commit(initStatus);
+    }
+    
+    private class CcTask {
+
+        final String prevalenceRetrievingQuery = "\n" +
```
chrisknoll: We usually take things like in-line sql and export them to a .sql file and put it into /resources.

src/main/resources/ddl/results/cohort_characterizations.sql
```diff
@@ -0,0 +1,22 @@
+CREATE TABLE @results_schema.cc_results
+(
+  type VARCHAR(255) NOT NULL,
+  cohort_characterization_generation_id BIGINT NOT NULL,
```
chrisknoll: This column will be too big for Oracle.

pavgra:
Following path is not secured `cohort-characterization/generation/*/design`

pavgra:
@chrisknoll ,

> Error during generation on MSSQL
fixed

> I think the issue is that createdBy is null when security is disabled
fixed in UI branch

pavgra:
> If it is stored in a generation-level field

@chrisknoll , since we don't have a physical table for generations, the `cc_generation` is a view, the design is persisted as a batch job parameter.

pavgra:
Also, @wivern , when you are done with permissions, pls make sure that MS SQL and Oracle migrations align to PG

pavgra:
> it's just that we need an additional table to capture the design state for the given generation/execution for the given source

@chrisknoll , I honestly cannot get why would you like to create separate table if it is now based on existing Spring Batch tables and fully satisfies requirements?

image

> do not see anywhere that is captured the user who performed the generation (but this should be captured)

agreed

pavgra:
@chrisknoll , ah, now I got it. Yes, I remember that we discussed it before. Makes sense.

pavgra:
Faced one more issue. Working on it
image

PR #591, opened 2018-09-18
Added logging of security-related events

aklochkova:


src/main/java/org/ohdsi/webapi/WebApi.java
```diff
@@ -13,15 +14,16 @@
  * will source this file).
  */
 @EnableScheduling
+@ComponentScan({"org.ohdsi.webapi","com.odysseusinc.logging"})
```
wivern: The spring-way is to:
1. library contains only pojo services those injects via @configuration and @bean
2. library has autoconfiguration option
but not explicit scan that looks nasty

src/main/java/org/ohdsi/webapi/WebApi.java
```diff
@@ -14,7 +13,6 @@
  * will source this file).
  */
 @EnableScheduling
-@ComponentScan({"org.ohdsi.webapi","com.odysseusinc.logging"})
```
chrisknoll: Hi, sorry, I am a spring noob, but isn't the ``@ComponentScan`` necessary to find the auto-wired beans, or is that something that is now obsolete with newer versions of Spring?

aklochkova: Hello, @SpringBootApplication includes @ComponentScan

Issue #603, opened 2018-09-22
Cohort Characterization displays inner FE analysis names for Preset analyses


pavgra:
Should use Atlas Feature Analysis names and ids instead

![image](https://user-images.githubusercontent.com/10010071/45911383-751da080-bde0-11e8-9b40-97ea11f4fceb.png)

Issue #585, opened 2018-09-12
Restrict official support of RDBMS of WebAPI to Postgresql


chrisknoll:
For WebAPI 3.0, I'd like to propose that we drop official support for mssql and oracle in favor of single support for PostgreSQL (we can decide a minimum version possibly 9.6 or 10).

### Important Note!
This is not a proposal to restrict CDM platforms to only postgresql.  SQL Render has already proven successful in supporting cross platform QUERIES and simple DDL (creating temp tables with specific data types).  That is not the point of this Issue.  SQL Render can not handle the use-cases necessary for the implementation of WebAPI.

### Background

The challenges have been that each platform requires a certain amount of re-writing in order to implement the proper functionality that WebAPI requires, but using platform-specific rules.  

When dealing with foreign keys, sequences, unique constraints, each platform has their own twist on how you create them, rules around modifying them, etc.

Instead, we can officially support postgresql for each release, but leave it to independent contributors to support other platforms of choice.  We do not need to drop the multi-platform capabilities of flyway (so that people who are forced to use oracle have an option), but I am proposing that those people who need this functionality provide the implementation themselves, and contribute back to the source.

pavgra:
@chrisknoll, fully support for the proposal.

gklebanov:
I am assuming you are talking about the database where ATLAS artifacts are being stored (a.k.a. "OHDSI schema"). In this case - yes, total sense.

t-abdul-basser:
Fully support this proposal @chrisknoll. It is long past time that we bundle our own postgres in a docker container (say, as part of BroadSea) and relieved installers of having to worry about the WebAPI DB at all.

pavgra:
Linking to https://github.com/OHDSI/WebAPI/issues/533

yradsmikham:
Is there an update to this issue? Would like to understand the roadmap since this could impact how we plan on using the WebAPI and other OHDSI services.

chrisknoll:
As of 2.8, the only platform supported is Postgres for WebAPI.  We, of course, can accept Pull Requests to add migration scripts for other platforms.

PR #604, opened 2018-09-22
Display Atlas Feature Analysis names instead of OHDSI Feature Extraction analysis names for results

pavgra:
Fixes https://github.com/OHDSI/WebAPI/issues/603


chrisknoll:
@pavgra , i had to make extensive chagnes to the MSSQL migrations due to that changing the PK field type on one table that is reference by foreign keys must have all their foreign key columns and constraints re-stated to the proper type.

This was made much more difficult by usign default constraint names.  I've had to alter the existing master-version of the cohort-characterization scripts because looking up the default constraint names were deemed too difficult, and this hasn't been published ina  release branch, so I made an exception.

The message we need to deliver to the team members is: if you are writing sql server migraitons, do not depend on default constraint names.  Everything must have a name.   I would always review prior migration scripts for examples on how constraints were applied/modified.  In addition, I'd drop the DEFAULT NEXT VALUE FOR constraints on the PK columns because it makes migration scripts that don't support defaults from sequeences (aka: oracle) look different; ie: the oracle migration will have to put sequence right in the inset statements, while the other platforms will assume the default value.  In addition, naming default constraints on mssql is very complicated. Let's just not do it.

All these problems will go away with #585.  Until then we have to suffer the pain of multi-database support.

PR #605, opened 2018-09-24
Fixes AD users import login case

wivern:


src/main/java/org/ohdsi/webapi/user/importer/DefaultUserImporter.java
```diff
@@ -96,10 +96,11 @@ public DefaultUserImporter(@Autowired(required = false) ActiveDirectoryProvider
   public void importUsers(List<AtlasUserRoles> users) {
 
     users.forEach(user -> {
+      String login = user.getLogin().toLowerCase();
```
pavgra: Why not to move to a separate method and re-use here and in `UpdateAccessTokenFilter`?

pavgra: The idea is to have such explicit method for retrieval of a new user's login that we don't miss it anymore. And, yes, that obviously has to be a separate user utils class

PR #606, opened 2018-09-24
Fixes Cancel Cohort generation button doesn't kill underlying DB query

### LINKED ISSUE #579: Cancel Cohort generation button doesn't kill underlying DB query
### Steps to reproduce:
1. Start cohort generation
2. Press Cancel button
3. Execute (in case of PG):
```
SELECT pid, age(clock_timestamp(), query_start), usename, query 
FROM pg_stat_activity 
WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%' 
ORDER BY query_start desc;
```

### Expected results
Cohort generation query is *not* presented in the results.

### Actual results
The query is presented in the results and still running.

wivern:
fixes #579


src/main/java/org/ohdsi/webapi/util/CancelableJdbcTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+      private String currSql;
+
+      @Override
+      public int[] doInStatement(Statement stmt) throws SQLException, DataAccessException {
+        int[] rowsAffected = new int[sql.length];
+        cancelOp.setStatement(stmt);
+        if (JdbcUtils.supportsBatchUpdates(stmt.getConnection())) {
+          for (String sqlStmt : sql) {
+            this.currSql = appendSql(this.currSql, sqlStmt);
+            stmt.addBatch(sqlStmt);
```
pavgra: Why is `stmt.addBatch` inside the loop? Shouldn't it be outside?

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -497,9 +505,25 @@ public Response cancelGenerateCohort(@PathParam("id") final int id, @PathParam("
     });
 
     cohortGenerationService.getJobExecution(source, id)
-            .ifPresent(job -> {
+            .ifPresent(jobExecution -> {
               try {
-                jobOperator.stop(job.getJobId());
+                Job job = cohortGenerationService.getRunningJob(jobExecution.getJobId());
```
pavgra: Shouldn't this logic be moved into a shared utils so that it can be applied in other places?

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
```diff
@@ -178,6 +171,10 @@ public RepeatStatus execute(final StepContribution contribution, final ChunkCont
 
   @Override
   public void stop() {
+    try {
```
pavgra: Shouldn't this method be moved into generic tasklet?

src/main/java/org/ohdsi/webapi/util/CancelableJdbcTemplate.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.jdbc.support.JdbcUtils;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+import javax.sql.DataSource;
+import java.sql.BatchUpdateException;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class CancelableJdbcTemplate extends JdbcTemplate {
```
pavgra: Why this tasklet is not applied to IR / CC?

src/main/java/org/ohdsi/webapi/util/CancelableJdbcTemplate.java
wivern: Generally sql is an array of statetement, but `Statement.addBatch` accepts single statement to batch execution. 
This can be replaced with join for `this.currSql` and `Arrays.forEach` for addBatch but in this case we have two loops instead one.
The code above just borrowed from Spring JdbcTemplate as is.

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
wivern: Hmmm, give me a time to think about that...

chrisknoll:
I think we need a merge from master here so we can ensure all the latest code is accounted for.

pavgra:
@wivern , would be very useful if canceled generation displays "Canceled" status

image

pavgra:
@wivern , would be good to display "Canceled" status for interrupted generations instead of "Failed". Otherwise, tested for Cohorts, CC, IRs and Pathways and looks good.

pavgra:
@wivern , would be good to show same results for both persons and records columns for canceled jobs. I would suggest using "..." in both cases

image

Issue #613, opened 2018-09-26
AD import issues


mpozhidaeva:
1. Issue
Public role should be granted automatically with user import. Hide from UI. Also check during user update/modification that he has public role

2. Issue
Problem with user modification in AD import:
I have tried to update all users by add the group to public first and then re-import all users for modify their groups and nothing happened. I had to drop all database and import all users as new users with all correct groups and everything is good now.
It seems like update (like adding AD group to new role) doesn’t affect existing users.

Steps to reproduce:
1.	Add AD group to some roles
2.	Import users
3.	Add AD group to another role
4.	Import users
And nothing happens on this point.

Fixed versions: 2.5.2, master
Origin issue https://odysseusdataservices.atlassian.net/browse/ATLP-526

PR #614, opened 2018-09-26
Fixed AD import

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/613

### DIFF
--- a/src/main/java/org/ohdsi/webapi/user/importer/DefaultUserImporter.java
+++ b/src/main/java/org/ohdsi/webapi/user/importer/DefaultUserImporter.java
@@ -94,16 +94,18 @@ public class DefaultUserImporter implements UserImporter {
 
   @Override
   @Transactional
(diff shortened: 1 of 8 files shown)


src/main/java/org/ohdsi/webapi/user/importer/DefaultUserImporter.java
```diff
@@ -75,7 +76,7 @@ public DefaultUserImporter(@Autowired(required = false) ActiveDirectoryProvider
             .map(user -> {
               AtlasUserRoles atlasUser = new AtlasUserRoles();
               atlasUser.setDisplayName(user.getDisplayName());
-              atlasUser.setLogin(user.getLogin());
+              atlasUser.setLogin(user.getLogin().toLowerCase());
```
pavgra: UserUtils

PR #616, opened 2018-09-26
Added login to kerberos for general datasource

aklochkova:


pom.xml
```diff
@@ -638,6 +641,11 @@
       <artifactId>spring-ldap-core</artifactId>
       <version>2.3.2.RELEASE</version>
     </dependency>
+    <dependency>
+      <groupId>com.odysseusinc</groupId>
+      <artifactId>krb-login</artifactId>
```
wivern: artifactId

Issue #568, opened 2018-08-22
Estimation & Patient Level Prediction Specification Editors


anthonysena:
# Overview

ATLAS & WebAPI provide the ability to design population level effect estimation  (PLE) and patient level prediction (PLP) studies. The current set of capabilities in ATLAS are limited in several respects:

- The ATLAS designer does not expose all of the options available in each of the underlying R methods libraries: CohortMethod in the case of PLE and PatientLevelPrediction in the case of PLP.
- The ATLAS designer is limited to a single comparison (Target/Comparator/Outcome) for PLE where the CohortMethod package supports designing for multiple comparisons with multiple outcomes.
- The ATLAS designer supports a single analysis specification whereas CohortMethod supports designing with multiple analysis specifications

There are other gaps but the idea is here is that we'll close these gaps and provide editors that support the full range of options available in the packages.

# High-Level Design

## Study Specification Details

The object models that provide the details for how Estimation/Prediction studies will be specified have been detailed in the OpenAPI 3.0 format and are currently in a branch in the OHDSI Specifications repository: https://github.com/OHDSI/Specifications/tree/init. The specifications are also posted on SwaggerHub whicih makes viewing/navigating through them a bit easier: 

Estimation: https://app.swaggerhub.com/apis/anthonysena5/Estimation/0.9.0
Prediction: https://app.swaggerhub.com/apis/anthonysena5/Prediction/0.9.0

The shared dependencies for Estimation & Prediction have been externalized into their own specification documents:

ConceptSet: https://app.swaggerhub.com/apis/anthonysena5/ConceptSet/0.9.0 
CohortDefinition: https://app.swaggerhub.com/apis/anthonysena5/CohortDefinition/0.9.0
Cyclops: https://app.swaggerhub.com/apis/anthonysena5/Cyclops/0.9.0
FeatureExtraction: https://app.swaggerhub.com/apis/anthonysena5/FeatureExtraction/0.9.0

While OpenAPI 3.0 supports detailing the REST endpoints, I've chosen to do that as part of the WebAPI work.

These objects will be concretely defined in Java and stored in the StandardizedAnalayisAPI repository: https://github.com/OHDSI/StandardizedAnalysisAPI. These objects, in turn, will be utilized in WebAPI to facilitate the CRUD type operations.

## Hydra

[...]

PR #620, opened 2018-10-02
Estimation & Prediction Specification Editors

anthonysena:
Per #568, this is the WebAPI code for this feature. It is not ready for merging yet but putting a PR out here for review.


src/main/java/org/ohdsi/webapi/service/PredictionService.java
```diff
@@ -0,0 +1,238 @@
+package org.ohdsi.webapi.service;
```
pavgra: For this class all comments from EstimationService are actual

src/main/java/org/ohdsi/webapi/prediction/specification/AdaBoostSettings.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+public class AdaBoostSettings   {
+  @JsonProperty("nEstimators")
+  private Integer nEstimators = 50;
+
+  @JsonProperty("learningRate")
+  private Integer learningRate = 1;
+
+  @JsonProperty("seed")
+  private Float seed = null;
```
pavgra: 6! classes have the same field and logically comes from the same place - why is there no parent class like MLSettings??

src/main/java/org/ohdsi/webapi/prediction/PredictionAnalysis.java
```diff
@@ -0,0 +1,131 @@
...
+     * @param specification the specification to set
+     */
+    public void setSpecification(String specification) {
+        this.specification = specification;
+    }
+
+    /**
+     * @return the createdBy
+     */
+    public UserEntity getCreatedBy() {
```
pavgra: These fields are already defined in `CommonEntity`

src/main/java/org/ohdsi/webapi/featureextraction/specification/CovariateSettings.java
```diff
@@ -0,0 +1,2740 @@
+package org.ohdsi.webapi.featureextraction.specification;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.Objects;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown=true)
+public class CovariateSettings {
```
pavgra: Let's talk about combining this w/ Feature Analyses

src/main/java/org/ohdsi/webapi/cyclops/specification/Control.java
```diff
@@ -0,0 +1,670 @@
...
+  @JsonProperty("maxBoundCount")
+  private BigDecimal maxBoundCount = new BigDecimal(5);
+
+  @JsonProperty("autoSearch")
+  private Boolean autoSearch = true;
+
+  @JsonProperty("algorithm")
+  private String algorithm = "ccd";
+
+  @JsonProperty("attr_class")
```
pavgra: There is 15!, **fifteen**, classes with the field. *Why* they do not have a common parent which would encapsulate the field?

PR #627, opened 2018-10-04
Fixes #625

### LINKED ISSUE #625: Get Cohort Definition List leads to N+1 read problem
### Expected Results
When pulling up the cohort definition list, only the cohort definition table should be accessed.

### Actual Results
Looking in the logs, you see these hibernate queries:
```
Hibernate: select cohortdefi0_.id as id1_14_, cohortdefi0_.created_by_id as created_7_14_, cohortdefi0_.created_date as created_2_14_, cohortdefi0_.modified_by_id as modified8_14_, cohortdefi0_.modified_date as modified3_14_, cohortdefi0_.description as descript4_14_, cohortdefi0_.expression_type as expressi5_14_, cohortdefi0_.name as name6_14_ from ohdsi.cohort_definition cohortdefi0_
```
Which is good, that's all the fields necessary to render the view list.

However, the following queries are done (one for each cohort definition in the system):

```
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
Hibernate: select cohortdefi0_.id as id1_15_0_, cohortdefi0_.expression as expressi2_15_0_, cohortdefi0_.hash_code as hash_cod3_15_0_ from ohdsi.cohort_definition_details cohortdefi0_ where cohortdefi0_.id=?
... and so on for all cohorts
```

It appears the breaking change was introduced here:
https://github.com/OHDSI/WebAPI/commit/7e35d03fd8825aebcb9ed89c61defbe8f0364f9d#diff-3baa57b4ea135b915bde96e563ef720aL354

chrisknoll:
Use entityManager.createQuery to avoid N+1 load via definitionRepository.findAll().

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
+++ b/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
@@ -359,20 +359,26 @@ public class CohortDefinitionService extends AbstractDaoService {
   @Produces(MediaType.APPLICATION_JSON)
   public List<CohortDefinitionListItem> getCohortDefinitionList() {
 
-    return getTransactionTemplate().execute(transactionStatus ->
-      StreamSupport.stream(cohortDefinitionRepository.findAll().spliterator(), false).map(def -> {
-        CohortDefinitionListItem item = new CohortDefinitionListItem();
-        item.id = def.getId();
-        item.name = def.getName();
-        item.description = def.getDescription();
-        item.expressionType = def.getExpressionType();
-        item.createdBy = UserUtils.nullSafeLogin(def.getCreatedBy());
-        item.createdDate = def.getCreatedDate();
-        item.modifiedBy = UserUtils.nullSafeLogin(def.getModifiedBy());
-        item.modifiedDate = def.getModifiedDate();
-        return item;
-      }).collect(Collectors.toList())
-    );
+		ArrayList<CohortDefinitionListItem> result = new ArrayList<>();
+		// we use a direct query here because using repository.findAll leads to N+1 query performance (because the One-To-One mapping to 
+		// CohortDefinitionDetails is not lazy loaded when using JPA to fetch data.
+		String query = "SELECT cd.id, cd.name, cd.description, cd.expressionType, cu.login, cd.createdDate, mu.login, cd.modifiedDate FROM CohortDefinition cd"
+						+	" LEFT JOIN cd.createdBy cu LEFT JOIN cd.modifiedBy mu";
+		
+		List<Object[]> defs = entityManager.createQuery(query).getResultList();
+		for (Object[] d : defs) {
+			CohortDefinitionListItem item = new CohortDefinitionListItem();
+			item.id = (Integer)d[0];
+			item.name = (String)d[1];
+			item.description = (String)d[2];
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -359,20 +359,26 @@ public GenerateSqlResult generateSql(GenerateSqlRequest request) {
...
-        item.createdBy = UserUtils.nullSafeLogin(def.getCreatedBy());
-        item.createdDate = def.getCreatedDate();
-        item.modifiedBy = UserUtils.nullSafeLogin(def.getModifiedBy());
-        item.modifiedDate = def.getModifiedDate();
-        return item;
-      }).collect(Collectors.toList())
-    );
+		ArrayList<CohortDefinitionListItem> result = new ArrayList<>();
+		// we use a direct query here because using repository.findAll leads to N+1 query performance (because the One-To-One mapping to 
+		// CohortDefinitionDetails is not lazy loaded when using JPA to fetch data.
```
pavgra: It seems that a more correct way is to change the relationship type - we should use lazy everywhere

chrisknoll: From what I recall, even with specifying the relationship is lazy, one to one relationships are always fetched.  I recall finding some information on this here: https://stackoverflow.com/questions/1444227/making-a-onetoone-relation-lazy.

There is a lot of risk playing with relationship annotations, in my experience.

This solution does work, and so for this release we should go with it (provided there are no errors) and think about a longer solution for the next release.

chrisknoll: More recent information here:
https://stackoverflow.com/questions/50818502/hibernate-one-to-one-lazy-loading-with-spring

Seems that you still need to resort to bytecode instrumentation to get the lazy one-to-one case to work.  I'd prefer my 'direct query' approach (which is still using JPA standards) over bytecode instrumentation.

PR #630, opened 2018-10-05
Moves conceptSets to FeatureExtractionAnalysis level

wivern:


chrisknoll:
This migrations cript: V2.6.0.20181002110845__fe_analysis_conceptsets.sql originally placed the column in the fe_analysis_criteria table:

```
ALTER TABLE ${ohdsiSchema}.fe_analysis_criteria ADD conceptsets VARCHAR(MAX);
```

Now that the conceptSet is moved over to fe_analysis_concept_set, i think you should drop the conceptsets column on fe_analysis_criteria.

PR #646, opened 2018-10-11
Add job invalidation

aklochkova:


src/main/java/org/ohdsi/webapi/JobInvalidator.java
```diff
@@ -0,0 +1,30 @@
...
+@Component
+public class JobInvalidator {
+
+    @Autowired
+    JobExplorer jobExplorer;
+    @Autowired
+    JobRepository jobRepository;
+
+    @PostConstruct
+    private void invalidateGenerations(){
```
wivern: postConstruct is out of transaction
[look there](https://stackoverflow.com/questions/17346679/transactional-on-postconstruct-method)

PR #657, opened 2018-10-19
Reflection of TODO entries in pom.xml

NEONKID:
Set the default DBMS to PostgreSQL


NEONKID:
I will work on the master branch again.

PR #661, opened 2018-10-19
Added condition for tmp keytab deletion

aklochkova:


src/main/java/org/ohdsi/webapi/service/AbstractDaoService.java
```diff
@@ -12,6 +12,9 @@
 import com.odysseusinc.datasourcemanager.krblogin.KrbConfig;
 import com.odysseusinc.datasourcemanager.krblogin.RuntimeServiceMode;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
```
wivern: pls use existing slf4j logger

PR #678, opened 2018-10-29
Implements distribution criteria analyses for Cohort Characterization

wivern:
Implements stage 1 from https://github.com/OHDSI/WebAPI/issues/495#issuecomment-412667979


src/main/resources/resources/cohortcharacterizations/sql/distributionWithCriteria.sql
```diff
@@ -0,0 +1,67 @@
...
+  count(*) as value_as_int
+into #events_count
+from @temp_database_schema.@targetTable c
+  join @cdm_database_schema.@domain_table v on v.person_id = c.subject_id
+where
+  cohort_definition_id = @cohortId
+group by v.person_id;
+
+with
+  t1 as (
```
chrisknoll: Can these CTEs be named something more informative than 't1', 't2', 't3'?  Code readability is very important.

src/main/resources/resources/cohortcharacterizations/sql/prevalenceWithCriteria.sql
```diff
@@ -6,13 +6,18 @@ insert into @results_database_schema.cc_results (type, fa_type, covariate_id, co
...
+    sum.sum_value as count_value,
     case
       when totals.total > 0 then (sum.sum_value * 1.0 / totals.total * 1.0)
       else 0.0
     end as stat_value,
     @cohortId as cohort_definition_id,
     @executionId as cc_generation_id
-from (select count(*) as sum_value from @temp_database_schema.@targetTable) sum,
+from (select count(*) as sum_value from(
+         WITH qualified_events AS (
```
chrisknoll: WITH must be placed as a precursor to the statement, you can not put WITH inside a subquery.

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -349,47 +365,77 @@ private boolean ccHasPresetDistributionAnalyses() {
...
-                    .collect(Collectors.toList());
+            final String targetTable = "cohort_" + SessionUtils.sessionId();
+            CohortExpressionQueryBuilder.BuildExpressionQueryOptions options = createDefaultOptions(cohortDefinition.getId());
+            options.generateStats = true;
+            options.targetTable = options.resultSchema + "." + targetTable;
+            List<String> queries = new ArrayList<>();
+            String createCohortSql = sourceAwareSqlRender.renderSql(sourceId, CREATE_COHORT_SQL, TARGET_TABLE, targetTable);
+            String exprQuery = queryBuilder.buildExpressionQuery(cohortDefinition.getExpression(), options);
+            ConceptSet[] conceptSets = Stream.concat(
+                Arrays.stream(cohortDefinition.getExpression().conceptSets),
```
pavgra: Why do you need cohort specific codesets for totally separate criteria?

src/main/java/org/ohdsi/webapi/Constants.java
```diff
@@ -36,4 +40,64 @@
     String ROLE_GROUP_MAPPING = "roleGroupMapping";
     String PRESERVE_ROLES = "preserveRoles";
   }
+
+  Map<StandardFeatureAnalysisDomain, String> DOMAIN_TABLES = ImmutableMap.<StandardFeatureAnalysisDomain, String>builder()
```
pavgra: Where is this used?

src/main/java/org/ohdsi/webapi/Constants.java
```diff
@@ -36,4 +40,64 @@
...
+          .put(StandardFeatureAnalysisDomain.CONDITION, "condition_occurrence")
+          .put(StandardFeatureAnalysisDomain.DEVICE, "device_exposure")
+          .put(StandardFeatureAnalysisDomain.DRUG, "drug_exposure")
+          .put(StandardFeatureAnalysisDomain.MEASUREMENT, "measurement")
+          .put(StandardFeatureAnalysisDomain.OBSERVATION, "observation")
+          .put(StandardFeatureAnalysisDomain.PROCEDURE, "procedure_occurrence")
+          .put(StandardFeatureAnalysisDomain.VISIT, "visit_occurrence")
+          .build();
+
+  Map<StandardFeatureAnalysisDomain, String> DOMAIN_ID_FIELDS = ImmutableMap.<StandardFeatureAnalysisDomain, String>builder()
```
pavgra: Where is this used?

pavgra:
@wivern ,

What I observed when pulled your PR:
1. Distribution criteria didn't work - failed due to incorrect SQL
1. Logic for distribution analysis wasn't correct - counts weren't matching similar preset analysis ("Visit Count Short Term" was used to compare):
image

1. Prevalence analysis relied on observation periods which is also not correct

What was done:
1. As we discussed, but for some reason it was not done, the WindowedCriteria was introduced - https://github.com/OHDSI/circe-be/pull/62
1. Code for Distribution analyses was fixed and now count and min / max values for Distribution Criteria match preset analysis results:
image

To do:
1. Distribution analysis should store WindowedCriteria as design instead of CriteriaGroup
1. P10, P25, Median, P75, P90 values should be fixed
1. Both Prevalance and Distribution Criteria should be verified and match results given by Preset analyses
1. Styling for Distirubtion tables should be fixed - columns shouldn't jump based on content
1. Following UI elements should be removed for Distribution analysis since WindowedCriteria doesn't support them (they don't make sense for that analysis type):
image
1. Add padding of `1rem` around Prevalance / Distribution selector
1. Resolve todos in code
1. Refactoring

src/main/resources/resources/cohortcharacterizations/sql/distributionWithCriteria.sql
```diff
@@ -31,24 +26,33 @@ with
...
+      total_cohort_count.cnt as population_size
+    from #events_count, total_cohort_count
+    group by total_cohort_count.cnt
+  ),
+  event_prep as (select row_number() over (order by value_as_int) as rn, value_as_int, count(*) as people_count from #events_count group by value_as_int),
+  events_dist as (
+    select s.value_as_int, sum(p.people_count) as people_count
+    from event_prep s join event_prep p on p.rn <= s.rn group by s.value_as_int
   ),
   events_p10_value as (
```
chrisknoll: We have a nice sql script that calculates the P10 - P90 distributions if you are to use it:
https://github.com/OHDSI/Achilles/blob/master/inst/sql/sql_server/analyses/106.sql#L17

You may want to adapt that approach to calculating these values.

chrisknoll:
Found some issues:

1. Mangling the parameter:

https://github.com/OHDSI/WebAPI/blob/cc_first_stage/src/main/java/org/ohdsi/webapi/service/CohortGenerationService.java#L168

```
    if (targetTable.indexOf('.') != -1) {
      String[] targetParts = targetTable.split("\\.");
      builder.addString(TARGET_DATABASE_SCHEMA, targetParts[0]);
      builder.addString(TARGET_TABLE, targetParts[1]);
    } else {
      builder.addString(TARGET_DATABASE_SCHEMA, resultsTableQualifier);
      builder.addString(TARGET_TABLE, targetTable);
    }
```
You shouldn't need to strip the target table apart.  From the source table,t he schema may have the format of DB.schema. So you don't want to split anything like that out.  

2.  Generating local cohort is overwriting the cohort_inclusion stats table.  

You can see here in the cohort generation query:
https://github.com/OHDSI/circe-be/blob/master/src/main/resources/resources/cohortdefinition/sql/generateCohort.sql#L121

there's a parameter that will not write the generation statistcis (like the inclusion rule impact). You should pass the parameter that sets this condition false:
```{@generateStats != 0}?{```

Which is set using this class:
https://github.com/OHDSI/circe-be/blob/master/src/main/java/org/ohdsi/circe/cohortdefinition/CohortExpressionQueryBuilder.java#L78

**Comment by chrisknoll** [comment]
Error when generating characterization:
``The function 'ROW_NUMBER' must have an OVER clause with ORDER BY.``

I located the expression in the output here:
```
WITH qualified_events  AS (
  SELECT ROW_NUMBER() OVER () AS event_id, E.subject_id AS person_id, E.cohort_start_date AS start_date, E.cohort_end_date AS end_date, OP.observation_period_start_date AS op_start_date, OP.observation_period_end_date AS op_end_date
```

File: src/main/resources/resources/cohortcharacterizations/sql/distributionWithCriteria.sql [hunk 7]
```diff
@@ -0,0 +1,82 @@
...
+  )
+insert into @results_database_schema.cc_results(type, fa_type, covariate_id, covariate_name, analysis_id, analysis_name, concept_id,
+  cohort_definition_id, cc_generation_id, count_value, min_value, max_value, avg_value, stdev_value, p10_value, p25_value, median_value, p75_value, p90_value)
+select
+  'DISTRIBUTION' as type,
+  'CRITERIA' as fa_type,
+  @covariateId as covariate_id,
+  '@covariateName' as covariate_name,
+  @analysisId as analysis_id,
+  '@analysisName' as analysis_name,
```
chrisknoll: This could lead to sql injection, since `@analysisName` can come from user input (for feature criteria).  What we have done with this situation in the past is that we have a separate metadata table for an execution where we populate the 'analysis lookup' table prior to analysis with the id -> name mapping, but inserting into this table is done via prepared statements (ie: insert into {table} values ?,?,?} .

You can see how the 'inclusion rules' for cohort definitions (where the inclusion rule name is user input, so be protected during insertion) is here: https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java#L107

chrisknoll:
Error when generating related to the 'event table' that's passed into the feature criteria (I think):

```
SQLServerException: Invalid object name '#qualified_events'.
```

The query in question has a cte named `qualified_events`, but the later subquery is referencing the event table as `#qualified_events`.

```
CREATE TABLE  #events_count
  WITH (LOCATION = USER_DB, DISTRIBUTION = REPLICATE) AS
WITH qualified_events  AS (
  SELECT ROW_NUMBER() OVER (partition by E.subject_id order by E.cohort_start_date) AS event_id, E.subject_id AS person_id, E.cohort_start_date AS start_date, E.cohort_end_date AS end_date, OP.observation_period_start_date AS op_start_date, OP.observation_period_end_date AS op_end_date
  FROM [DEV_CDM_Optum_Extended_SES_v515].cknoll1.cohort_uq2xeajj E
    JOIN [DEV_CDM_Optum_Extended_SES_v515].dbo.observation_period OP ON E.subject_id = OP.person_id AND E.cohort_start_date >= OP.observation_period_start_date AND E.cohort_start_date <= OP.observation_period_end_date
  WHERE cohort_definition_id = ?
)
 SELECT
v.person_id as person_id,
  count(*) as value_as_int

FROM
( SELECT 0 as index_id, p.person_id, p.event_id
FROM #qualified_events P
INNER JOIN
(
  -- Begin Visit Occurrence Criteria
select C.person_id, C.visit_occurrence_id as event_id, C.visit_start_date as start_date, C.visit_end_date as end_date, C.visit_concept_id as TARGET_CONCEPT_ID, C.visit_occurrence_id
from 
(
  select vo.* 
  FROM [DEV_CDM_Optum_Extended_SES_v515].dbo.VISIT_OCCURRENCE vo

) C


-- End Visit Occurrence Criteria

) A on A.person_id = P.person_id and A.START_DATE >= DATEADD(day,-90,P.START_DATE) AND A.START_DATE <= DATEADD(day,-1,P.START_DATE) ) v
group by v.person_id;
``

Note the subquery refers to the event table as `#qualified_events`.

File: src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java [hunk 8]
```diff
@@ -346,68 +264,108 @@ private boolean ccHasPresetDistributionAnalyses() {
...
-            List<String> queries = new ArrayList<>();
-            CriteriaGroup expression = feature.getExpression();
+            Long conceptId = 0L;
+            String queryFile;
+            String groupQuery;
+            String[] paramNames = CRITERIA_PARAM_NAMES.toArray(new String[0]);
+
+            if (CcResultType.PREVALENCE.equals(analysis.getStatType())) {
+                queryFile = COHORT_STATS_QUERY;
+                groupQuery = queryBuilder.getCriteriaGroupQuery(((FeAnalysisCriteriaGroupEntity)feature).getExpression(), "#qualified_events");
```
chrisknoll: Check these queries for the proper specification for the events table.  There is an error occurring where the query uses a CTE called 'qualified events' but the generated subquery thinks the event table is `#qualified_events`

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -346,68 +264,108 @@ private boolean ccHasPresetDistributionAnalyses() {
...
+            } else if (CcResultType.DISTRIBUTION.equals(analysis.getStatType())) {
+                queryFile = COHORT_DIST_QUERY;
+                if (feature instanceof FeAnalysisWindowedCriteriaEntity) {
+                  WindowedCriteria criteria = ((FeAnalysisWindowedCriteriaEntity) feature).getExpression();
+                  criteria.ignoreObservationPeriod = true;
+                  groupQuery = queryBuilder.getWindowedCriteriaQuery(criteria, "#qualified_events");
```
chrisknoll: check `#qualified_events`

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -346,68 +264,108 @@ private boolean ccHasPresetDistributionAnalyses() {
...
+                } else if (feature instanceof FeAnalysisDemographicCriteriaEntity) {
+                  DemographicCriteria criteria = ((FeAnalysisDemographicCriteriaEntity)feature).getExpression();
+                  groupQuery = queryBuilder.getDemographicCriteriaQuery(criteria, "#qualified_events");
```
chrisknoll: check `#qualified_events`

Issue #677, opened 2018-10-26
Incidence Rates' generation is broken


aklochkova:
`failed to lazily initialize a collection of role: org.ohdsi.webapi.ircalc.IncidenceRateAnalysis.executionInfoList, could not initialize proxy - no Session`

PR #682, opened 2018-10-30
Fixed retrieving of incidence rates with execution info

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/677


src/main/java/org/ohdsi/webapi/ircalc/IncidenceRateAnalysisRepository.java
```diff
@@ -25,6 +25,8 @@
 public interface IncidenceRateAnalysisRepository extends CrudRepository<IncidenceRateAnalysis, Integer> {
   
   @Query("select ira from IncidenceRateAnalysis AS ira LEFT JOIN FETCH ira.details as d")          
-  public Iterable<IncidenceRateAnalysis> findAll();
-  
+  Iterable<IncidenceRateAnalysis> findAll();
+
+  @Query("SELECT ira FROM IncidenceRateAnalysis AS ira JOIN FETCH ira.executionInfoList WHERE ira.id = ?1")
```
pavgra: Why not `LEFT JOIN`? As soon as you query for IR w/o executions - this returns null. And then you get NPE at `IRAnalysisService.java` line 494

pavgra:
IRs fail to be displayed also when a DS used for generation was deleted:
image

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -280,8 +284,10 @@ public Response delete(@PathParam("sourceId") Integer sourceId) throws Exception
     Source source = sourceRepository.findBySourceId(sourceId);
     if (source != null) {
       final String sourceKey = source.getSourceKey();
+      List<ExecutionInfo> executions = irExecutionInfoRepository.findBySource(source);
+      irExecutionInfoRepository.delete(executions);
       sourceRepository.delete(source);
-        publisher.publishEvent(new DeleteDataSourceEvent(this, sourceId, source.getSourceName()));
+      publisher.publishEvent(new DeleteDataSourceEvent(this, sourceId, source.getSourceName()));
```
chrisknoll: This change seems out of place inside a PR that deals with retrieving incidence rate execution info?

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -59,6 +61,8 @@
   private Environment env;
   @Autowired
   private ApplicationEventPublisher publisher;
+  @Autowired
+  private IRExecutionInfoRepository irExecutionInfoRepository;
```
pavgra: Where is it used?

src/main/java/org/ohdsi/webapi/service/SourceService.java
```diff
@@ -281,7 +285,7 @@ public Response delete(@PathParam("sourceId") Integer sourceId) throws Exception
     if (source != null) {
       final String sourceKey = source.getSourceKey();
       sourceRepository.delete(source);
-        publisher.publishEvent(new DeleteDataSourceEvent(this, sourceId, source.getSourceName()));
+      publisher.publishEvent(new DeleteDataSourceEvent(this, sourceId, source.getSourceName()));
```
pavgra: Dummy diff

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -104,6 +105,7 @@
  * @author Chris Knoll <cknoll@ohdsi.org>
  */
 @Path("/ir/")
+@Singleton
```
pavgra: Where is the Service created more than once? Maybe we should fix cause but not consequences?

chrisknoll:
Hi,
 This PR is changing to the structore of Service -> ServiceImpl. 

We haven't talked about this project structure, and I am opposed to this needless layer.  We need to review what was done on CC and pathways to determine which elements we liked, which elements we didn't like, but we are not going to push that structure to every other part of the application.

This is something we can review on the WebAPI call.

chrisknoll:
Hi, @aklochkova ,
  I need to step-back from my previous objection:  @pavgra walked me through the Spring core code that is having the challenge with handling javax.ws (Jersey) endpoints and the interop with spring:  Jersey and Spring both have their own dependency injections, and in this particular case the IRAnalyssisService extending an interface is making the mechanism in Spring create an instance of the service per request, which is leading ot the post-create call being invoked multiple times, which is leading to the incomplete job cleanup being fired multiple times.  All this is bad, and the simplest fix was to extract all the Jersey annotations out into a single interface so that the service is not re-initialized each request.


So, this change isn't to push the controller-service-serviceImpl pattern, but rather to just fix a problem that was introdued when adding the 'GeneratesNotifiation' as an implemented interface...fine.

Note for 3.0:  the cleanup of the incomplete services should probably not have been implemented as a function of each service. Rather there should have been a single bean that gets invoked once-per-startup of WebAPI that performs the cleanup across all services.  This separates the concern from 'Do this activity at startup' of this initialization bean from 'Handle requests from clients' that is the job of each of the services.

This last point isn't placing blame at anyone, just wanted to remember this idea for when we do the larger-scale refactoring for webAPI 3.0.

PR #684, opened 2018-10-31
check if security enabled on server

vantonov1:
Notifications icon will show unread notifications at every login, because there is no user entity to store last viewed time. Should we create guest/anonymous entity in DB (and return it as a current user from permissions manager, instead of exception)?

### DIFF
--- a/src/main/java/org/ohdsi/webapi/job/NotificationServiceImpl.java
+++ b/src/main/java/org/ohdsi/webapi/job/NotificationServiceImpl.java
@@ -6,17 +6,16 @@ import org.ohdsi.webapi.shiro.Entities.UserRepository;
 import org.ohdsi.webapi.shiro.PermissionManager;
 import org.springframework.batch.admin.service.SearchableJobExecutionDao;
 import org.springframework.batch.core.JobExecution;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


vantonov1:
closes #1082

anthonysena:
@vantonov1 @pavgra - one thing to consider regarding the notifications area: this was originally done using localStorage on the client side when security was disabled since the design was to make the area personalized to the current user. If I'm reading this PR correctly, the change here would give all anonymous users the same list of jobs which doesn't meet the original need for this feature. Perhaps in the case of disabled security we should leverage the localStorage as was done originally?

vantonov1:
local storage is not the best solution here, because user could login from another comuter, or just from another browser.

anthonysena:
That is true regarding localStorage - it will be machine+browser specific. However, I'm unsure how else you'd accomplish this on the server when security is disabled? Perhaps it is worth a quick discussion with @pavgra and then we can post any decisions here for others to review.

fdefalco:
Going back to the ecommerce days, you could store an anonymous persistent browser cookie with an identifier.. we used to do that for anonymous shopping carts for people who were shopping but not yet registered, it would persist across sessions.  So when then did come back and potentially register we could track conversions and not lose their shopping carts.  We could use a similar approach for environments without security.  Its true, if the person goes to a different machine, browser or cookie it does get lost, but, it still allows for some level of job tracking in environments without authentication.

vantonov1:
ok, after investigation it was decided to store last notification viewed date in local storage for insecure mode only

PR #692, opened 2018-11-13
Updated 2 DDLs to play nicer with PDW

chrisknoll:
Had to merge the original PR over to a WebAPI local branch for testing/updating.


chrisknoll:
There are other isues with PDW: the create index statements for PDW have default columns of GETDATE() specified, but you can't have a column default value as an expression in PDW.

mpozhidaeva:
@chrisknoll - could you please clarify for me what value (from what parameter) for @vocab_database_schema (init_concept_hierarchy.sql) should be used?

chrisknoll:
> @chrisknoll - could you please clarify for me what value (from what parameter) for @vocab_database_schema (init_concept_hierarchy.sql) should be used?

In the ddl service here:
https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/service/DDLService.java#L92

It only supports the following parameters: dialect (for sql translation) and schema (to replace the name of the `@results_schema` token.   There is no parameter that lets a user specify the schema that their vocabulary tables are in. So we'd need to allow the user to pass in the value as a new parameter, and the SQLRender calls need to pass that new paramater in to replace `@vocabulary__database_schema`.

The reason why this issue came up is:  in the past, it wasn't expected that you'd need to read data in from the cdm/vocabulary schema to initialize the results schema.  Ie: to create your results schema, you just need to create some tables and _maybe_ insert some records into a lookup table (like heracles_analysis).  But, with the introduction of concept_hierarchy table, it gets populated based on the values in the vocabulary's concept_ancestor.  So now we need to pass in `@vocabulary_database_schema`.

chrisknoll:
@mpozhidaeva : you mentioned that you need changes in this PR, but I believe we have conflicts due to seperate work that you may have worked on that got merged into master.

Would you mind pulling this branch and resolving conflicts such that the changes in this PR still work for your other PR that was merged (I'm sorry I can't find the reference for it right now, but you mentioend that this PR had changes that your other PR needed to work)

chrisknoll:
Ok, I'll take a look and see where the conflicts are and see if i can resolve them.

PR #693, opened 2018-11-14
Issue 1078 ui improvements

wivern:
closes OHDSI/Atlas#1078


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -83,9 +80,11 @@
 
     private static final String GENERATION_NOT_FOUND_ERROR = "generation cannot be found by id %d";
     private static final String[] GENERATION_PARAMETERS = {"cohort_characterization_generation_id"};
+    private static final String[] PREVALENCE_STATS_PARAMS = {"cdm_database_schema", "cdm_results_schema", "cc_generation_id", "analysisId", "cohort_id", "covariate_id"};
```
pavgra: image

pavgra:
WebAPI uses local Feature Analysis ID instead of FE analysis ID to query data for "Explore" for Preset analysis. The preset analyses use FE IDs in DB. Therefore, there are empty results

pavgra:
Explore doesn't seem to work properly. Even for concepts which represent ingredients and have biggest counts, there is zero returned for descendants

PR #695, opened 2018-11-15
Fixes Number of persons with continuous observation in each year

### LINKED ISSUE #318: Heracles 109   Number of persons with continuous observation in each year
https://github.com/OHDSI/WebAPI/blob/c92863dbdd5cec90c5e17c29368b91f9b703c6b9/src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql#L644

> -- 109   Number of persons with continuous observation in each year

What we want is all the possible calendar years between min(observation_period_start_date) and max(observation_period_end_date) for the population in cohort.

All this SQL seems to be doing is

```
  SELECT DISTINCT 
  YEAR(observation_period_start_date) AS obs_year,
	DATEFROMPARTS(YEAR(observation_period_start_date),1,1) as obs_year_start,
	DATEFROMPARTS(YEAR(observation_period_start_date),12,31) as obs_year_end
  INTO
  #temp_dates_1
  from @CDM_schema.PERSON p1
  inner join (select subject_id, cohort_definition_id as cohort_definition_id from @results_schema.COHORT where cohort_definition_id in (@cohort_definition_id)) c1
  on p1.person_id = c1.subject_id
  inner join
  @CDM_schema.observation_period op1
  on p1.person_id = op1.person_id
  ;
```
creating a list of unique calendar years corresponding to observation_period_start_date for any person in the cohort. This approach will skip some calendar years, if the observation_periods pans across years.

One approach is to do something like this for all dates between min(observation_period_start_date) and max(observation_period_start_date)

https://stackoverflow.com/questions/21425546/how-to-generate-a-range-of-numbers-between-two-numbers
where the low number is min(year(observation_period_start_date)) and high number is max(year(observation_period_start_date))

Same applies to Heracles 110, 116, 117,

wivern:
closes #318


src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql
```diff
@@ -998,17 +1019,26 @@ DROP TABLE #raw_107;
...
+		op_date_range AS (
+			select
+				MIN(observation_period_start_date) AS op_start_date,
+				MAX(observation_period_end_date) AS op_end_date
+			FROM @CDM_schema.PERSON p1
+			inner join (select subject_id, cohort_definition_id, cohort_start_date, cohort_end_date from #HERACLES_cohort) c1
+			on p1.person_id = c1.subject_id
+			inner join
+			@CDM_schema.observation_period op1
+			on p1.person_id = op1.person_id)
```
pavgra: Cannot we move repetitive code chunks into temp tables to reduce code duplication?

wivern: we can, but each report could be run separately. so that temp tables has to be present on each report with no profit.

Issue #317, opened 2018-02-13
Heracles Analysis id 108


gowthamrao:
https://github.com/OHDSI/WebAPI/blob/c92863dbdd5cec90c5e17c29368b91f9b703c6b9/src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql#L611

>   -- 108   Number of persons by length of observation period, in 30d increments

Should it be
>  -- 108   Number of persons by length of **first** observation period, in 30d increments

In both cases, is the 108 SQL correct? 

https://github.com/OHDSI/WebAPI/blob/c92863dbdd5cec90c5e17c29368b91f9b703c6b9/src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql#L620

Maybe this is more accurate

`inner join (select distinct subject_id, cohort_definition_id from #HERACLES_cohort) c1 `

PR #705, opened 2018-11-21
Fixed persons counts in Heracles analyses

mpozhidaeva:
Closes https://github.com/OHDSI/WebAPI/issues/317


chrisknoll:
Hi, @mpozhidaeva ,

Let's drop the flyway migration scripts for the heracles_analysis tables. As of 2.4, we've introduced the DDL service that has all the necessary instructions to build the results schema. So, we don't need to maintain the WebAPI's schema that has some of the results tables schemas copied into it.

So let's drop the flyway migrations from this PR.

src/main/resources/resources/cohortanalysis/sql/runHeraclesAnalyses.sql
```diff
@@ -825,7 +831,7 @@ DROP TABLE #raw_107;
   COUNT_BIG(DISTINCT p1.person_id) AS count_value
```
pavgra: Why the distinct left here and in some places below?

mpozhidaeva: Join to observation_period may cause "dublicates"

PR #715, opened 2018-11-29
prevent truncation of floats

### LINKED ISSUE #698: Data Density Plot -> Records Per Person Looks Rounded
### Expected behavior

The records per person on the Data Sources -> Data Density should show values between 0 and 1.

### Actual behavior

It appears that numbers are rounded to 0 or 1 based on this screenshot:

image

Open a data source, navigate to data density and review the 'Records per Person' plot.

fdefalco:
closes #698


chrisknoll:
Note for WebAPI 3.0:

we should change this result set to have the name of the column as 'record_count_pp' (pp = per person, which is a naming convention that we have adopted elsewhere).  I think the source of this bug came from the interpretation that the value was an actual count (int) and not some kind of average across all people.  But, to change the field now would be API breaking (the field values would change their names) so probably not appropriate for 2.7).

PR #716, opened 2018-11-29
Record count retrieval fixes

### LINKED ISSUE #714: Record count cache missing device counts
Per OHDSI/Atlas#828, we're missing record counts for devices. Looking at the code in:

https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cdmresults/sql/loadConceptRecordCountCache.sql

It appears we're missing the device analyses from Achilles: 

https://github.com/OHDSI/Achilles/blob/master/inst/csv/achilles/achilles_analysis_details.csv

Specifically: `2101 :: Number of device exposure records, by device_concept_id` and `2105 :: Number of device exposure records, by device_concept_id by device_type_concept_id`

anthonysena:
- Per this discussion: https://github.com/OHDSI/WebAPI/commit/c11f1425c27f7301aaede2922120afd37bd7558b#comments: In reviewing the code that was part of this commit, I'm unsure how the new caching mechanism requires the need for concept ID's with 0 records (RC) and 0 descendant record counts (DRC). To that end, I've reverted back the HAVING clause to the query used to retrieve the concept counts. My thinking here is that if a concept ID does not exist in the cache _after it has been warmed_ then it is assumed to have no RC/DRC. Furthermore, I've revised the CDMResultsService to check the status of the cache and if it has not been warmed, it will attempt to query the Achilles results table for the record counts. In addition, I revised the function to chunk up the concept IDs that are passed into the PreparedStatementRenderer to respect the limits imposed by the RDBMS hosting the CDM's results schema.

Tagging @pavgra for thoughts and comments since I could be mis-understanding his revised implementation in this area.

- Fixes #714 by adding in device related Achilles results analysis IDs.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
+++ b/src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
@@ -19,7 +19,10 @@ import org.ohdsi.webapi.job.JobExecutionResource;
 import org.ohdsi.webapi.job.JobTemplate;
 import org.ohdsi.webapi.source.Source;
 import org.ohdsi.webapi.source.SourceDaimon;
+import org.ohdsi.webapi.util.PreparedSqlRender;
 import org.ohdsi.webapi.util.PreparedStatementRenderer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.JobParametersBuilder;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
@@ -33,6 +36,7 @@ import org.springframework.stereotype.Component;
 @Path("/cdmresults")
 @Component
 public class CDMResultsService extends AbstractDaoService {
+    private final Logger logger = LoggerFactory.getLogger(CDMResultsService.class);
 
     private CDMResultsAnalysisRunner queryRunner = null;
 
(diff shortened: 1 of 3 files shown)


pavgra:
@anthonysena , my approach allowed the cache to be warmed as it used (and not wait till the global warm job is finished which quite regularly takes a lot of time). The behavior of the job was not affected. So e.g.:
- start Atlas
- query Vocabularies for "Aspirin" (you will wait e.g. 15 sec)
- query it for Aspirin once more:
    - in my version you will get results immediately
    - in your version - you will have to wait that 15 sec again

Regarding the cache size if we store zeros: yes, 1.5m differs from 300k, but for those 1.2m you store only 2 integers (key and value), which equals to 8 (2 x 4bytes) bytes, so not that much additional memory. And for the access you use HashMap, so you will rarely go out of O(1), not so much collisions should appear for linear IDs used as keys

anthonysena:
Thanks for that @pavgra - appreciate the details around the design thinking here. To that end, I removed (again) the HAVING clause to bring it back to your original design. 

Please take a look at the revisions made to the CDMResultsService.java. One of the behaviors I noticed when testing is that regardless of the state of the source's results cache, some search terms were consistently querying the database for concepts that were not in the cache which seemed odd. The reason it was happening in my testing is that I'm sourcing a different vocabulary than the one that is used in the CDM I've configured for record counts. As a result, there are concepts in that vocabulary that do not exist in that CDM and are never returned by the `getConceptRecordCount` query and ultimately are never in the cache. So, I added in a check for the sourceCache.warm flag so if we know that the cache job is complete, there should be no need to check record counts against the CDM directly.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -82,15 +86,31 @@ public void warmCaches(){
...
     }
+    
+    protected List<SimpleEntry<Integer, Long[]>> executeGetConceptRecordCount(Integer[] identifiers, Source source, CDMResultsCache sourceCache) {
+        List<SimpleEntry<Integer, Long[]>> returnVal = new ArrayList<>();
+        if (identifiers.length == 0) {
+            return returnVal;
+        } else {
+            int parameterLimit = PreparedSqlRender.getParameterLimit(source);
+            if (parameterLimit > 0 && identifiers.length > parameterLimit) {
+                returnVal = executeGetConceptRecordCount(Arrays.copyOfRange(identifiers, parameterLimit, identifiers.length), source, sourceCache);
```
pavgra: Shouldn't there be a loop like

```
for (int i = 0; i < identifiers.length / parameterLimit; i++)
```

to query all identifiers?

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -82,15 +86,31 @@ public void warmCaches(){
             .filter(Objects::nonNull)
             .collect(Collectors.toList());
 
-        if (notCachedRecordIds.size() > 0) {
+        if (!sourceCache.warm && notCachedRecordIds.size() > 0) {
```
pavgra: `!sourceCache.warm` won't let to use the on-demand cache results

anthonysena: The .map operation above (line 77) will allow for the use the on-demand cache results and if any concepts are not found, they will be part of the `notCachedRecordIds`. If the cache has not completed the 'warming' operation, then it will attempt to retrieve the counts for the non-cached records. Let me know if this addresses your concern regarding the on-demand cache results.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
anthonysena: I'm using a recursive approach as is done in a few other places in the code base where this type of operation occurs for setting large numbers of parameters on the PreparedSqlStatement.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
pavgra: Ah, my bad, was too quick in conclusions: I see, the stream addresses it, right. But, nevertheless, I cannot understand why would you need this additional condition? If your cached was warmed properly, you won't ever get `notCachedRecordIds.size() > 0`, so why do you need this extra check then? But if smth went wrong inside global cache warming and the cache Map was not populated, you will stop getting the results even from DB

PR #723, opened 2018-12-17
Added option to disable concept_hierarchy in DDL Service

pavgra:
Since concept_hierarchy may be generated through Achilles, it is useful to have option to not print concept_hierarchy DDL + DML in results of DDL Service

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/DDLService.java
+++ b/src/main/java/org/ohdsi/webapi/service/DDLService.java
@@ -20,6 +20,7 @@ package org.ohdsi.webapi.service;
 
 import static org.ohdsi.webapi.service.SqlRenderService.translateSQL;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
@@ -47,7 +48,6 @@ public class DDLService {
 		"/ddl/results/cohort_inclusion_result.sql",
 		"/ddl/results/cohort_inclusion_stats.sql",
 		"/ddl/results/cohort_summary_stats.sql",
-		"/ddl/results/concept_hierarchy.sql",
 		"/ddl/results/feas_study_inclusion_stats.sql",
 		"/ddl/results/feas_study_index_stats.sql",
 		"/ddl/results/feas_study_result.sql",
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/service/DDLService.java
```diff
@@ -89,8 +93,18 @@
 	@GET
 	@Path("results")
 	@Produces("text/plain")
-	public String generateResultSQL(@QueryParam("dialect") String dialect, @DefaultValue("results") @QueryParam("schema") String schema) {
-            return generateSQL(dialect, "results_schema", schema, RESULT_DDL_FILE_PATHS, RESULT_INIT_FILE_PATHS, RESULT_INDEX_FILE_PATHS);
+	public String generateResultSQL(
```
chrisknoll: This breaks the API. anyone who has called the old signature will have failures calling this new sig. Should we be concerned?

To maintain backwards compatibility, we'd need to leave generateResultSQL as-is, and introduce a new function that accepts the additional parameter.

If this is not worth the effort, I have an ask:  can we add 2 additional params: 'cdmSchema' and 'vocabSchema'.  We can ignore these params for now, but they place holders for accepting those params and creating the resultsDDL to reference those schemas. I only ask this since we're making a change to this function anyway.

chrisknoll: I think we may be thinking of different things: you are talking about the endpoint not passing the value as a ``@QueryParam`` and yes, that doesn't break anything between releases, people who were previously calling that endpoint without the new param will continue to work.

My concern was in cases where someone has code that calls the method getResultSql() directly via java, and their code will break if they call that method via:

```generateResultSql("someDialect", "someSchema);```

They would have up date their code to pass in the new param (Unless Java now will let you call a method without the parameter if there is a ```@DefaultValue``` in the sig? I don't think so, I think ```DefaultValue``` is just for endpoints).

So, we don't have to pursue this, since there is probably no external consumer of this class that will be impacted by this change. I thought it would be an exercise in maintaining compatibility (at the Java signature level, not the rest-endpoint level).  But, again, I'm not concerned about people calling this method directly.

Issue #690, opened 2018-11-09
Cohort pathways errors


chen-regen:
Per Anthony @anthonysena , I am reporting the issues here:
Tag 2.6.0 "Cohort Pathways" tool fails against MSSQL and Oracle.
Below is the stack trace from Oracle:
```
Hibernate: select daimons0_.SOURCE_ID as SOURCE_ID5_50_0_, daimons0_.SOURCE_DAIMON_ID as SOURCE_DAIMON_ID1_50_0_, daimons0_.SOURCE_DAIMON_ID as SOURCE_DAIMON_ID1_50_1_, daimons0_.DAIMON_TYPE as DAIMON_TYPE2_50_1_, daimons0_.PRIORITY as PRIORITY3_50_1_, daimons0_.SOURCE_ID as SOURCE_ID5_50_1_, daimons0_.TABLE_QUALIFIER as TABLE_QUALIFIER4_50_1_ from OHDSI.source_daimon daimons0_ where daimons0_.SOURCE_ID=?
2018-11-08 08:45:11.771 DEBUG taskExecutor-2 org.springframework.jdbc.core.JdbcTemplate -  - Executing SQL statement [create table OHDSI.cohort_ic5a78e3
(
  COHORT_DEFINITION_ID int NOT NULL,
  SUBJECT_ID NUMBER(19) NOT NULL,
  cohort_start_date date NOT NULL,
  cohort_end_date date NOT NULL
);]
2018-11-08 08:45:12.254 DEBUG taskExecutor-2 org.springframework.orm.jpa.JpaTransactionManager -  - Closing JPA EntityManager [org.hibernate.jpa.internal.EntityManagerImpl@12f7eac3] after transaction
2018-11-08 08:45:12.254 DEBUG taskExecutor-2 org.springframework.orm.jpa.EntityManagerFactoryUtils -  - Closing JPA EntityManager
2018-11-08 08:45:12.291 ERROR taskExecutor-2 org.springframework.batch.core.step.AbstractStep -  - Encountered an error executing step generatePathwayAnalysis.createCohortTable in job generatePathwayAnalysis
org.springframework.jdbc.BadSqlGrammarException: StatementCallback; bad SQL grammar [create table OHDSI.cohort_ic5a78e3
(
  COHORT_DEFINITION_ID int NOT NULL,
  SUBJECT_ID NUMBER(19) NOT NULL,
  cohort_start_date date NOT NULL,
  cohort_end_date date NOT NULL
[...]

PR #725, opened 2018-12-19
Fixed pathway execution scripts

aklochkova:
Fixes: https://github.com/OHDSI/WebAPI/issues/690


src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -88,8 +88,8 @@ into
       |B--|B--|
 */
 
-IF OBJECT_ID('tempdb..#split_overlapping_events', 'U') IS NOT NULL
-DROP TABLE #split_overlapping_events;
+IF OBJECT_ID('tempdb..#split_overlay_events', 'U') IS NOT NULL
+DROP TABLE #split_overlay_events;
```
pavgra: What's wrong with old naming? As for me, it was more self-descriptive

pavgra:
@chrisknoll, the table creation is prepended with `IF OBJECT_ID('@temp_database_schema.cohort_inclusion', 'U') IS NULL` so no issue should have occurred if the split and translation were handled properly: @aklochkova did the translation in the wrong way (`IF NOT EXISTS` translation rule relies on semi-colon at the end of a statement, whereas she uses `splitSql` which cuts out the semicolons). @aklochkova , pls address this

Regarding the table creation - it is part of "kinda"-temp tables set (used in process of isolated cohort calculation to work out temp tables between transactions; see `temp_database_schema`) which we discussed and implemented earlier, so no changes from a business logic perspective or tech paradigms introduced in this PR.

chrisknoll:
@pavgra , none of the chort_inclusion tables, cohort_summary_stats, etc are required to instantiate a cohort for characterization purposes.  Those temp tables are not necessary. So, we should remove those temp tables, and if they are required somewhere, we should add the appropriate logic to bypass the 'inclusion impact statistics' part of cohort generation because that just adds needless time to the overall cohort generation.

src/main/resources/resources/pathway/runPathwayAnalysis.sql
aklochkova: These names were too long for one of three databases (I don't remember for which exactly, most likely for Oracle).

pavgra:
@aklochkova , how did you test it? I'm getting the exception:
```
2019-01-21 17:01:52.257 ERROR taskExecutor-3 org.springframework.batch.core.step.AbstractStep -  - Encountered an error executing step cohortDefinition.generateCohort in job generateLocalCohort
java.lang.RuntimeException: org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [DELETE FROM five_three_plus_temp.cohort_inclusion WHERE cohort_definition_id = ?;]; nested exception is org.postgresql.util.PSQLException: ERROR: relation "five_three_plus_temp.cohort_inclusion" does not exist
  Position: 13
```

Have you removed all tables from your temp schema before testing?

pavgra:
@chrisknoll , I found the issue due to which the inclusion tables were still required. Addressed it here:
https://github.com/OHDSI/WebAPI/pull/761

pavgra:
@aklochkova , still getting error with Oracle:

```
2019-01-22 11:24:55.173 INFO taskExecutor-9 org.springframework.batch.core.job.SimpleStepHandler -  - Executing step: [generateCohortCharacterization.generate]
2019-01-22 11:24:56.347 ERROR taskExecutor-9 org.springframework.batch.core.step.AbstractStep -  - Encountered an error executing step generateCohortCharacterization.generate in job generateCohortCharacterization
java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.springframework.jdbc.BadSqlGrammarException: ConnectionCallback; bad SQL grammar [BEGIN
  EXECUTE IMMEDIATE 'TRUNCATE TABLE d2g9hs72cov_ref';
  EXECUTE IMMEDIATE 'DROP TABLE d2g9hs72cov_ref';
EXCEPTION
  WHEN OTHERS THEN
    IF SQLCODE != -942 THEN
      RAISE;
    END IF;
END]; nested exception is java.sql.SQLException: ORA-06550: line 9, column 3:
PLS-00103: Encountered the symbol "end-of-file" when expecting one of the following:

   ; <an identifier> <a double-quoted delimited-identifier>
The symbol ";" was substituted for "end-of-file" to continue.
```

src/main/java/org/ohdsi/webapi/util/PreparedStatementRenderer.java
```diff
@@ -271,7 +271,7 @@ public void setValues(PreparedStatement ps) throws SQLException {
 
   public String getSql() {
     String translatedSql = SqlTranslate.translateSql(sql, targetDialect, sessionId, null);
-    return DBMSType.ORACLE.getOhdsiDB().equals(targetDialect) ? translatedSql.replaceAll(";$", "") : translatedSql;
+    return DBMSType.ORACLE.getOhdsiDB().equals(targetDialect) && !translatedSql.startsWith("BEGIN") ? translatedSql.replaceAll(";$", "") : translatedSql;
```
chrisknoll: I would look to finding a sql-render solution here, we shouldn't have dialect specific rules here.  In the case where getSql() was originally stripping out `;` out of the translated sql, I think the correct thing to do here is always splitSql the renderd sql in the case where a single `CREATE TABLE()` statement might be split into 2 distinct statements.

pavgra: > I think the correct thing to do here is always splitSql the renderd sql 

That would require mass changes in WebAPI code and easy to miss until we have comprehensive test coverage

> I would look to finding a sql-render solution here

Not sure how it is possible: first you run sql render -> then you do sql split which strips out `;` -> then you may need to return `;` back in case of a statement in `BEGIN - END` block

All in all, I agree with the @chrisknoll idea in general but not sure how to combine with existing state of art w/o mass refactoring

schuemie: This might not be relevant, but have you guys looked at [the `splitSql` function ](https://github.com/OHDSI/SqlRender/blob/master/java/org/ohdsi/sql/SqlSplit.java#L30) in SqlRender?

chrisknoll: The trouble is that we don't always split sql everywhere in our code (sometimes it's just a single statement like a prepared statement), so this is just an effort to make it more user-friendly by handling it even in the single translateSql case instead of having to split every time.

Maybe there's a way for the translateSql to apply the logic found in splitSql directly?  Our issue is just finding everywhere we call a sql statement and making sure we do a splitSql is pronte to error and a bit more work than we want to take on, and having it handled in translateSql keeps it centralized.

pavgra:
Generation was tested using Oracle and Postgres

Issue #353, opened 2018-03-21
Non-transactional cohort definition creation.


SBylemans:
Hi there,

I'm doing some custom development with regards to WebAPI and Atlas for a project and me and my colleagues stumbled upon an issue when editing the creation of a cohort (`public CohortDefinitionDTO createCohortDefinition(CohortDefinitionDTO def) throws Exception` in `CohortDefinitionService`).

The first thing we noticed was that this method was not transactional while multiple saves to the database are done. This can cause a corruption in the database.

Second, the `put` and `delete` permissions on this new definition are done later as a part of a filter in `AtlasSecurity`. Even when making the creation of a cohort transactional, the creation of those permissions will not be part of that transaction. This can again create a corruption in the database, even in such a way that the creator of the cohort cannot access it anymore.

Seeing as my colleagues and me are new to this community, we are trying to get a feel as to what is the opinion of you guys regarding solving this issue. Would this be an adaptation that is acceptable for you or is there a reason why it is implemented like this? We have seen some other issues that we may want to solve, mostly similar to the one described above.

PR #729, opened 2018-12-28
Made permissions adding transactional

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/353


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -151,7 +153,13 @@ public CcServiceImpl(
     public CohortCharacterizationEntity createCc(final CohortCharacterizationEntity entity) {
         entity.setCreatedBy(getCurrentUser());
         entity.setCreatedDate(new Date());
-        return saveCc(entity);
+        CohortCharacterizationEntity savedEntity = saveCc(entity);
+        try {
+            ((ProcessResponseContentFilter)security.getFilters().get(CREATE_COHORT_CHARACTERIZATION)).doProcessResponseContent(savedEntity.getId().toString());
+        } catch (Exception e) {
+            log.error("Failed to add permissions to cohort characterization with id = " + savedEntity.getId(), e);
```
pavgra: 1. The core issue was in the ability to receive entity w/o permissions due to separate transactions used for the creation of the entity and creation of the permissions (versus a single transaction required). The introduced approach doesn't fix the issue since, because in case of permission creation failure, the entity transaction won't rollback and method will finish execution successfully (the exception is not rethrown).

2. This approach uses an imperative approach and splits the security layer. I would look for something that leaves all security related things in the Shiro related classes and doesn't explicitly expose it to business layer

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
```diff
@@ -151,43 +190,43 @@ protected FilterChainBuilder setupProtectedPaths(FilterChainBuilder filterChainB
...
-      .addProtectedRestPath("/ir/*", "jwtAuthc, authz")
+      .addProtectedRestPath("/ir", JWT_AUTHC, AUTHZ, CREATE_IR)
+      .addProtectedRestPath("/ir/*/copy", CREATE_COPY_IR)
+      .addProtectedRestPath("/ir/*", JWT_AUTHC, AUTHZ)
       .addProtectedRestPath("/ir/*/execute/*")
 
       // comparative cohort analysis (estimation)
-      .addProtectedRestPath("/comparativecohortanalysis", "createPermissionsOnCreatePle")
-      .addProtectedRestPath("/comparativecohortanalysis/*", "deletePermissionsOnDeletePle")
+      .addProtectedRestPath("/comparativecohortanalysis", CREATE_PLE)
```
pavgra: Why do you need the `CREATE_PLE` here if it is already executed from the service-layer method? Seems like either exhaustive workload or even duplicated permissions

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
```diff
@@ -269,44 +308,43 @@ protected FilterChainBuilder setupProtectedPaths(FilterChainBuilder filterChainB
...
-    return filters;
+  public Map<FilterTemplates, Filter> getFilters(){
+    return new HashMap<>(filters);
+  }
+  
+  private void fillFilters() {
+    filters.put(NO_SESSION_CREATION, new NoSessionCreationFilter());
+    filters.put(FORCE_SESSION_CREATION, new ForceSessionCreationFilter());
+    filters.put(AUTHZ, new UrlBasedAuthorizingFilter());
+    filters.put(CREATE_COHORT_DEFINITION, this.getCreatePermissionsOnCreateCohortDefinitionFilter());
```
pavgra: The same question as above - why would you need to manage the `filters` map if it is not required and the methods are actually used in service-layer

src/main/java/org/ohdsi/webapi/events/DeleteCcEvent.java
```diff
@@ -0,0 +1,16 @@
+package org.ohdsi.webapi.events;
+
+import org.springframework.context.ApplicationEvent;
+
+public class DeleteCcEvent extends ApplicationEvent {
+    private final Integer id;
```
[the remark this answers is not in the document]
aklochkova: fixed

wivern:
1. Looks like delete methods is out of transaction, since directly referenced from the same class by event handlers
2. Too much event handlers used simply as markers
Not sure but probably it's better to:
1. Make a single event class with Entity name inside and
2. Make an event handler that delegate deletion requests to appropriate service
and this would eliminate the issues pointed above and reduce number of classes.
How about it?

src/main/java/org/ohdsi/webapi/shiro/filters/ProcessResponseContentFilterImpl.java
```diff
@@ -0,0 +1,65 @@
...
+            authorizer.addPermissionsFromTemplate(currentUserPersonalRole, template, id);
+        } catch (Exception ex) {
+            eventPublisher.publishEvent(getDeletionEvent(this, entityName, Integer.parseInt(id)));
+            log.error("Failed to add permissions to " + entityName.getName() + " with id = " + id, ex);
+        }
+    }
+
+    @Override
+    protected boolean shouldProcess(ServletRequest request, ServletResponse response) {
+        if (shouldHttpMethodCheck) {
```
pavgra: Why all these complex comparisons and actions tightly coupled to app business logic (`StringUtils.endsWithIgnoreCase(path, "copy")`) are required? Just take `httpMethod` as a constructor parameter and leave only following in the method:

```
return httpMethod.equalsIgnoreCase(WebUtils.toHttp(request).getMethod());
```

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurityHelper.java
```diff
@@ -0,0 +1,173 @@
...
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_FEATURE_ANALYSIS;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_PATHWAY_ANALYSIS;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_PLE;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_PLP;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_PREDICTION;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.DELETE_SOURCE;
+import static org.ohdsi.webapi.shiro.management.FilterTemplates.JWT_AUTHC;
+
+@Component
+public class AtlasSecurityHelper {
```
pavgra: Why have you moved it to separate class?

src/main/java/org/ohdsi/webapi/shiro/management/ResponseContentFilterBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.ohdsi.webapi.events.EntityName;
+import org.ohdsi.webapi.shiro.PermissionManager;
+import org.ohdsi.webapi.shiro.filters.ProcessResponseContentFilterImpl;
+import org.springframework.context.ApplicationEventPublisher;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ResponseContentFilterBuilder {
```
pavgra: With your current usage approach, doesn't seem that there is a big need in the builder

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
```diff
@@ -143,170 +200,60 @@ public AtlasSecurity() {
...
+    addProcessEntityFilter(CREATE_ESTIMATION, estimationPermissionTemplates);    
+  }
+
+  
+  
+  private void addProcessEntityFilter(FilterTemplates template, Map<String, String> permissionTemplates){
+    filters.put(template, helper.createResponseContentBuilder(template, permissionTemplates, authorizer, eventPublisher).build());
+  }
+
+  private void addProcessEntityFilterWithCheck(FilterTemplates template, Map<String, String> permissionTemplates){
```
pavgra: Why do you need the 2 methods, add the `httpMethod` param into the filter and pass it through

src/main/java/org/ohdsi/webapi/shiro/filters/ProcessResponseContentFilterImpl.java
wivern: Moreover, in this case `shouldHttpMethodCheck` becoming redundant

Issue #724, opened 2018-12-17
Failed to import cohort characterization design


chrisknoll:
When importing a CharacterizationDesign from a 2.6 instance into latest master, the import failed:

```
com.fasterxml.jackson.databind.JsonMappingException: Unexpected token (END_OBJECT), expected FIELD_NAME: missing property 'criteriaType' that is to contain type id  (for class org.ohdsi.webapi.feanalysis.dto.BaseFeAnalysisCriteriaDTO)
 at [Source: N/A; line: -1, column: -1] (through reference chain: org.ohdsi.webapi.cohortcharacterization.dto.CcExportDTO["featureAnalyses"]->java.util.ArrayList[6])

```

We should have compatibility between 2.6 and the next release with respect to importing. Anything that is newly added should have a default value coming from 2.6.

aklochkova:
@chrisknoll Could you please provide this Characterization? I can't reproduce.

chrisknoll:
[CC_with_CriteriaFeature.txt](https://github.com/OHDSI/WebAPI/files/2737647/CC_with_CriteriaFeature.txt)

The attached export was created in the 2.6 release version of webapi.  When you attempt to import this design into an instance running from latest master in WebAPI, you get this error in the console:
![image](https://user-images.githubusercontent.com/6818777/50845412-ca9f9e00-133a-11e9-9c0f-a352767e0b01.png)

![image](https://user-images.githubusercontent.com/6818777/50845420-d12e1580-133a-11e9-8caf-57781cb2ed1f.png)

PR #741, opened 2019-01-15
Set 'Prevalence' stat_type as default for old cohort characterizations

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/724


src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -290,7 +290,7 @@ private boolean ccHasPresetDistributionAnalyses() {
             String groupQuery = getCriteriaGroupQuery(analysis, feature);
             String[] paramNames = CRITERIA_PARAM_NAMES.toArray(new String[0]);
 
-            if (CcResultType.PREVALENCE.equals(analysis.getStatType())) {
+            if (CcResultType.PREVALENCE.equals(analysis.getStatType()) || analysis.getStatType() == null) {
```
pavgra: 1. This line is repeated 5 times - why not to move it to util method?
1. Why not migrate all empty fields to prevalence and that's it? instead of adding logic to multiple places

aklochkova: 2. I thought it might be useful to distinguish old CC with null statType from new

src/main/resources/db/migration/oracle/V2.7.0.20190116183005__default_stat_type.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE WEBAPI_EMPTY_2.FE_ANALYSIS
```
pavgra: What is `WEBAPI_EMPTY_2`?

chrisknoll: Might be a schema name in a test env. We should be using ${ohdsiSchema} for schema placholders.

Issue #712, opened 2018-11-29
hibernate deprecation


fdefalco:
Currently seeing a few dozen deprecation warnings.  Should we address as part of 2.7 or push to 2.8?

WARN main org.hibernate.orm.deprecation -  - HHH90000014: Found use of deprecated [org.hibernate.id.SequenceHiLoGenerator] sequence-based id generator; use org.hibernate.id.enhanced.SequenceStyleGenerator instead.  See Hibernate Domain Model Mapping Guide for details.

PR #742, opened 2019-01-15
Replaced SequenceGenerator with GenericGenerator

mpozhidaeva:
Fixed https://github.com/OHDSI/WebAPI/issues/712

### DIFF
--- a/src/main/java/org/ohdsi/webapi/executionengine/entity/AnalysisFile.java
+++ b/src/main/java/org/ohdsi/webapi/executionengine/entity/AnalysisFile.java
@@ -16,11 +16,11 @@ public class AnalysisFile {
 
     @Id
     @GenericGenerator(
-        name = "ANALYSIS_INPUT_FILE_SEQUENCE_GENERATOR",
+        name = "analysis_input_file_generator",
         strategy = "org.hibernate.id.enhanced.SequenceStyleGenerator",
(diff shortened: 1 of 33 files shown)


src/main/java/org/ohdsi/webapi/executionengine/entity/AnalysisFile.java
```diff
@@ -8,12 +8,19 @@
...
+import org.hibernate.annotations.GenericGenerator;
+import org.hibernate.annotations.Parameter;
 
 @Entity(name = "input_files")
 public class AnalysisFile {
 
     @Id
-    @GeneratedValue
+    @GenericGenerator(
+        name = "ANALYSIS_INPUT_FILE_SEQUENCE_GENERATOR",
```
pavgra: Why the naming is not consistent? In most cases you have lower-cased names while somewhere the names are typed in upper-case

mpozhidaeva: I just use the names from the current code (with lower and upper cased names) - do you suggest change all names to lower-case format, right?

mpozhidaeva: Ok, thanks - updated

PR #745, opened 2019-01-17
Pre-fixed admin permission checking..

### LINKED ISSUE #743: Method for source priority update is not secured
While the application is started in secured mode, I can POST to the endpoint w/o providing an auth token and having admin permissions:

image

NEONKID:
Fixed #743 
I think it seems that the source endpoint subdivision is not being filtered in the filtering endpoint.
I would like you to confirm this part. thank you.


pavgra:
@NEONKID , since we have granular permission checks the following approach was more appropriate:
https://github.com/OHDSI/WebAPI/pull/764

Issue #706, opened 2018-11-26
Error in miredot plugin during compile


chrisknoll:
Receiving the following error during maven build:

```
[ERROR] Critical error: MireDot was unable to process org.ohdsi.webapi.cohortcharacterization.CcController. Interface skipped.
```

Steps to Reproduce:
1. Perform maven build
2. Look in output for the mire dot plugin to fire, and report errors with the CcController interface.

PR #754, opened 2019-01-21
Fixed miredot config

mpozhidaeva:
Fixed https://github.com/OHDSI/WebAPI/issues/706


chrisknoll:
Related to #1507 , it looks like our license has expired again.

@mpozhidaeva , can you provide instructions on how you acquired an updated key?  We can put that into a process for the future.

Issue #694, opened 2018-11-14
Generated Results DDL is incorrect for Impala


konstjar:
There are few issues with generated DDL for Impala
http://<host>/WebAPI/ddl/results?schema=cdm_results&dialect=impala

1. There is no IF EXISTS statement for 'CREATE TABLE @results_database_schema.concept_hierarchy' or 'CREATE TABLE cdm_results.cc_results' or 'CREATE TABLE cdm_results.pathway_analysis_stats'
2. There is no translation for @results_database_schema and @vocab_database_schema in few places
3. 'CREATE TABLE jan11s5ktemp_period ' - This part is causing issues in some environments. This code assumes we have permissions to write in default database, while in most enterprise environments it's prohibited. Suggestion - add @results_database_schema to create temporary table in results schema.
4. "mode_id INT DEFAULT 0" - Impala does not allow DEFAULT statement
5. Met few CAST issues for "Populate the hierarchy lookup table per treemap"

chrisknoll:
The issue with `@results_database_schema` is going to be addressed in #692.

The issue with @vocab_database_schema is that the interface to the results DDL schema does not accept a parameter for the vocabulary schema...I think the assumption there is that the caller needs to do their own find-replace on that token when using the script (or we add a new method to the results schema that lets you pass in the additional parameter.

For the 'permanent temp tables' there should be an 'oracle_temp_schema' param in sqlrender which will prefix the temp schema to temp tables.  But where do we read this value from?  Does it become another parameter?

The lack of default values in impala is disappointing.  I think this brings up, again, the notion of 'minimum required functionality' to be supported on a CDM.

For the CAST issues, could you contribute commits to the pdw-ddl branch (which #692 is based on) and I can check the cast fixes against PDW?

mpozhidaeva:
1. fix for **concept_hierarchy** here - https://github.com/OHDSI/WebAPI/pull/692/files#diff-5b86cac7b3e2e0683182dc6543db2b74
2. https://github.com/OHDSI/WebAPI/pull/723/files
3. 
4. https://github.com/OHDSI/SqlRender/pull/144
5.

mpozhidaeva:
I would like to remind I need https://github.com/OHDSI/WebAPI/pull/692/files#diff-5b86cac7b3e2e0683182dc6543db2b74 for the current issue.

PR #762, opened 2019-01-23
Fixed result DDL for Impala

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/694

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/DDLService.java
+++ b/src/main/java/org/ohdsi/webapi/service/DDLService.java
@@ -42,6 +42,7 @@ public class DDLService {
 	public static final String VOCAB_SCHEMA = "vocab_schema";
 	public static final String RESULTS_SCHEMA = "results_schema";
 	public static final String CEM_SCHEMA = "cem_results_schema";
+	public static final String TEMP_SCHEMA = "oracle_temp_schema";
 
 	private static final Collection<String> RESULT_DDL_FILE_PATHS = Arrays.asList(
 		"/ddl/results/cohort.sql",
@@ -113,6 +114,7 @@ public class DDLService {
 		Map<String, String> params = new HashMap<String, String>() {{
 			put(VOCAB_SCHEMA, vocabSchema);
 			put(RESULTS_SCHEMA, resultSchema);
+			put(TEMP_SCHEMA, resultSchema);
 		}};
 
 		return generateSQL(dialect, params, resultDDLFilePaths, RESULT_INIT_FILE_PATHS, RESULT_INDEX_FILE_PATHS);
@@ -158,7 +160,7 @@ public class DDLService {
 		statement.targetDialect = dialect.toLowerCase();
 		statement.sql = sql;
 		statement.parameters = new HashMap<>(params);
-		TranslatedStatement translatedStatement = translateSQL(statement);
+		TranslatedStatement translatedStatement = translateSQL(statement, params.get(TEMP_SCHEMA));
 		return translatedStatement.targetSQL;
 	}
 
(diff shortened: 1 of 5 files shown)


src/main/resources/ddl/results/init_concept_hierarchy.sql
```diff
@@ -6,12 +6,12 @@ INSERT INTO @results_schema.concept_hierarchy
 	(concept_id, concept_name, treemap, level1_concept_name, level2_concept_name, level3_concept_name, level4_concept_name)
 SELECT
 	snomed.concept_id,
-	snomed.concept_name AS snomed_concept_name,
-	'Condition'         AS treemap,
-	pt_to_hlt.pt_concept_name,
-	hlt_to_hlgt.hlt_concept_name,
-	hlgt_to_soc.hlgt_concept_name,
-	soc.concept_name    AS soc_concept_name
+	CAST(snomed.concept_name AS VARCHAR(400))  AS snomed_concept_name,
```
mpozhidaeva: Formally, I need CAST only for treemap field - but @konstjar met with STRING type for concept.concept_name.  But here https://github.com/OHDSI/CommonDataModel/blob/v5.3.0/Impala/OMOP%20CDM%20impala%20ddl.txt type is ok.

src/main/java/org/ohdsi/webapi/service/DDLService.java
```diff
@@ -113,6 +114,7 @@ public String generateResultSQL(
 		Map<String, String> params = new HashMap<String, String>() {{
 			put(VOCAB_SCHEMA, vocabSchema);
 			put(RESULTS_SCHEMA, resultSchema);
+			put(TEMP_SCHEMA, resultSchema);
```
pavgra: I would introduce optional `tempSchema` query param and use `ObjectUtils.firstNonNull`

src/main/java/org/ohdsi/webapi/service/DDLService.java
```diff
@@ -158,7 +160,7 @@ private String translateSqlFile(String sql, String dialect, Map<String, String>
 		statement.targetDialect = dialect.toLowerCase();
 		statement.sql = sql;
 		statement.parameters = new HashMap<>(params);
-		TranslatedStatement translatedStatement = translateSQL(statement);
+		TranslatedStatement translatedStatement = translateSQL(statement, params.get(TEMP_SCHEMA));
```
pavgra: Why not extend `SourceStatement` with `oracleTempSchema`?

src/main/java/org/ohdsi/webapi/service/DDLService.java
```diff
@@ -156,6 +160,7 @@ private String translateSqlFile(String sql, String dialect, Map<String, String>
 
 		SourceStatement statement = new SourceStatement();
 		statement.targetDialect = dialect.toLowerCase();
+		statement.oracleTempSchema = params.get(TEMP_SCHEMA) != null ? params.get(TEMP_SCHEMA).toLowerCase() : null;
```
pavgra: Why do you need `toLowerCase()`?

src/main/java/org/ohdsi/webapi/service/SqlRenderService.java
```diff
@@ -51,7 +51,7 @@ static TranslatedStatement translateSQL(SourceStatement sourceStatement) {
             if ((sourceStatement.targetDialect == null) || ("sql server".equals(sourceStatement.targetDialect))) {
                 translated.targetSQL = renderedSQL;
             } else {
-                translated.targetSQL = SqlTranslate.translateSql(renderedSQL, sourceStatement.targetDialect);
+                translated.targetSQL = SqlTranslate.translateSql(renderedSQL, sourceStatement.targetDialect, null, sourceStatement.oracleTempSchema);
```
pavgra: I believe, we need `null` -> `SessionUtils.sessionid()`

Issue #8, opened 2015-02-04
Error from spring framework when searching for concept that does not exist.


chrisknoll:
Receiving the following error when going to this WebAPI service:
http://localhost:8084/WebAPI/vocabulary/concept/-1

javax.servlet.ServletException: org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0

Expected Result:
No error should be returned, and an empty object should return, or a HTTP 404 returned?  Not sure how the restAPI should report the not found...can it return null?

alexfranken:
Currently, a 404 is returned and the exception is logged as ERROR.  With that said, I can add code to catch the exception or query for a list and get the first.  The queryForObject method that is used expects a single result.

http://localhost:8080/WebAPI:> get vocabulary/concept/-1
> GET http://localhost:8080/WebAPI/vocabulary/concept/-1

< 404 NOT_FOUND
< Server: Apache-Coyote/1.1
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Headers: origin, content-type, accept, authorization
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Methods: GET, POST
< Access-Control-Max-Age: 1209600
< Content-Length: 0
< Date: Wed, 04 Feb 2015 16:38:08 GMT

leeevans:
This issue would apply to any WebAPI call that now uses getJdbcTemplate()
I believe this is a change in behavior from the previous release of WebAPI, before the change to spring , in the way no data found is handled.

This might be a useful background reference:

http://stackoverflow.com/questions/10606229/jdbctemplate-query-for-string-emptyresultdataaccessexception-incorrect-result

alexfranken:
I pushed two commits that you can review.  I changed /vocabulary/concept/{id} and /cohortdefinition/{id} endpoints.  Let me know if you any questions about the commits.  I logged a debug level message for requests that result in 0 results, and continued to return 404.

alexfranken:
Sorry Chris, I just reread your initial comment.  If returning null, a 204 NO CONTENT is returned.  This may be appropriate to signal a correct endpoint with no data vs an incorrect endpoint. Currently an invalid endpoint returns 404 and these endpoints where no resource by id is found.

alexfranken:
I think I prefer the 404 as the resource at the uri is not found.

leeevans:
[...]

PR #767, opened 2019-01-24
Changed response code for getting of missing concepts

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/8


pavgra:
I would expand the 404 code for all cases with not found entities (all entities) and would create a standardized UI handler for this.

pavgra:
1. 500 instead of 404:

image

2. Conflicts

mpozhidaeva:
@pavgra  - what Atlas version do you use? I haven't find the usage of the first rest method from your screenshot. Will add the check.
Also it's better to recheck IR after https://github.com/OHDSI/WebAPI/pull/794

Issue #453, opened 2018-05-30
Issue w/ filters in Visit during cohort period report


pavgra:
When following filters are selected:
- Visit = all
- Visit type = Visit derived from encounter on medical professional claim paid

No data is returned (zero counts).

But if you pick:
- Visit = different from "all"
- Visit type = Visit derived from encounter on medical professional claim paid

There are counts returned

pavgra:
With "All" visits:
![image](https://user-images.githubusercontent.com/10010071/48368477-d1cf6600-e681-11e8-85d1-5d9140a01bde.png)

With specific visit:
![image](https://user-images.githubusercontent.com/10010071/48368498-e4499f80-e681-11e8-9aaa-7a13777bc116.png)

PR #772, opened 2019-01-25
Added condition for correct Select

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/453


src/main/java/org/ohdsi/webapi/cohortresults/CohortResultsAnalysisRunner.java
```diff
@@ -1632,7 +1632,8 @@ public HealthcareVisitUtilizationReport getHealthcareVisitReport(JdbcTemplate jd
 		String reportPath = BASE_SQL_PATH + "/healthcareutilization/getVisitUtilization.sql";
 		String reportSql = ResourceHelper.GetResourceAsString(reportPath);
 		
-		String summarySql = SqlRender.renderSql(reportSql, new String[] {"is_summary"}, new String[]{"TRUE"});
+		String summarySql = SqlRender.renderSql(reportSql, new String[] {"is_summary", "visit_concept_id"}, new String[]{"TRUE", 
+				visitConceptId == null ? "" : visitConceptId.toString()});
```
pavgra: Why the same approach is not applied to `visit_type_concept_id`?

PR #773, opened 2019-01-25
removes DEFAULT column value for Impala

### LINKED ISSUE #694: Generated Results DDL is incorrect for Impala

wivern:
closes #694 for `impala,"DEFAULT 0",""`

### DIFF
--- a/src/main/java/org/ohdsi/webapi/Constants.java
+++ b/src/main/java/org/ohdsi/webapi/Constants.java
@@ -35,5 +35,6 @@ public interface Constants {
     String LDAP_PROVIDER = "provider";
     String ROLE_GROUP_MAPPING = "roleGroupMapping";
     String PRESERVE_ROLES = "preserveRoles";
+    String DB_SUPPORTS_DEFAULT = "SupportsDefault";
   }
 }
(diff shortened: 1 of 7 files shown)


anthonysena:
From our discussion on the Atlas/WebAPI WG: we've proposed to drop all DEFAULT constraints for tables in the results schema since not all RDMBS' have this type of constraint. Therefore, the code that will perform CRUD operations against the results schema tables will need to set the appropriate values instead of leaving it to the RDBMS default.

Issue #768, opened 2019-01-24
Cohort Characterization - Feature Analysis fails to load


anthonysena:
When Atlas attempts to load the list of features analyses under cohort characterization, this REST endpoint throws an error: `WebAPI/feature-analysis?size=100000`. 

Here is the error (minus the stack trace returned from the service:
`
"identifier": 1009,
"entityName": "org.ohdsi.webapi.feanalysis.domain.FeAnalysisCriteriaEntity",
"localizedMessage": "Object [id=1009] was not of the specified subclass [org.ohdsi.webapi.feanalysis.domain.FeAnalysisCriteriaEntity] : Discriminator: null",
"message": "Object [id=1009] was not of the specified subclass [org.ohdsi.webapi.feanalysis.domain.FeAnalysisCriteriaEntity] : Discriminator: null",
"suppressed": []
`

As a note, the `cohort_features` table in the environment where we encountered the error is empty. Please let me know if you need any further details.

anthonysena:
Combing the logs on this issue, reveals that this hibernate query fails:

`select design0_.fe_analysis_id as fe_analy5_29_0_, design0_.id as id2_29_0_, design0_.id as id2_29_1_, design0_.expression as expressi3_29_1_, design0_.fe_analysis_id as fe_analy5_29_1_, design0_.name as name4_29_1_, design0_.criteria_type as criteria1_29_1_ from dbo.fe_analysis_criteria design0_ where design0_.fe_analysis_id=?
`

Looking at the table `fe_analyis_criteria` with ID = 1009 reveals that it contains a `NULL `for `criteria_type`:

![image](https://user-images.githubusercontent.com/12902366/51694135-21110b80-1fce-11e9-9ee9-30e4b1b5a67b.png)

Hoping this additional information is helpful.

pavgra:
@anthonysena , I believe the issue occurred due to migrations appearing in wrong order. Now we have:

```
V2.6.0.20181107165252__fe_criteria_type.sql
V2.7.0.20190116183005__default_stat_type.sql
```

while actually, the first one relies on the second one (requires it to be executed before). So, in short, I believe that applying the first migration again, after the second is done, should fix the issue. I'm preparing the fix now and would be grateful for your testing.

chrisknoll:
[...]

PR #776, opened 2019-01-25
Fixes NULL values in Feature Analyses Criteria

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/768

### DIFF
--- a/src/main/resources/db/migration/oracle/V2.7.0.20190125141500__fe-analysis-criteria_stat-type-fix.sql
+++ b/src/main/resources/db/migration/oracle/V2.7.0.20190125141500__fe-analysis-criteria_stat-type-fix.sql
@@ -1,0 +1,18 @@
+ALTER TABLE ${ohdsiSchema}.fe_analysis MODIFY (stat_type NOT NULL);
+
+UPDATE fe_analysis_criteria
+  SET criteria_type =
+    CASE WHEN criteria_type IS NULL THEN
+      CASE
+        WHEN fa.stat_type = 'PREVALENCE' THEN 'CRITERIA_GROUP'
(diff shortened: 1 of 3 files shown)


src/main/resources/db/migration/oracle/V2.7.0.20190125141500__fe-analysis-criteria_stat-type-fix.sql
```diff
@@ -0,0 +1,18 @@
+ALTER TABLE ${ohdsiSchema}.fe_analysis MODIFY (stat_type NOT NULL);
```
chrisknoll: will this work if there is existing data in the db?  Should the UPDATE apply first and then the column MODIFY?

anthonysena: [V2.7.0.20190116183005__default_stat_type.sql](https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/oracle/V2.7.0.20190116183005__default_stat_type.sql) takes care of this since it will be applied prior to this change:

```
ALTER TABLE ${ohdsiSchema}.fe_analysis
  MODIFY stat_type DEFAULT 'PREVALENCE';

UPDATE ${ohdsiSchema}.fe_analysis
SET stat_type = 'PREVALENCE'
WHERE stat_type IS NULL;
```

chrisknoll: Thanks for the clarification @anthonysena , i'm blind.

chrisknoll:
No objections, thanks to you both.

PR #769, opened 2019-01-25
Added Impala/Oracle temporary tables cleanup after failed Cohort/IR job

wivern:
Fixes #576

PR #781, opened 2019-01-29
Added isolated cohort execution for IR analyses

pavgra:
fixes https://github.com/OHDSI/Atlas/issues/1098


chrisknoll:
Spoke with @pavgra , and we've decided to do this PR, merge to master, and then update #769 with these latest changes.  Then #769 will be checked that the temp tables for IR isolation are properly cleaned up.

PR #786, opened 2019-01-30
Added BigQuery support

pavgra:
Notes:
- Table names were changed to lower case since BigQuery is case-sensitive in terms of table references and official OMOP DDL for BigQuery uses lower case


chrisknoll:
@pavgra , let's handle JDBC drivers the same way we handle it for impala:  there's a property in pom.xml that you can override via settings.xml that can point to the drivers in your local env.  This keeps large binaries like JDBC drivers out of the repository.  Example here:
https://github.com/OHDSI/WebAPI/blob/master/pom.xml#L835

pavgra:
@chrisknoll , ok, since the property is already in place, I've just removed JARs from the project

src/main/resources/resources/cohortresults/sql/cohortSpecific/byConcept/drugOccursRelativeToIndex.sql
```diff
@@ -60,7 +60,7 @@ FROM   (SELECT cohort_definition_id,
                           AND cohort_definition_id = @cohortDefinitionId 
                    GROUP  BY cohort_definition_id, 
                              Cast(stratum_1 AS INTEGER) 
-                   HAVING Sum(count_value) > @minCovariatePersonCount) ct1 
+                   HAVING Sum(heracles_results.count_value) > @minCovariatePersonCount) ct1
```
chrisknoll: this syntax is a bit mysterious to me:  
`Having {aggregate function > value} ct1`

What's the purpose of aliasing the having as `ct1`?

chrisknoll: Thanks @pavgra, i misread the opening/closing parens.

anthonysena:
From discussion: this needs full testing against all Atlas functionality

chrisknoll:
@anthonysena : I beleive there is a settings in pom.xml for big query drivers that is left to some defaullt, which you need to specifyi in your local settings.xml file to point to a local directory that has the drivers. The idea here is that this keeps the JDBC drivers out of the repo, and I woudln't recommend putting them under /src/main (assuming that's the root of the git repo) instead just make a completely different folder for those.

anthonysena:
Thanks @chrisknoll - think I figured it out. The dependency list under the `webapi-bigquery` profile needed some adjustment. The groupId field should be com.google.* I'm making the adjustment locally and will push an update to this branch assuming it all works. 

To your point about the JDBC driver: I put those under the $\src\main\extras\bigquery manually but maven still attempts to put the files that are in the profile there since they are listed as dependencies in the pom for that profile. I believe that maven will attempt to download those dependencies required by the JDBC driver but not the JDBC driver itself (you have to configure that location like you said via the settings.xml).

anthonysena:
@pavgra - With the changes in https://github.com/OHDSI/WebAPI/pull/786/commits/cc1abae4c7dbbc106934f068b623eab2d0e993cf, I'm now able to run WebAPI and connect to the Google BigQuery DB properly. I'll test this tomorrow but if you like, it would be a good exercise to pull down the change I made just to make sure it doesn't break your environment.

anthonysena:
Also, making a note of how I was able to set up NetBeans to test these changes using a multiple profile configuration since it may be of interest to others.

### Download the Google BigQuery Drivers

Go to https://cloud.google.com/bigquery/partners/simba-drivers/, download the JDBC 4.2 drivers. Unzip them to a location such as `C:\bigquery`.

### NetBeans Project Properties Configuration

Go to File -> Project Properties -> Actions. Select "Debug Project" and set it as follows:

image

From the Actions list, select "Debug project" and set the following:

- **Execute Goals:** org.springframework.boot:spring-boot-maven-plugin:1.2.1.RELEASE:run
- **Activate Profiles:** webapi-bigquery
- **Set Properties:** jpda.listen=maven

What this will do is add the `webapi-bigquery` profile to your current profile (in my case my profile is `webapi-local`) so that when you start up debugging, you'll see the new profile added in the maven command in the NetBeans output window (note the **_bold italic_** maven -P switch):

Starting

> JPDA Listening Start...
> JPDA Address: SENA:58956
> Port:58956

> cd C:\Git\OHDSI\WebAPI; "MAVEN_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,server=n,address=58956" "JAVA_HOME=C:\\Program Files\\Java\\jdk1.8.0_171" cmd /c "\"\"C:\\Program Files\\NetBeans 8.2\\java\\maven\\bin\\mvn.bat\" -Djpda.listen=maven -Djpda.address=58956 -Dmaven.ext.class.path=\"C:\\Program Files\\NetBeans 8.2\\java\\maven-nblib\\netbeans-eventspy.jar\" -Dfile.encoding=UTF-8 _ **-Pwebapi-bigquery,webapi-local**_ org.springframework.boot:spring-boot-maven-plugin:1.2.1.RELEASE:run\""

> Running NetBeans Compile On Save execution. Phase execution is skipped and output directories of dependency projects (with Compile on Save turned on) will be used instead of their jar artifacts.
> Scanning for projects...


### Configuring the maven settings.xml

Now, override the pom.xml `bigquery.classpath` in your settings.xml file by adding the following profile:

```
<profile>
	<id>webapi-bigquery</id>
	<properties>
		<bigquery.classpath>C://bigquery</bigquery.classpath>
	</properties>
</profile>
```

This will start up WebAPI with your normal profile (`webapi-local`) and add in the `webapi-bigquery` profile and allow NetBeans to load up the appropriate Google BigQuery drivers.

pavgra:
@anthonysena , thank you for the research and fixes! I haven't checked the updated PR myself but from the code perspective - why the following block was removed?

```
              <execution>	
                <id>GoogleBigQueryJDBC42</id>	
                <phase>initialize</phase>	
                <goals>	
                  <goal>install-file</goal>	
                </goals>	
                <configuration>	
                  <groupId>GoogleBigQueryJDBC42</groupId>	
                  <artifactId>GoogleBigQueryJDBC42</artifactId>	
                  <version>1.24.1</version>	
                  <packaging>jar</packaging>	
                  <file>${bigquery.classpath}/GoogleBigQueryJDBC42.jar</file>	
                </configuration>	
              </execution>
```

anthonysena:
@pavgra - I took out that block because it seemed extraneous since maven is not managing the JDBC driver for BigQuery. In thinking about your question, I took another look at the dependency list for BigQuery: 

- google-api-client
- google-api-services-bigquery
- google-cloud-bigquery
- google-http-client
- google-http-client-jackson2
- google-oauth-client

The corresponding execution section references attempts to install the plugins for the maven dependencies above with the exception of the artifact: google-cloud-bigquery. I think that this might also be extraneous. Let me take another look at this since I will likely have one more small change for discussion here.

anthonysena:
@pavgra - here are the changes I made after testing: https://github.com/OHDSI/WebAPI/pull/786/commits/1d04be2ce5f6a5a0c067c7ead62da6c0b9f76ac1. 

I double checked the pom.xml and the `google-cloud-bigquery` artifact is required for the JDBC driver to load properly. It does not require an execution step to install since it is likely picking up the JDBC driver from the class path.

As discussed via Skype, I'd like to do another round of testing this branch against our internal CDMs since there were enough changes to the core PreparedSql classes that I'd like to just make sure we don't have any regressions. Could you review my changes and if they look good merge in master to this branch and I'll do one more round of testing?

Thanks!

anthonysena:
Things looked good after another round of testing - everything passed except for one. I ran the 'quick analysis' cohort report on a cohort using an internal PDW and then on BigQuery and the results lead me to believe that there is an issue with grouping by age. Here's the tornado plot with PDW:

image

Here it is with BigQuery:

image

Looking at the WebAPI response, it looks like the BigQuery stratum are done per age vs 5 year ranges.

pavgra:
@anthonysena , good catch! Should be fixed by the commit above

Issue #753, opened 2019-01-21
HIBERNATE_SEQUENCE is not exists in sql server


NEONKID:
In Microsoft SQL Server, there is an error that SEQUENCE does not exist while running Reporting function in ATLAS Cohort Definition.

I do not know the minimum version of Microsoft SQL Server that ATLAS supports. To use SEQUENCE, must be at least version 2012. If this is the minimum version, it would be better to add a query to the WebAPI that generates a SEQUENCE for SQL Server.

fdefalco:
I'm seeing this cause errors on save of new Sources through when security is enabled via Atlas.

2019-01-21 15:51:10.397 DEBUG https-jsse-nio-8443-exec-3 org.hibernate.SQL -  - select next value for dbo.hibernate_sequence
2019-01-21 15:51:10.397 DEBUG https-jsse-nio-8443-exec-3 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - could not extract ResultSet [n/a]
com.microsoft.sqlserver.jdbc.SQLServerException: Invalid object name 'dbo.hibernate_sequence'.
...
at org.ohdsi.webapi.service.SourceService.createSource(SourceService.java:218)

fdefalco:
When this is resolved there is an additional error:

2019-01-21 16:32:37.000 WARN https-jsse-nio-8443-exec-8 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - SQL Error: 544, SQLState: S0001
2019-01-21 16:32:37.000 ERROR https-jsse-nio-8443-exec-8 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - Cannot insert explicit value for identity column in table 'source' when IDENTITY_INSERT is set to OFF.

Looks like the Source persistence needs some attention.

chrisknoll:
@NEONKID ,
  In the original implementation of Hibernate, we were using a JPA annotation of `AUTO` on identities, which would use a sequence for Oracle/Postgresql, but use a table-based sequence for MSSQL. You can see the difference bettween platforms in these scripts:

SQL: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/sqlserver/V1.0.0.2__schema-create_jpa.sql

Postgresql: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/postgresql/V1.0.0.2__schema-create_jpa.sql

[...]

PR #791, opened 2019-01-31
Issue 753 hibernate seq for mssql

ssuvorov-fls:
fixes https://github.com/OHDSI/WebAPI/issues/753


chrisknoll:
Ok, so that I'm clear here: this PR is going to remove all identity columns from MSSQL tables.  Then each identity that is removed, a corresponding sequence is created?

I'm looking through your script, and you are including more tables than are necessary. There are some tables which don't have identity columns (such as those tables which we created to represent tables in the results schema, which those need to be dropped at some point. examples: ir_analysis_dist, ir_analysis_result, cohort_summary_stats, etc).

In other cases, there are tables that are part of Spring batch which I do not believe we can modify to remove identity columns for those:  BATCH_JOB_EXECUTION_SEQ, BATCH_STEP_EXECUTION_SEQ.

I think the issue you are trying to address is where we have auto number (identity) columns set up on **_Entity tables_**, and I agree we can migrate those over to use sequences. But for those cases, we need 
to replace the column as you describe but also introduce a new sequence to host the identity value, and change the JPA annotations on the entities to use the sequence.

Finally, if you're making a stored procedure for a temporary migration step, at the end, i think that procedure should be dropped. we shouldn't leave obsolete procedures around.

So let's take a more focused approach to adjusting those tables that **_only_** need to be changed, introduce the new sequences, make sure that we adjust the JPA notations to use the new sequences. I do not believe this is a widespread probelm, I think only a few entities (like Source) are effected, since we recently had an update to 'reset' all the sequences, most entities should have sequences for them already. Source was just one that was missed, I think.

Some background: in the original Hibernate implementation, the `@generatedId` annotation was set to AUTO, which meant that for MSSQL, it would use identity, and Postgresql and Oracle would use a default sequence 'HIBERNATE_SEQUENCE' seen here:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/postgresql/V1.0.0.2__schema-create_jpa.sql

In the end, tho it was annoying that MSSQL had identity columns and the other platforms had sequences, and since MSSQL 2012 had sequences (origionally we supported MSSQL 2008!) we moved all autonumber fields to sequences which you can see here:
Postgresql: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/postgresql/V2.2.5.20180215105415__separate-sequences.sql
SqlServer: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/sqlserver/V2.2.5.20180215105415__separate-sequences.sql

We did this without any stored procedures, maybe what you should do here is copy parts of the sqlserver scripts to alter the tables and add the sequences?  Up to you, but we already have a decent model on how to perform this maneuver.

Also, please make sure that this works consistently across Oracle, Postgresql and SqlServer.  I don't think we can just adjust the schema on MSSQL (especially where sequence creation is needed) without also considering Oracle and Postgresql.

Thanks!

-Chris

chrisknoll:
I believe this will be superseded by #795.  However, if we want to also address eliminating the HIBERNATE_SEQENCE from mssql, we can make that part of #795 as well, tho the focus of #795 is to addrss the Source and SourceDaimon entities specifically.

ssuvorov-fls:
There're 3 additional tables - analysis_execution, cca_execution and heracles_visualization_data. Should I modify script to fix only these tables?

ssuvorov-fls:
He @chrisknoll 

Fixed according to your suggestions

src/main/resources/db/migration/oracle/V2.7.0.20190128134827__create_absent_sequences.sql
```diff
@@ -0,0 +1,46 @@
+-- analysis_execution
+-- sequence already exists
+
+CREATE OR REPLACE TRIGGER ${ohdsiSchema}.analysis_execution_bir
```
chrisknoll: Triggers are an interesting idea, but we haven't' set up triggers on any of the other tables, and the probelm is, compared to the other DBMS, in oracle if you insert a hard-coded value for the row_id, the trigger will overwrite it, while on the other platforms, it only assigns an ID when the column's insert statement will use the default (ie: the id column is not specified).

So, my concern is the inconsistency of applying a trigger in some tables but not others and the problem where the trigger always overwrites the value.   Plus, I believe the logic will work out such that the Hibernate framework will get th sequence from the DB, then the trigger will overwirte that value with a different sequence....let's forget about adding triggers on these tables.  Since we'll probably be dropping oracle in WebAPI 3.0, don't worry too much about Oracle, just get the simplest implementation to work.

ssuvorov-fls: Done, I removed triggers from scripts

Issue #792, opened 2019-01-31
FK_fac_sucid error during Flyway migration in SQL Server


NEONKID:
When running WebAPI on SQL Server, the following error appears.

```
The object 'created_by_id' is dependent on column 'FK_fac_su_cid'
```
```
The object 'created_date is dependent on column 'DF___...'
```
```
The object 'modified_by_id' is dependent on column 'FK_fac_su_mid'
```
Dependent values that use foreign keys are fixed names, so clearing the constraint will resolve them, but 'create_date' is the default value and there is a variability in the name.

YangDaeIl:
Oh, I have the same error. 
Did you happen to exclude additional modified_by_id; columns?

anthonysena:
Sorry for this issue @NEONKID and @YangDaeIl - we are in the process of applying a fix for this issue here: https://github.com/OHDSI/WebAPI/pull/787.

As noted by @chrisknoll in that PR, once we merge this code into master, you'll need to rename the constraint on fe_analysis_criteria to DF_fe_analysis_criteria_cd (it will likely be something like DF__fe_analys_criter___....) in order for the amended flyway scripts to update your database properly.

NEONKID:
@anthonysena Thanks. It seemed that the dependent value was DEFAULT and needed a defined name.
It seems to be the code resolved if the value dependent on this PR is DEFAULT. #793

anthonysena:
That's exactly right @NEONKID - we made an attempt to revise the flyway script in #787 to provide a name for the default constraint in the hopes that people hadn't yet migrated to that point in master. I would agree that the approach that @wivern put together in #793 is a better approach and will allow us to handle the problem you and @YangDaeIl face. We'll work to merge that all together and put it into master very soon. Thanks!

chrisknoll:
@NEONKID , we've merged various PRs into master to address this, would you mind pulling latest master and trying again? If your issue is resolved, go ahead and close this issue.

PR #787, opened 2019-01-30
SQL Server & Oracle Flyway Patch

anthonysena:
For Oracle/SQL Server, the Flyway script `V2.7.0.20190125113000__fe-analysis-created-modified.sql` created the following columns and foreign key constraints: 

```
ALTER TABLE ${ohdsiSchema}.fe_analysis_criteria ADD created_by_id NUMBER(10) NULL CONSTRAINT FK_fac_su_cid REFERENCES ${ohdsiSchema}.sec_user(id);
ALTER TABLE ${ohdsiSchema}.fe_analysis_criteria ADD modified_by_id NUMBER(10) NULL CONSTRAINT FK_fac_su_mid REFERENCES ${ohdsiSchema}.sec_user(id);
```

Per #782, these columns were created on the wrong table and a patch was issued as part of `V2.7.0.20190129083000__fe-analysis-created-modified-fix.sql`. While attempting to apply this fix in our SQL Server environment, I found that the `fe-analysis-created-modified-fix.sql` script failed to work since we had not dropped the FKs. Furthermore, for SQL Server, I also found that in the original `fe-analysis-created-modified.sql` script did not explicitly name the default constraint which would make creating a patch to drop that default impossible (I presume that SQL Server's column default naming is non-deterministic based on the odd name of: DF__fe_analys_creat__E83F84). 

So, this PR does the following:

- For SQL Server: I've modified the existing scripts `V2.7.0.20190125113000__fe-analysis-created-modified.sql` and `V2.7.0.20190129083000__fe-analysis-created-modified-fix.sql` to use a named default for the column constraint. 
- For both SQL Server & Oracle, I've added in a new script named `V2.7.0.20190129073000__fe-analysis-drop-constraints.sql` which given the flyway naming convention allows it to sit between the two original scripts. This script will drop the FK constraints and for SQL Server, it will also drop the column default constraint.

I attempted to avoid revising existing flyway scripts as discussed in https://github.com/OHDSI/WebAPI/issues/768#issuecomment-457713045 but in this instance I struggled to find a way to safely drop the default since it was not explicitly named originally. 

Please let me know if you have any questions or concerns.

PR #793, opened 2019-01-31
fixes failed MSSQL migration

wivern:
This fixes failed migration for MS Sql Server.
Migration fails due to column constraints that prevents deletion of columns.


anthonysena:
@pavgra: I'm thinking that we should merge this work into #787 since the approach that @wivern has put together would allow us to remove this script: resources/db/migration/sqlserver/V2.7.0.20190129073000__fe-analysis-drop-constraints.sql

I still think we should revise the previous flyway script to explicitly name the default constraint (https://github.com/OHDSI/WebAPI/pull/787/files#diff-24ab3b4f6329de1c8bac632c94144351R5) but @wivern's solution will ensure that the default is dropped in the event of folks that experienced the problem in #792.

pavgra:
@anthonysena , just several minutes ago I was saying the same to @wivern regarding a need to combine the PRs. Will you be able to make this happen?

anthonysena:
Yes, I'll work to merge these together - thanks!

Issue #788, opened 2019-01-30
Fix opening of the new IR


mpozhidaeva:
1)create a new IR
2)try to open it

AR:
![image](https://user-images.githubusercontent.com/31308645/51976367-83b64b80-2495-11e9-8f62-79aec3b1ffb1.png)
Reason (?): https://github.com/OHDSI/WebAPI/commit/cab62ddd195e5d3a25c03c18c0981fa8aae6713b#diff-090531c3c0a520eb294afeb0acb6fb29R451
ER: no errors, returning of the new ir

PR #794, opened 2019-01-31
Added step for checking if IR has executions

aklochkova:
https://github.com/OHDSI/WebAPI/issues/788

### DIFF
--- a/src/main/java/org/ohdsi/webapi/ircalc/IncidenceRateAnalysisRepository.java
+++ b/src/main/java/org/ohdsi/webapi/ircalc/IncidenceRateAnalysisRepository.java
@@ -28,5 +28,8 @@ public interface IncidenceRateAnalysisRepository extends CrudRepository<Incidenc
   Iterable<IncidenceRateAnalysis> findAll();
 
   @Query("SELECT ira FROM IncidenceRateAnalysis AS ira LEFT JOIN FETCH ira.executionInfoList e WHERE ira.id = ?1 AND e.source.deletedDate = NULL")
-  IncidenceRateAnalysis findOneWithExecutions(int id);  
+  IncidenceRateAnalysis findOneWithExecutionsOnExistingSources(int id);
+
+  @Query("SELECT ira FROM IncidenceRateAnalysis AS ira LEFT JOIN FETCH ira.executionInfoList e WHERE ira.id = ?1")
+  IncidenceRateAnalysis findOneWithExecutionsOnAllSources(int id);
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -448,7 +448,11 @@ public JobExecutionResource performAnalysis(final int analysisId, final String s
 
   @Override
   public List<AnalysisInfoDTO> getAnalysisInfo(final int id) {
-    IncidenceRateAnalysis analysis = irAnalysisRepository.findOneWithExecutions(id);
+    IncidenceRateAnalysis analysis = irAnalysisRepository.findOneWithExecutionsOnAllSources(id);
+    
+    if (analysis.getExecutionInfoList().size() > 0){
```
wivern: Just to clarify, why we need to find analysis twice?

src/main/java/org/ohdsi/webapi/ircalc/IncidenceRateAnalysisRepository.java
```diff
@@ -27,6 +27,6 @@
   @Query("SELECT ira FROM IncidenceRateAnalysis AS ira LEFT JOIN FETCH ira.details AS d")          
   Iterable<IncidenceRateAnalysis> findAll();
 
-  @Query("SELECT ira FROM IncidenceRateAnalysis AS ira LEFT JOIN FETCH ira.executionInfoList e WHERE ira.id = ?1 AND e.source.deletedDate = NULL")
-  IncidenceRateAnalysis findOneWithExecutions(int id);  
+  @Query("SELECT ira FROM IncidenceRateAnalysis AS ira LEFT JOIN ira.executionInfoList e LEFT JOIN Source s ON s.id = e.source.id AND s.deletedDate = NULL WHERE ira.id = ?1")
```
pavgra: Why not `LEFT JOIN FETCH`?

aklochkova: Because the following sql is generated when using `LEFT JOIN FETCH`:
`select *
from ir_analysis ir
       left outer join ir_execution e on ir.id = e.analysis_id
       cross join source s
where e.source_id = s.SOURCE_ID
  and ir.id=?
  and (s.deleted_date is null)`

When IncidenceRate doesn't have executions both `e.source_id` and `s.SOURCE_ID` equals to `null`. So this line `where e.source_id = s.SOURCE_ID` is the cause of problem here.

PR #795, opened 2019-02-01
Modified sequence query in sql server

### LINKED ISSUE #753: HIBERNATE_SEQUENCE is not exists in sql server

NEONKID:
Fixed #753 
@chrisknoll I have migrated to SQL Server with reference to PostgreSQL query.

In addition, regardless of this issue, we have verified that the data type of the data column in the Heracles visualization data table is ```TEXT```. However, when use ```select``` query in JPA, exception is caused by using ```distinct```. I think it would be right to use ```union all```. However, for compatibility with other RDBMS, I changed it to ```VARCHAR``` instead of TEXT and set the length to MAX.

### DIFF
--- a/src/main/resources/db/migration/sqlserver/V2.7.0.20190201140945__sources-created-modified-fix.sql
+++ b/src/main/resources/db/migration/sqlserver/V2.7.0.20190201140945__sources-created-modified-fix.sql
@@ -1,3 +1,7 @@
+-- DECLARE current id
+DECLARE @cur_id_val INT;
+DECLARE @sql NVARCHAR(MAX);
+
 -- RENAME source table
 ALTER TABLE ${ohdsiSchema}.source DROP CONSTRAINT DF_source_SOURCE_DIALECT;
 ALTER TABLE ${ohdsiSchema}.source DROP CONSTRAINT source_key_unique;
@@ -7,9 +11,16 @@ EXEC sp_rename '[${ohdsiSchema}].[source]', 'source_bak';
 
 CREATE SEQUENCE ${ohdsiSchema}.source_sequence
   AS BIGINT
-  MINVALUE 0
+  MINVALUE 1
   NO CYCLE
   CACHE 1;
+
+-- GET current ID for source table
+-- and resetting number
+SELECT @cur_id_val = coalesce(MAX(SOURCE_ID), 1) FROM ${ohdsiSchema}.source_bak;
+SET @sql = N'ALTER SEQUENCE ${ohdsiSchema}.source_sequence RESTART WITH ' + CAST(@cur_id_val as NVARCHAR(20)) + ';';
+
+EXEC sp_executesql @sql;
 
 CREATE TABLE ${ohdsiSchema}.source
 (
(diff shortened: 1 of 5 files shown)


NEONKID:
@chrisknoll I created a sql file for the form and modified the query again. Please confirm Thanks.

chrisknoll:
@NEONKID , when you are copying the data from the existing source and source_daimon tables, you are not resetting the sequence value (which will result in future inserts leading to identity collisions when the sequencer yields a value that is already in use). 

You can see an example of how this is done here:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/db/migration/sqlserver/V2.2.5.20180215105415__separate-sequences.sql#L22

You should look at the corresponding scripts for oracle/postgresql to see how that is handled in those environments.

NEONKID:
@chrisknoll Sorry. I did not care about the existing value by testing the query against other schemas only. Thank you for pointing out. Make the commit again by editing the part again.

src/main/resources/db/migration/sqlserver/V2.7.0.20190201140945__sources-created-modified-fix.sql
```diff
@@ -0,0 +1,71 @@
...
+-- GET current ID for source table
+-- and resetting number
+SELECT @cur_id_val = coalesce(MAX(SOURCE_ID), 1) FROM ${ohdsiSchema}.source_bak;
+SET @sql = N'ALTER SEQUENCE ${ohdsiSchema}.source_sequence RESTART WITH ' + CAST(@cur_id_val as NVARCHAR(20)) + ';';
+
+EXEC sp_executesql @sql;
+
+CREATE TABLE ${ohdsiSchema}.source
+(
+  SOURCE_ID         int default NEXT VALUE FOR ${ohdsiSchema}.source_sequence not null constraint PK_source primary key nonclustered,
```
chrisknoll: This will result in an un-named default constraint.  Please look through the other scripts that are under MSSQL to see how named default constraints are defined, and you may notice other things that have been done that could apply to this PR.  I think you are doing it with the source_dialect column, can you try to apply both the PK constratint and the default value together with names?

src/main/resources/db/migration/sqlserver/V2.7.0.20190201140945__sources-created-modified-fix.sql
```diff
@@ -0,0 +1,71 @@
...
+-- GET current id for previous source_daimon table
+-- and resetting number
+SELECT @cur_id_val = coalesce(MAX(source_daimon_id), 1) FROM ${ohdsiSchema}.source_daimon_bak;
+SET @sql = N'ALTER SEQUENCE ${ohdsiSchema}.source_daimon_sequence RESTART WITH ' + CAST(@cur_id_val as NVARCHAR(20)) + ';';
+
+EXEC sp_executesql @sql;
+
+CREATE TABLE ${ohdsiSchema}.source_daimon
+(
+  source_daimon_id int default NEXT VALUE FOR ${ohdsiSchema}.source_sequence not null primary key,
```
chrisknoll: Same non-named default constraint here.

chrisknoll:
@NEONKID , some of these changes were addressed in a separate PR related to missing sequences, see #791 .  I'm sorry that this has been a back-and-forth, but it looks like the sequences addressed in the #791 PR were:
- analysis_execution_sequence
- cca_execution_sequence
- heracles_visualization_data_sequence

Could you alter this PR to just focus on source and source_daimon sequences? and remove migration scripts related to the other PR? I'm sorry the streams got crossed here, and I appreciate your help on this.  If you want us to carry on the last bit of this PR we understand and can pick it up if you want.

PR #796, opened 2019-02-01
Fixes Cohort characterization fails, no further information

wivern:
fixes OHDSI/Atlas#970


pavgra:
@wivern , looks like a potential security issue:
image

I suggest escaping sensitive info if a user doesn't have admin permissions, e.g.:
`jdbc:posgtresql://...` -> `[DSName connection string]`
`postgres` -> `[DSName username]`
`secret` -> `[DSName password]`

pavgra:
@wivern , the code fails when I query cohort definition which has generations for Sources that were removed from DB:
image
image

Issue #243, opened 2017-09-01
Concept record counts bug


anthonysena:
Came across an issue internally with the concept: 594249 - Measles Vaccine. Internally, for DOD, this concept_id is reported to have 131 records and 262 descendant records. This was not matching another report that I had generated.

This query helped me to identify the bug directly against the source:

```
with concepts AS(
	select c2.*
	from concept c
	INNER JOIN concept_ancestor ca ON ca.ancestor_concept_id = c.concept_id
	INNER JOIN concept c2 ON c2.concept_id = ca.descendant_concept_id
	where c.concept_id = 594249
), exactMatch AS (
	select COUNT(*) rec_cnt, COUNT(DISTINCT person_id) person_cnt
	from drug_exposure de 
	where de.drug_concept_id = 594249
), descMatch AS (
	select COUNT(*) desc_rec_cnt, COUNT(DISTINCT de.person_id) desc_person_count
	from drug_exposure de 
	INNER JOIN concepts c ON de.drug_concept_id = c.concept_id
)
SELECT *
FROM exactMatch, descMatch
```
The query above returns 0 for exact matches for that concept_id and 131 for descendants; Atlas says there are 131 exact record matches and 262 descendants.

I think the join at this section of the record count cache points to the issue whereby we're considering both ancestors and descendants for each concept:

https://github.com/OHDSI/WebAPI/blob/68d9fe2fb9356feb3c620767cfb5f47d961dd333/src/main/resources/resources/cdmresults/sql/loadConceptRecordCountCache.sql#L53

Tagging @fdefalco for assistance on this one.

mpozhidaeva:
@anthonysena, @fdefalco, @pavgra - I spent some time investigation the issue and I guess the reason is in including records from _concept_ancestor_ table with _ancestor_concept_id = descendant_concept_id_.

In odysseus-test environment (master webapi, datasource - synpuf 110k, vocabularies - 05-OCT-18) I see 
![image](https://user-images.githubusercontent.com/31308645/52225429-5fa2a200-28bb-11e9-847c-64fdbf1b8efb.png.
The result count is a sum of the analyzes 701 and 901.
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cdmresults/sql/getConceptRecordCount.sql

For 901 analysis I get all descendant records only for record ancestor_concept_id = descendant_concept_id (=594249). I think they are incorrect.

 concept_id | record_count | descendant_record_count
[...]

PR #803, opened 2019-02-05
Excluded some records from concept_ancestor for counting DRC & RC

mpozhidaeva:
Fixes https://github.com/OHDSI/WebAPI/issues/243


anthonysena:
From discussion with @pavgra and @chrisknoll: we'll want to review this one to ensure that we don't lose the ability to include the current concept in the DRC count since this is helpful to identify a concept + its descendants that have records.

anthonysena:
@mpozhidaeva - I hate to do this since you put in time to make these changes but unfortunately this approach (which I suggested in #243) would change the behavior in ATLAS since the current concept's records would not be included in the descendant record count. The concept I used in #243 has some special properties which I detailed in that issue. We'll revisit this for the next release and I'll close out this PR.

Issue #779, opened 2019-01-28
(treatment) pathways produces some unanticipated values


mick-iqvia:
We are looking at the behaviour of treatment pathways in order to implement a version running against a spark cluster.

There are a number of results that we are not sure are what was intended.

These are outlined in the image below, where 100, 200 etc are person ids.

![pathwaysanomolies](https://user-images.githubusercontent.com/47112987/51849191-18466f80-2317-11e9-9028-40dbd8197ad9.png)

The columns are: 
1. Input events - event cohorts - assume all these events fall within target
2. Current output - these are current pathways
3. Required output - these are the results we expected
4. Henry has modified runPathways SQL and these are the outputs following his changes. He can comment on this.

I'll add the test data in a follow up comment

mick-iqvia:
Here is some test data that has been copied from a test fixture we have in a scala file. Hopefully it is straightforward to understand. The tests from "Overlapping" on are not producing expected outputs.

`case class PathwaysTestExample(name: String, comboWindow: Int, allowRepeats: Boolean, maxDepth: Int,
                               targetCohortIds: Seq[Int], eventsCohortIds: Seq[Int],
                               inputCohorts: String, expectedEvents: String, expectedStats: String) {
  val inputTable = TTable(inputCohorts)
  val expectedEventTable = TTable(expectedEvents)
  val expectedStatsTable = TTable(expectedStats)
}

object PathwaysTestExamples {
  val examples: Seq[PathwaysTestExample] = Seq(
    PathwaysTestExample("Filtered events", 0, allowRepeats = false, 5, Seq(1), Seq(2, 3, 4, 5),
      inputCohorts =
        """cohort_definition_id|subject_id|cohort_start_date|cohort_end_date
          |int                 |bigint    |date             |date
          |1                   |999       |2017-01-18       |2017-01-18
          |2                   |999       |2017-01-18       |2017-01-18
          |3                   |999       |2017-01-17       |2017-01-19
          |4                   |999       |2017-01-18       |2017-01-19
          |5                   |999       |2017-01-17       |2017-01-18""",
      expectedEvents =
        """pathway_analysis_generation_id |target_cohort_id |combo_id |subject_id |ordinal |cohort_start_date |cohort_end_date
          |bigint                         |int              |bigint   |bigint     |int     |date              |date
          |1                              |1                |1        |999        |1       |2017-01-18        |2017-01-18""",
      expectedStats =
[...]

PR #808, opened 2019-02-08
Fix to overlap logic

hms1:
Implements the changes discussed in [https://github.com/OHDSI/WebAPI/issues/779](url) .

It is important to note that the new overlap logic will recognise a one day overlap between two cohorts where `cohort1.cohort_end_date = cohort2.cohort_start_date`. On its own, this is probably technically correct given the OMOP convention for cohort dates to be inclusive.

However, there is an interaction with the collapse window functionality. If one `cohort_end_date` is within `collapse_window` of another `cohort_start_date`, a one day overlap will be artificially created. While I'm guessing these scenarios are very much edge cases, I doubt it's the behaviour we want. 

[mick-iqvia](/mick-iqvia) and I have been discussing some alternative ways that closely related event windows could be collapsed. We may want to consider incorporating into the same release as this PR.

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -19,9 +19,9 @@ FROM (
 	  e.subject_id,
 	  e.cohort_start_date,
 	  e.cohort_end_date
-	FROM @target_database_schema.@target_cohort_table e
+	FROM @target_cohort_table e
 	  JOIN ( @event_cohort_id_index_map ) ec ON e.cohort_definition_id = ec.cohort_definition_id
-	  JOIN @target_database_schema.@target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_end_date <= t.cohort_end_date AND t.subject_id = e.subject_id
+	  JOIN @target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_end_date <= t.cohort_end_date AND t.subject_id = e.subject_id
 	WHERE t.cohort_definition_id = @pathway_target_cohort_id
 ) RE;
 
@@ -168,7 +168,7 @@ SELECT
   pathway_count.cnt AS pathways_count
 FROM (
   SELECT COUNT(*) cnt
-  FROM @target_database_schema.@target_cohort_table
+  FROM @target_cohort_table
   WHERE cohort_definition_id = @pathway_target_cohort_id
 ) target_count,
 (
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -19,9 +19,9 @@ FROM (
 	  e.subject_id,
 	  e.cohort_start_date,
 	  e.cohort_end_date
-	FROM @target_cohort_table e
+	FROM @target_database_schema.@target_cohort_table e
 	  JOIN ( @event_cohort_id_index_map ) ec ON e.cohort_definition_id = ec.cohort_definition_id
-	  JOIN @target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_end_date <= t.cohort_end_date AND t.subject_id = e.subject_id
+	  JOIN @target_database_schema.@target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_end_date <= t.cohort_end_date AND t.subject_id = e.subject_id
 	WHERE t.cohort_definition_id = @pathway_target_cohort_id
 ) RE;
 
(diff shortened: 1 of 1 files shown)


chrisknoll:
I'm concerned about the issue you raised:  I understand that certain people in the OHDSI community consider the date diff between 1/1/2001 and 1/1/2001 to be 1 day.  however, that's fur purposes of calculating the duration of a time span assuming the end date includes all 24 hours and the start date does not.  It's this confusion of how to interpret a date is something we've avoided in CIRCE by just treating those dates as 'face value' such that it works out if I add 7 days to 1/1 I get 1/8, and when i subtract 1/8 from 1/1 i get the 7 days.  Treating the 1/8 day as 'special end date handling' means there is actually 8 days between 1/1 and 1/8.

This is all to say:  we should be treating the individual dates as 'face value dates'.  In the case of TxP:  I would not expect the following:

Cohort|start|end|
---|----|----|
A|1/1|1/8
B|1/8|1/12

to result in the pathway:

A -> A+B -> B

It should just be:

A -> B

Like you said, this is only an issue where you have an A that ends exactly when B starts, but with collapse window, you're forcing this to happen to make overlaps not appear.

So, does your change _depend_ on a end date being treated as `date + 1` or can we just leave the current semantics that dates are just dates?

src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -88,62 +88,32 @@ into
...
-  FROM #split_overlapping_events
-)
-SELECT SUM(DISTINCT POWER(2, e.event_cohort_index)) as combo_id, subject_id, cohort_start_date, cohort_end_date
+WITH 
+cohort_dates AS (
+	SELECT DISTINCT subject_id, cohort_date 
+	FROM (
+		  SELECT subject_id, cohort_start_date cohort_date FROM #collapsed_dates_events 
+		  UNION 
+		  SELECT subject_id, DATEADD(day, 1, cohort_end_date) cohort_date FROM #collapsed_dates_events
```
chrisknoll: I see this is where we're padding the end date by 1 day.  is this critical to the algorithm working correctly?  Or is this just a mater of the perspective that an end date includes all hours of the current day?

hms1: The addition of one day is used to define non-overlapping time periods, which is important if we want to follow the general approach suggested. 

It is later on in the query (line 113) where we determine which event cohorts are 'active' during each time period. It'd probably be better to make alterations related to interpretation of cohort start/end date inclusivity/exclusivity here.

hms1:
> Cohort	start	end
> A	1/1	1/8
> B	1/8	1/12
> to result in the pathway:
> 
> A -> A+B -> B
> 
> It should just be:
> 
> A -> B
> 

While I totally agree that A -> B makes more sense intuitively, it is tricky to reconcile a single treatment with other examples. For example, what about the following:

| Cohort  | Start | End |
| ------------- | ------------- | ------------- | 
| A  | 1/1  | 1/1 |
| B  | 1/1  | 1/8 |

Would we not want A + B -> B in this case? 

You could potentially handle these one day cohorts by changing line 113 to:

`WHERE (e.cohort_start_date <= tp.next_cohort_date AND e.cohort_end_date > tp.cohort_date)
OR  (e.cohort_start_date = tp.cohort_date)` (this would need much more careful thought and testing, but you get the idea)

It gets a little messy for my liking, though. The fact that you can have a cohort that starts and ends on the same day makes me think that cohort dates are inclusive (I'm happy to be persuaded otherwise though).

Another approach would be for the SQL to return the most granular pathways as possible, i.e. assume that cohort dates are inclusive, do not collapse any dates and return all overlaps (however small), and then do some post-processing based on user-defined inputs to tidy up. @mick-iqvia has some suggestions related to this, which I'm guessing you'll discuss next week :)

chrisknoll:
> Would we not want A + B -> B in this case?

Yes, and my understanding of how this would resolve is:

| Cohort  | Start | End |
| ------------- | ------------- | ------------- | 
| A  | 1/1  | 1/1 |
| B  | 1/1  | 1/1 |
| B  | 1/1  | 1/8 |

And because a and b have the exact same start/end, we get A+B -> B

I need to think more of your suggestion there, but I think we get into some problem space when we have to think about 'inclusive end dates' vs non-inclusive...

chrisknoll:
Per our Conference Call on 2/26:  we're going to adjust the algorithm to instead of 'collapsing dates' we decided to go with 'minimum span length' where we will drop any segment that is < the minimum span length.

For simplicity, we are going to use the 'collapseWindow' value as the minimum span length in the algorithm, and separately we will rename the fields in the objects to reflect 'minSpanLength' in the Entity/Tables/paramaters in a new PR (after this has been merged with master).

chrisknoll:
Hi @hms1 , the pathway analsysis query has been updated in master since this PR, but the conflicting lines is just realted to some CAST operations that were required for some SQL dialects. Would you mind syncing your upstream with the latest master branch in WebAPI, then resolve conflicts?

Otherwise, what we can do is take this PR into a side-branch on the OHDSI repository, resolve the conflicts there, and make another PR with your updatse that will be based of an OHDSI/WebAPI branch.

chrisknoll:
Hi, @hms1. I don't think we'll be able to fold this in for the 2.7 release.  We're pretty far over-due with making this release, we'd still want to do extensive testing on the changes, and I think it is best that we not overload the 'collapse window' value with the 'minimum span length'. To handle the new paramater, we really should think about updating the object model to add the new parameter and alter the database tables to store it.  So, this would better be a 2.8 item.  Sorry for the inconvenience.

hms1:
> Hi, @hms1. I don't think we'll be able to fold this in for the 2.7 release. We're pretty far over-due with making this release, we'd still want to do extensive testing on the changes, and I think it is best that we not overload the 'collapse window' value with the 'minimum span length'. To handle the new paramater, we really should think about updating the object model to add the new parameter and alter the database tables to store it. So, this would better be a 2.8 item. Sorry for the inconvenience.

hms1:
No problem at all... when is 2.8 planned for? 

I've been working with the Pathways scripts quite a lot over the past few weeks, and have a couple of additional suggestions that could be considered if there's time.

chrisknoll:
2.8 should be sometime end of April.  Sorry I did not reply to this sooner.  Did you mean to close this PR?  I was hoping we could start working on incorporating these changes.

hms1:
Sorry, must have closed it by accident.

Happy to start working on this again if there's anything else you need.

pavgra:
@chrisknoll , there are conflicts in PR. Could you resolve?

chrisknoll:
It's in a fork, so I can't commit anything to it. However, I wanted to create a new branch that incorporates the proposed changes here, allows the existing functionality of collapse window, and introduces the new paramater of 'minSegmentLength' described here: https://github.com/OHDSI/StandardizedAnalysisAPI/pull/23 and here: https://github.com/OHDSI/WebAPI/pull/1049. Unfortunately, my attention has been very distracted with other issues and I haven't been able to find time to drill into this work and complete it. I plan on making this my solitary focus int he coming week.

anthonysena:
Linking back to #779.

chrisknoll:
Will use #1211 for this fix.

Issue #756, opened 2019-01-21
Foreign key on source_daimon source_id


SBylemans:
There should be a foreign key constraint on the `source_daimon` table, on the `source_id` column. It should point to `source_id` of the `sources` table.

    ALTER TABLE ${ohdsiSchema}.source_daimon ADD CONSTRAINT FK_source_daimon_source_id FOREIGN KEY (source_id) REFERENCES ${ohdsiSchema}.source (source_id);

PR #814, opened 2019-02-12
Add source seq for sqlserver...

NEONKID:
@chrisknoll I have confirmed the overlapped part of PR. So I worked on the name of another branch and removed the part already applied. (Continuous #795)

PR #818, opened 2019-02-13
Added FK to source_daimon

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/756


chrisknoll:
This might depend on #814, in that the identity column is being reset in that PR so we should ensure that the #814 merges in before this one.  Also, to confirm, the migration scripts from #814 should be pre-ordered to this one:

From #814: V2.7.0.201902130900__source-sequences.sql
From #818: V2.7.0.20190213161124__add-fk-to-source-daimon.sql

So I believe the #814 will be applied prior to #818 (provided both #814 and #818 are merged together).  So, this PR should not be tested without the changes for #814 applied first.

Issue #819, opened 2019-02-13
WebAPI fails to retrieve CC generations from MS SQL 2017


pavgra:
WebAPI logs show:
```
2019-02-13 12:23:32.235 WARN http-nio-8080-exec-9 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - SQL Error: 245, SQLState: S0001
2019-02-13 12:23:32.236 ERROR http-nio-8080-exec-9 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - Conversion failed when converting the varchar value 'false' to data type int.
2019-02-13 12:23:43.483 WARN http-nio-8080-exec-7 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - SQL Error: 245, SQLState: S0001
2019-02-13 12:23:43.483 ERROR http-nio-8080-exec-7 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - Conversion failed when converting the varchar value 'false' to data type int.
2019-02-13 13:02:27.146 WARN http-nio-8080-exec-6 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - SQL Error: 245, SQLState: S0001
```

And execution of the following query:
```
SELECT cc.id
FROM webapi_pg.cc_generation ccg
  JOIN webapi_pg.cohort_characterization cc ON ccg.cc_id = cc.id
  JOIN webapi_pg.source s ON s.SOURCE_ID = ccg.source_id
WHERE ccg.cc_id = 1;
```

Throws the exception:
```
[S0001][245] Conversion failed when converting the varchar value 'Generating cohort characterization 1 : Local SynPUF 110k (local_110k)' to data type int.
```

Which indicates MPP-style of MSSQL 2017 behavior (seems like it processes joins / where after select, therefore the CAST in select is executed for all possible values from the column).

The solution which is going to be applied is usage of `TRY_CAST` instead of `CAST`:
`CAST(cc_id_param.string_val AS INTEGER)` -> `TRY_CAST(cc_id_param.string_val AS INTEGER)`

PR #820, opened 2019-02-13
Fixed cc_generation view in MS SQL 2017

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/819


anthonysena:
Noting this MSFT article for reference for this fix since TRY_CAST is compatible with SQL Server 2012 and up:

https://docs.microsoft.com/en-us/sql/t-sql/functions/try-cast-transact-sql?view=sql-server-2017

PR #823, opened 2019-02-14
Modified temp table query for digits..

### LINKED ISSUE #822: Results DDL fails to SELECT for SQL Server
```sql
select 0 as n union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9 into #digits;
```
In SQL Server, when executing the query, SELECT INTO must be the first query in a statement containing a UNION, INTERSECT or EXCEPT operator error.

NEONKID:
Fixed #822

### DIFF
--- a/src/main/resources/ddl/results/init_heracles_periods.sql
+++ b/src/main/resources/ddl/results/init_heracles_periods.sql
@@ -1,6 +1,5 @@
-select 0 as n union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9
-into #digits;
-
+select 0 as n into #digits union all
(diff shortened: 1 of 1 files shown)


chrisknoll:
@neonkid, would you mind structuring this query into the following format:

```
select n  into #digits 
FROM (
select 0 as n union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9;
) N

```

The concern is how SqlRender is going to treat this when translating across different dialects, and usually wrapping it into a subquery is safer than doing the select...into...union all...select.

NEONKID:
Thanks for your review. I modified it as advised and confirmed the operation in SQL Server.

Issue #813, opened 2019-02-12
concept_hierarchy init should include truncate


alondhe:
In the DDL Results service, if using the concept_hierarchy initialization, the script should truncate concept_hierarchy before the inserts

PR #825, opened 2019-02-14
Added truncation of concept_hierarchy

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/813


src/main/resources/ddl/results/init_concept_hierarchy.sql
```diff
@@ -251,6 +251,9 @@ LEFT JOIN @vocab_schema.concept c3 ON ca3.ANCESTOR_CONCEPT_ID = c3.concept_id
 GROUP BY obs.concept_id, obs.concept_name;
 
 /********** PROCEDURE **********/
+
+TRUNCATE TABLE @results_schema.concept_hierarchy;
```
pavgra: @aklochkova , what do you think is going to happen with inserted data for conditions, drugs, measurements, observations?

aklochkova: @pavgra it was a mistake, fixed

PR #834, opened 2019-02-15
Add spring-batch sequence for sql server...

### LINKED ISSUE #800: SQL Server batch job sequence tables
For SQL Server, we still are utilizing tables to keep track of the last used ID instead of sequences. Specifically these tables:

- BATCH_JOB_EXECUTION_SEQ
- BATCH_JOB_SEQ
- BATCH_STEP_EXECUTION_SEQ

This is inconsistent with the other RDBMS platforms so noting it here so we can make it consistent.

NEONKID:
Fixed #800


chrisknoll:
So, I'm a little concerned about this change: Spring Batch is just a third party library that provided the DDL for the various platforms (sql, postgresql, oracle) and I'm not sure if we should change the published DDL that they provided to use sequences.

@pavgra do you have any thoughts?

pavgra:
I would also not touch the tables managed by a separate module (Spring Batch in this case)

chrisknoll:
Ok, thanks @pavgra .

@NEONKID : I think we should close this PR since we don't want to change the tables defined by the external library.

NEONKID:
Good. I agree with this and I will close this PR.

chrisknoll:
Re-opening this PR.  There is a known issue with SQL Server and using table-based sequences.  The issue is described at a higher-level in the spring framework here: https://github.com/spring-projects/spring-framework/issues/21425.  However, I believe there are extension 'hooks' in spring batch allowing you to provide a custom implemenation of the key generator:  the custom generator would do a `select next value for {sequence}` call instead of using the `_SEQ` tables.

All the work in this PR at the database-level to migrate from the table sequences to the sequence-based sequences can stay.  The missing part is overriding the key generator in the event it is SQLServer.  

If we decide to drop MSSQL/Oracle support in WebAPI (ie: everything is based on postgresql), this issue becomes moot. However, I don't think we have official position on this change, so we may still need to keep MSSQL and Oracle supported. (Would be easier for us to simply support postgresql, but this may be a change that is introduced in the 3.0 release, and not 2.8).

chrisknoll:
Going to close this for now, since it's more involved to get spring batch sequences implemented on SQL Server, and we will probably be focusing on Postgres for WebAPI.

PR #839, opened 2019-02-18
fixes security restrictions for version info endpoint

wivern:
fixes security restrictions for version info endpoint


pavgra:
@wivern , this won't pass the test - https://github.com/OHDSI/WebAPI/commit/8ae6aaf8784070772c092d1c79b66c7650fae34d - seems like we need to add exclusions list for the test and put the `/info` endpoint there

PR #842, opened 2019-02-19
fixes PLP/PLE execution

wivern:
fixes execution for PLP/PLE analyses

### DIFF
--- a/src/main/java/org/ohdsi/webapi/executionengine/controller/ScriptExecutionController.java
+++ b/src/main/java/org/ohdsi/webapi/executionengine/controller/ScriptExecutionController.java
@@ -46,7 +46,7 @@ import org.springframework.stereotype.Component;
 public class ScriptExecutionController {
 
     private final Log logger = LogFactory.getLog(ScriptExecutionController.class);
-    private static final String DATASOURCE_PERMISSION = "source:%s:get";
+    private static final String DATASOURCE_PERMISSION = "cohortdefinition:*:generate:%s:get";
     private static final String FORBIDDEN_MESSAGE = "Access to source with key %s is forbidden";
 
     @Value("${executionengine.resultCallback}")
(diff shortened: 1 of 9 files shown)


pavgra:
@wivern , could you verify that upgrade from 2.5.2 (2.5.3) to 2.7.0 won't fail after the PR?

wivern:
> @wivern , could you verify that upgrade from 2.5.2 (2.5.3) to 2.7.0 won't fail after the PR?

I've checked upgrade to 2.7 with Sql Server and Oracle, it's ok with #844

src/main/java/org/ohdsi/webapi/executionengine/controller/ScriptExecutionController.java
```diff
@@ -52,6 +46,8 @@
 public class ScriptExecutionController {
 
     private final Log logger = LogFactory.getLog(ScriptExecutionController.class);
+    private static final String DATASOURCE_PERMISSION = "source:%s:get";
```
pavgra: Invalid permission for the case

src/main/java/org/ohdsi/webapi/executionengine/controller/ScriptExecutionController.java
```diff
@@ -97,6 +95,10 @@ public JobExecutionResource runScript(ExecutionRequestDTO dto) {
         logger.info("Received an execution script to run");
         Source source = sourceRepository.findBySourceKey(dto.sourceKey);
 
+        if (securityEnabled && !SecurityUtils.getSubject().isPermitted(String.format(DATASOURCE_PERMISSION, source.getSourceKey()))) {
+            throw new ForbiddenException(String.format(FORBIDDEN_MESSAGE, source.getSourceKey()));
```
pavgra: Throws 500 instead of 403

PR #844, opened 2019-02-20
Adds checks providing safe migration from 2.5.2

wivern:


pavgra:
@wivern , why don't we provide the same checks for PG and Oracle?

chrisknoll:
Guys, just to be clear: we shoudln't need to do those checks.  If you're making a migration that says 'if this doesn't exist yet, then do it', I think you are doing it wrong.  Each migration should have a firm set of assumptions about what is and what is not int he current schema.  If that assumption is wrong, then the migration should fail and it probably means someone did something 'out of band' that they shouldn't have.

pavgra:
Dear Chris,

I would be so grateful if you propose another way to resolve the case:
- Atlas 2.5.2 deployed to client A
- Atlas 2.7.0 is in master and being developed
- A critical issue is found, migration is added into 2.7.0
- The part of migration is also required in a maintenance version (2.5.3). But only a part, therefore, the fix goes as a separate migration
- A time comes for upgrading client A from 2.5.3 to 2.7.0. So the migration from 2.7.0 is applied and affects the same tables/columns as the ones which we were already affected by 2.5.3. Crash.

And specifically, to avoid too theoretical thoughts: we had to transfer creation of sequence into 2.5.3 version while the sequence was already in migrations of 2.7.0 - your solution?

chrisknoll:
> * The part of migration is also required in a maintenance version (2.5.3). But only a part, therefore, the fix goes as a separate migration
> * A time comes for upgrading client A from 2.5.3 to 2.7.0. So the migration from 2.7.0 is applied and affects the same tables/columns as the ones which we were already affected by 2.5.3. Crash.
> 
> 

You wouldn't do 'part' of the migration over to 2.5.3, you would have to carry it over completely as an atomic unit.  (and keep the same name, so that when they go 2.5.3->2.7.0 the 2.7.0 deploy will recognize you got the migration and won't apply again).  You have to adhere to that principle otherwise you can't do it.

pavgra:
If there is such simple case, I wouldn't do and didn't do (e.g. https://github.com/OHDSI/WebAPI/pull/829/files) the partial approach. But I _would_ do the part if the full migration cannot be transferred to the maintenance version (e.g. new migration uses newer tables which are not available in the older version, and which are created by separate migrations).

chrisknoll:
@pavgra , understood, there's an exception to every rule, of course, just when creating these migrations, we should be checking if the migration is constructed in a way that can be applied to the maintenance branch by just moving the migration script into the maint branch,

To your use cases:

> And specifically, to avoid too theoretical thoughts: we had to transfer creation of sequence into 2.5.3 version while the sequence was already in migrations of 2.7.0 - your solution?

I would have brought all the sequence migration work into the 2.5.3 branch directly using the same migration scripts, and handled the necessary Entity `@sequence` annotations in the Java source as a commit on the 2.5.3 branch. If the single migration script handled both creating new tables and migrating old sequences, then I would have said that we would need to be more careful about migration scripts trying to do too much in 1 migration and did what we needed to do to get it to work, including conditional checks.  

> If there is such simple case, I wouldn't do and didn't do

I think we're both in agreement here that you should try to avoid conditionals that might make assumptions about the state of a schema and allow a migration to proceed when it should otherwise fail, and a certain amount of flexibility is required in order to get the simplest solution in place, such as conditional checks. I hope you remember this when we talk about code structure and interface layers later :).  For purposes of this discussion, it would have been enough to say 'we would try to do that, except in this case there were migration scripts that were doing multiple things in a single migration, so were were not able to pull in only the necessary change that we needed.'  When I saw it in the migration, i just wanted to remind us of what we should be trying to do. as we come across complex cases, we'll have to make those exceptions.

wivern:
> @wivern , why don't we provide the same checks for PG and Oracle?

Do not need this for PG and Oracle since with Sql Server the issue was with autoincrement fields but Oracle and PG does not support autoincrement.
I've tested upgrading with Oracle and it was passed.

PR #849, opened 2019-02-21
Added condition to retrieve only results above threshold

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/1357


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -66,12 +66,14 @@
 public class CcServiceImpl extends AbstractDaoService implements CcService, GeneratesNotification {
 
     private static final String GENERATION_NOT_FOUND_ERROR = "generation cannot be found by id %d";
-    private static final String[] GENERATION_PARAMETERS = {"cohort_characterization_generation_id"};
+    private static final String[] GENERATION_PARAMETERS = {"cohort_characterization_generation_id", "threshold_level"};
```
chrisknoll: I'm not sure why we want the threshold_level supplied as a generation_parameter since we only filter to a threshold level when fetching the results?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -66,12 +66,14 @@
...
     private static final String[] PREVALENCE_STATS_PARAMS = {"cdm_database_schema", "cdm_results_schema", "cc_generation_id", "analysis_id", "cohort_id", "covariate_id"};
-    private final String QUERY_RESULTS = ResourceHelper.GetResourceAsString("/resources/cohortcharacterizations/sql/queryResults.sql");
+    private final String QUERY_FULL_RESULTS = ResourceHelper.GetResourceAsString("/resources/cohortcharacterizations/sql/queryFullResults.sql");
+    private final String QUERY_POPULAR_RESULTS = ResourceHelper.GetResourceAsString("/resources/cohortcharacterizations/sql/queryPopularResults.sql");
```
chrisknoll: Is the difference between Popular_results and _full_results just the WHERE clause?  Because it seems that this is redundant when we could just paramaterize a single query?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
aklochkova: I thought it would be good to give a name to "0.01" so I added it to parameters

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
aklochkova: yes, fixed

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -413,8 +412,8 @@ public CcGenerationEntity findGenerationById(final Long id) {
         final CcGenerationEntity generationEntity = ccGenerationRepository.findById(generationId)
                 .orElseThrow(() -> new IllegalArgumentException(String.format(GENERATION_NOT_FOUND_ERROR, generationId)));
         final Source source = generationEntity.getSource();
-        String generationResults = sourceAwareSqlRender.renderSql(source.getSourceId(), isFullRetrieve ? QUERY_FULL_RESULTS : QUERY_POPULAR_RESULTS, GENERATION_PARAMETERS, 
-                new String[]{String.valueOf(generationId), String.valueOf(thresholdLevel)});
+        String generationResults = sourceAwareSqlRender.renderSql(source.getSourceId(), QUERY_RESULTS, GENERATION_PARAMETERS,
```
chrisknoll: I'm still confused about this 'GENERATION_PARAMATERS':  is this really paramaters for generation, or is it parameters for fetching the results?  If it is for the results, call these constants: QUERY_RESULTS and PARAMATERS_RESULTS.  If it is for generation, then it is properly named, but i don't see the GENERATION_PARAMATERS used during the generation process, or am I missing it?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
```diff
@@ -234,8 +236,8 @@ public CcExportDTO getGenerationDesign(
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)
     public List<CcResult> getGenerationsResults(
-            @PathParam("generationId") final Long generationId) {
-        List<CcResult> ccResults = service.findResults(generationId);
+            @PathParam("generationId") final Long generationId, @DefaultValue("false") @QueryParam("fullRetrieve") final boolean isFullRetrieve) {
```
pavgra: Why not `@DefaultValue("0.01") @QueryParam("threshold") final Float threshold)`?

src/main/resources/resources/cohortcharacterizations/sql/queryResults.sql
```diff
@@ -22,4 +22,4 @@ select
        r.strata_id,
        r.strata_name
 from @results_database_schema.cc_results r
-where r.cc_generation_id = @cohort_characterization_generation_id
+where r.cc_generation_id = @cohort_characterization_generation_id and r.avg_value > @threshold_level
```
pavgra: @chrisknoll, do we want to apply this for Distribution analyses? I am very unsure from the applied perspective

chrisknoll: Ah, the avg_value in a distribution isn't the same as a prevalence.  In the case of a prevalence, the avg_value goes 0 to 1 (0% to 100%).  In the cases of distribution, it's the avg value per person, so like could be avg records per person which can go from 0 to any number upper bound.

So, I don't think it would really mean anything here, but if it's complicated to make it only apply to prevalence statistics, it probably won't have any meaningful effect on distribution statistics because most things we count won't have an avg value of < .01.

But, applying it to distribution stats might be confusing looking for those who have never read this thread. So ideally, we treat this filter towards prevalence stats only.

pavgra: > if it's complicated to make it only apply to prevalence statistics

WHERE r.type = 'PREVALENCE'

Issue #856, opened 2019-02-22
Cohort / IR results for a DS are shown to users who don't have access to the DS


pavgra:
![image](https://user-images.githubusercontent.com/10010071/53251467-b9acb100-368a-11e9-8511-42f18e6d0f70.png)

PR #860, opened 2019-02-22
Fixes access to results of Cohorts / IRs

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/856

### DIFF
--- a/src/main/java/org/ohdsi/webapi/ShiroConfiguration.java
+++ b/src/main/java/org/ohdsi/webapi/ShiroConfiguration.java
@@ -86,6 +86,7 @@ public class ShiroConfiguration {
   }
 
   @Bean
+  @ConditionalOnMissingBean(value = DisabledSecurity.class)
   public DataSourceAccessBeanPostProcessor dataSourceAccessBeanPostProcessor(DataSourceAccessParameterResolver parameterResolver) {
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/ohdsi/webapi/ShiroConfiguration.java
```diff
@@ -86,7 +86,6 @@ public LockoutStrategy lockoutStrategy(){
   }
 
   @Bean
-  @ConditionalOnMissingBean(value = DisabledSecurity.class)
   public DataSourceAccessBeanPostProcessor dataSourceAccessBeanPostProcessor(DataSourceAccessParameterResolver parameterResolver) {
```
wivern: Why we require DataSource access checking when security is disabled?

src/main/java/org/ohdsi/webapi/shiro/management/datasource/BaseDataSourceAccessor.java
```diff
@@ -1,17 +1,32 @@
...
+  private DisabledSecurity disabledSecurity;
+
+  public void checkAccess(T s) {
+    if (!hasAccess(s)) {
       throw new ForbiddenException();
     }
   }
 
+  public boolean hasAccess(T s) {
+    Source source = extractSource(s);
```
wivern: We have redundant query here when security is disabled

PR #845, opened 2019-02-20
Adds generation of new PLE/PLP via ExecutionEngine

wivern:
closes #805

PR #861, opened 2019-02-22
Import for PLE/PLP

anthonysena:
Aims to add import methods for PLE/PLP to address OHDSI/Atlas#1335. Some notes related to the changes:

- Refactored `EstimationCohortDefinition` and `PredictionCohortDefinition` into **`AnalysisCohortDefinition`**. Also refactored `EstimationConceptSet` and `PredictionCohortDefinition` into **`AnalysisConceptSet`**. This eliminated code redundancy and also enabled me to add methods for importing these portions of the specification in a consistent manner as detailed in the `org.ohdsi.webapi.analysis.converter` and `org.ohdsi.webapi.analysis.importer` namespaces, respectively.
- Updated the StandardizedAnalysisAPI https://github.com/OHDSI/StandardizedAnalysisAPI/pull/17 based on testing that corrected a bug in the way that the Cyclops Prior is specified.
- Removed redundant EstimationTypeEnum from WebAPI in favor of the one in StandardizedAnalysisAPI.

Per discussion with @pavgra, we'll hold this PR until #845 is approved and merged. Once that is done, I'll merge those changes into this branch for review. In the meantime, if there are any concerns with what has been done so far, please let me know and I'll address them while #845 is still in progress.


anthonysena:
This is now ready for review. Updates in this commit: https://github.com/OHDSI/WebAPI/pull/861/commits/71a4e27dc7e20f31419e40d446a01ba6f74368ac address OHDSI/Atlas#1221

src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
```diff
@@ -260,8 +262,109 @@ public EstimationAnalysisImpl exportAnalysis(Estimation est) {
...
+            // Open up the analysis and get the relevant portions for import
+            Settings settings = analysis.getEstimationAnalysisSettings().getAnalysisSpecification();
+            ComparativeCohortAnalysisImpl ccaSpec = (ComparativeCohortAnalysisImpl) settings;
+            List<TargetComparatorOutcomesImpl> tcoList = ccaSpec.getTargetComparatorOutcomes();
+            List<CohortMethodAnalysisImpl> ccaList = ccaSpec.getCohortMethodAnalysisList();
+            
+            // Create all of the cohort definitions 
+            // and map the IDs from old -> new
+            Map<Long, Long> cohortIds = new HashMap<>();
+            AnalysisCohortDefinitionImportService cohortImportService = new AnalysisCohortDefinitionImportService(security, userRepository, cohortDefinitionRepository);
```
pavgra: Why do you use bean as POJO?

src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
```diff
@@ -187,9 +193,77 @@ public PatientLevelPredictionAnalysisImpl exportAnalysis(int id) {
...
-    public void hydrateAnalysis(org.ohdsi.webapi.prediction.specification.PatientLevelPredictionAnalysisImpl plpa, OutputStream out) throws JsonProcessingException {
-        String studySpecs = Utils.serialize(plpa, true);
+    @Override
+    public PredictionAnalysis importAnalysis(PatientLevelPredictionAnalysisImpl analysis) throws Exception {
+        try {
+            // Create all of the cohort definitions 
+            // and map the IDs from old -> new
+            List<BigDecimal> newTargetIds = new ArrayList<>();
+            List<BigDecimal> newOutcomeIds = new ArrayList<>();
+            AnalysisCohortDefinitionImportService cohortImportService = new AnalysisCohortDefinitionImportService(security, userRepository, cohortDefinitionRepository);
```
pavgra: Why do you use bean as POJO?

src/main/java/org/ohdsi/webapi/analysis/importer/AnalysisCohortDefinitionImportService.java
```diff
@@ -0,0 +1,33 @@
...
+    private final CohortDefinitionRepository cohortRepository;
+    private final AnalysisCohortDefinitionToCohortDefinitionConverter conversionService = new AnalysisCohortDefinitionToCohortDefinitionConverter();
+    
+    public AnalysisCohortDefinitionImportService(Security security, UserRepository userRepository, CohortDefinitionRepository cohortRepository) {
+        this.security = security;
+        this.userRepository = userRepository;
+        this.cohortRepository = cohortRepository;
+    }
+    
+    public CohortDefinition persistCohort(final AnalysisCohortDefinition analysisCohortDefinition) {
```
pavgra: 1) Such approach will create duplicates
2) Required method already exists - `DesignImportService.persistCohortOrGetExisting`

src/main/java/org/ohdsi/webapi/analysis/importer/AnalysisConceptSetImportService.java
```diff
@@ -0,0 +1,31 @@
...
+    private final ConceptSetService conceptSetService;
+    private final AnalysisConceptSetToConceptSetConverter csConverter = new AnalysisConceptSetToConceptSetConverter();
+    private final AnalysisConceptSetToConceptSetItemConverter csiConverter = new AnalysisConceptSetToConceptSetItemConverter();
+    
+    public AnalysisConceptSetImportService(ConceptSetService conceptSetService) {
+        this.conceptSetService = conceptSetService;
+    }
+    
+    public ConceptSetDTO persistConceptSet(final AnalysisConceptSet analysisConceptSet) {
+        ConceptSetDTO cs = csConverter.convert(analysisConceptSet);
```
pavgra: Will create duplicates. I suggest applying the same approach as used for cohorts in `
2\. DesignImportService.persistCohortOrGetExistin`g

src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
```diff
@@ -260,8 +262,109 @@ public EstimationAnalysisImpl exportAnalysis(Estimation est) {
...
+                CohortDefinition cd = cohortImportService.persistCohort(analysisCohortDefinition);
+                cohortIds.put(Long.valueOf(oldId), Long.valueOf(cd.getId()));
+                analysisCohortDefinition.setId(cd.getId());
+                log.debug("cohort created: " + cd.getId());
+            });
+            
+            // Create all of the concept sets and map
+            // the IDs from old -> new
+            Map<Integer, Integer> conceptSetIdMap = new HashMap<>();
+            AnalysisConceptSetImportService conceptSetImportService = new AnalysisConceptSetImportService(conceptSetService);
```
pavgra: Why do you use bean as POJO?

src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
```diff
@@ -187,9 +193,77 @@ public PatientLevelPredictionAnalysisImpl exportAnalysis(int id) {
...
+                if (analysis.getOutcomeIds().contains(oldId)) {
+                    newOutcomeIds.add(new BigDecimal(cd.getId()));
+                }
+                analysisCohortDefinition.setId(cd.getId());
+            });
+            
+            // Create all of the concept sets and map
+            // the IDs from old -> new
+            Map<Integer, Integer> conceptSetIdMap = new HashMap<>();
+            AnalysisConceptSetImportService conceptSetImportService = new AnalysisConceptSetImportService(conceptSetService);
```
pavgra: Why do you use bean as POJO?

src/main/java/org/ohdsi/webapi/analysis/importer/AnalysisCohortDefinitionImportService.java
anthonysena: Refactored to use `DesignImportService` and added an overloaded method for `persistCohortOrGetExisting` which also checks the cohort name along with the expression hash so that when cohorts are imported, they retain their name.

src/main/java/org/ohdsi/webapi/analysis/importer/AnalysisConceptSetImportService.java
anthonysena: So I stopped short of adding in functionality to check the concept set as this feels like it could be fairly complex unless we introduce a hash on the concept set as was done for cohort definition details. Otherwise, I was starting down the path of: identify the concept set by name, then filter by the items. But this requires a number of checks on the various flags for each concept set item, etc. For now, I think we may need to live with duplicate concept sets upon import of a PLE/PLP unless you have some ideas on how to do this without a lot of extra work at this point? This may need to be addressed in 2.8.

anthonysena: The other thought I had here: I could move the code from `org.ohdsi.webapi.analysis.importer.AnalysisConceptSetImportService` and make it part of the `DesignImportService`. This way the design import functions are in a single place and it eliminates the need for a single class in a single namespace (since I was able to remove the `AnalysisCohortDefinitionImportService`)

src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
```diff
@@ -75,6 +80,8 @@
             "createdBy",
             "modifiedBy"
     );
+    
+    private final AnalysisCohortDefinitionToCohortDefinitionConverter cohortConversionService = new AnalysisCohortDefinitionToCohortDefinitionConverter();
```
pavgra: Why do you instantiate converter manually? Use `conversionService` instead

src/main/java/org/ohdsi/webapi/common/DesignImportService.java
```diff
@@ -46,12 +51,21 @@ public CohortDefinition persistCohortOrGetExisting(final CohortDefinition cohort
...
-
         List<CohortDefinitionDetails> detailsFromDb = detailsRepository.findByHashCode(details.calculateHashCode());
         return detailsFromDb
                 .stream()
                 .filter(v -> Objects.equals(v.getStandardizedExpression(), details.getStandardizedExpression()))
                 .findFirst()
                 .map(CohortDefinitionDetails::getCohortDefinition);
     }
+    
+    private Optional<CohortDefinition> findCohortByExpressionHashcodeAndName(final CohortDefinitionDetails details, final String cohortName) {
```
pavgra: I would not copy implementation: I would create `findCohortByExpressionHashcodeAndPredicate` which accepts `CohortDefinitionDetails details` and predicate (lambda) as inputs. The predicate is passed into the `filter`. And two methods override it: `findCohortByExpressionHashcode` which passes `c -> true` as predicate and `findCohortByExpressionHashcodeAndName` which passes `c -> Objects.equals(c.getName(), cohortName)` (use Objects.equals to be null safe)

src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
```diff
@@ -54,6 +61,8 @@
             "createdBy",
             "modifiedBy"
     );
+    
+    private final AnalysisCohortDefinitionToCohortDefinitionConverter cohortConversionService = new AnalysisCohortDefinitionToCohortDefinitionConverter();
```
pavgra: Why do you instantiate converter manually? Use `conversionService` instead

src/main/java/org/ohdsi/webapi/analysis/importer/AnalysisConceptSetImportService.java
pavgra: > I could move the code from `org.ohdsi.webapi.analysis.importer.AnalysisConceptSetImportService` and make it part of the `DesignImportService`

Yes, I would do this

anthonysena:
1. Removed `org.ohdsi.webapi.analysis.importer.AnalysisConceptSetImportService` and made it part of the `DesignImportService`
2. Refactored to use `conversionService`
3. Refactored to use `Predicate<CohortDefinition>`

Let me know if you have any other feedback. Thanks!

anthonysena:
> Why do you have private final AnalysisCohortDefinitionToCohortDefinitionConverter cohortConversionService = new AnalysisCohortDefinitionToCohortDefinitionConverter();? - what's wrong with converting data via conversionService?

Oversight on my part; pushed a commit that removes this from the code.

> Just for my curiosity, could you explain why do you introduce AnalysisCohortDefinition? - why not utilize existing CohortDTO / CohortDefinition. And what's the reason for separate class for AnalysisConceptSet (doesn't seem that it introduces any new functionality or fields)?

I'm struggling to recall why I did this but I believe I needed to use the `@JsonIgnoreProperties(ignoreUnknown=true)` for these classes to get things working. I can revisit this for 2.8 since I suspect we can drop these extraneous classes or figure out a way to handle the root of the problem without the need for creating more classes. Let me know if you have any concerns with this approach.

src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
```diff
@@ -260,8 +265,109 @@ public EstimationAnalysisImpl exportAnalysis(Estimation est) {
...
+            analysis.getPositiveControlSynthesisArgs().getCovariateSettings().setIncludedCovariateConceptIds(new ArrayList<>());
+            analysis.getPositiveControlSynthesisArgs().getCovariateSettings().setExcludedCovariateConceptIds(new ArrayList<>());
+            
+            // Remove the ID
+            analysis.setId(null);
+            
+            // Create the estimation 
+            Estimation est = new Estimation();
+            est.setDescription(analysis.getDescription());
+            est.setName(String.format(ENTITY_COPY_PREFIX, analysis.getName()));
```
pavgra: Potential collision: I would use the same approach as applied for Pathways (`PathwayController.copy`)

src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
```diff
@@ -187,9 +199,76 @@ public PatientLevelPredictionAnalysisImpl exportAnalysis(int id) {
...
+                cs.setExcludedCovariateConceptIds(new ArrayList<>());
+            });
+            
+            // Remove the ID
+            analysis.setId(null);
+            
+            // Create the prediction analysis
+            PredictionAnalysis pa = new PredictionAnalysis();
+            pa.setDescription(analysis.getDescription());
+            pa.setName(String.format(ENTITY_COPY_PREFIX, analysis.getName()));
```
pavgra: Potential collision: I would use the same approach as applied for Pathways (`PathwayController.copy`)

PR #873, opened 2019-02-26
Patching runPathwayAnalysis.sql for Impala compatibility

AlexTawse:
- `FIRST` is a reserved word in Impala SQL so it cannot be used as an alias.  I have replaced this.
- I have added `CAST` statements where Impala is not able to coerce types automatically due to risk of precision loss.

I have performed cursory manual testing to show that the changes to this template do indeed work with Impala though have not had the time to check that they also work with PostgreSQL.  I will aim to do this ASAP.

No formal tests exist for this.

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -92,37 +92,37 @@ IF OBJECT_ID('tempdb..#split_overlay_events', 'U') IS NOT NULL
 DROP TABLE #split_overlay_events;
 
 SELECT
-	CASE WHEN ordinal < 3 THEN first.id ELSE second.id END as id,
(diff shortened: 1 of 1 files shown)


src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -92,37 +92,37 @@ IF OBJECT_ID('tempdb..#split_overlay_events', 'U') IS NOT NULL
...
-		first.cohort_end_date
+		f.cohort_end_date
 		WHEN 3 THEN
-		first.cohort_end_date
+		f.cohort_end_date
 		WHEN 4 THEN
-		second.cohort_end_date
-		END as cohort_end_date
+		s.cohort_end_date
+		END as cohort_end_datel
```
pavgra: Why this alias was changed (and changed only here)?

pavgra:
Fixes were required: closed by https://github.com/OHDSI/WebAPI/pull/903

Issue #851, opened 2019-02-21
CC design hash stored in generations is improperly calculated / doesn't match the current CC design


pavgra:
All highlighted executions were kicked off for the same design (so there were no updates to the CC between executions):
![image](https://user-images.githubusercontent.com/10010071/53203098-3477bc80-35f6-11e9-92a1-2a883bead0b4.png)

chrisknoll:
Could this be the 'missing cohort definitions' issue we've been trying to track down? Are you able to visually identify what the differences is between the designs between the two executions (by clicking onthe design hash and looking at the JSON?) I suspect that when you do this, one of them will have an empty `cohorts: []` array.

pavgra:
The difference is in elements of arrays changing order:
![image](https://user-images.githubusercontent.com/10010071/53205169-88d16b00-35fb-11e9-97a4-9f1676c90f31.png)
![image](https://user-images.githubusercontent.com/10010071/53205185-971f8700-35fb-11e9-9896-8942645c2852.png)

PR #874, opened 2019-02-26
Excluded 'updatedAt' from serialization and ordered sets inside CC

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/851


pavgra:
Although the designs now are identical, the hashes do differ. E.g.:

- Run analysis
- Change name to new one
- Run
- Change name to the original one
- Run

Hashes of the first and third differ.

PR #885, opened 2019-03-01
Fixed saving of CC params. Fixed copying of CC

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/881


aklochkova:
if I copy CC with 2 stratas ([stratas.txt](https://github.com/OHDSI/WebAPI/files/2918998/stratas.txt) ) only one of them appears in copied version

PR #887, opened 2019-03-01
Fixed null comparison

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/851


src/main/java/org/ohdsi/webapi/cohortcharacterization/domain/CcStrataEntity.java
```diff
@@ -78,6 +80,6 @@ public void setCohortCharacterization(CohortCharacterizationEntity cohortCharact
 
   @Override
   public int compareTo(CcStrataEntity o) {
-    return this.id.compareTo(o.id);
+    return ObjectUtils.compare(this.name, o.name);
```
wivern: Why entity is compared by name but DTO by id? Probably we should compare the combination of id, name and design?

src/main/java/org/ohdsi/webapi/cohortcharacterization/domain/CcParamEntity.java
```diff
@@ -86,6 +88,6 @@ public int hashCode() {
 
     @Override
     public int compareTo(CcParamEntity o) {
-        return this.id.compareTo(o.id);
+        return ObjectUtils.compare(this.name, o.name);
```
wivern: It is not clear to me why compareTo was added but hashCode and equals was not?

src/main/java/org/ohdsi/webapi/cohortcharacterization/converter/BaseCcDTOToCcEntityConverter.java
```diff
@@ -34,16 +35,21 @@ public CohortCharacterizationEntity convert(T source) {
 
     cohortCharacterization.setId(source.getId());
 
-    final Set<CohortDefinition> convertedCohorts = new TreeSet<>(converterUtils.convertSet(source.getCohorts(), CohortDefinition.class));
+    Set<CohortDefinition> convertedCohorts = new TreeSet<>((o1, o2) -> ObjectUtils.compare(o1.getId(), o2.getId()));
```
wivern: code duplication

PR #888, opened 2019-03-01
created job for warming source cache on user request(726)

### LINKED ISSUE #726: refresh record counts on source refresh
Can we include reloading the concept record count caches on source refresh to avoid having to do a WebAPI restart when we update sources?

ssuvorov-fls:
fixes #726 
see also https://github.com/OHDSI/Atlas/pull/1407


chrisknoll:
I would ask that we consider #892 first.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -168,15 +177,25 @@ public JobExecutionResource warmCache(@PathParam("sourceKey") final String sourc
...
             return new JobExecutionResource();
         }
     }
 
+    @GET
+    @Path("{sourceKey}/refreshCache")
+    @Produces(MediaType.APPLICATION_JSON)
+    public JobExecutionResource refreshCache(@PathParam("sourceKey") final String sourceKey) {
+        Source source = getSourceRepository().findBySourceKey(sourceKey);
+        if (sourceAccessor.hasAccess(source) && jobService.findJobByName(Constants.WARM_CACHE, getWarmCacheJobName(sourceKey)) == null) {
```
pavgra: If results of `jobService.findJobByName(Constants.WARM_CACHE, getWarmCacheJobName(sourceKey))` are not null, I would let the user know it (so return another response and show it on UI)

ssuvorov-fls: fixed - when job already exists services returns the execution resource of this job and don't create new job

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -168,15 +177,30 @@ public JobExecutionResource warmCache(@PathParam("sourceKey") final String sourc
...
         } else {
             return new JobExecutionResource();
         }
     }
 
+    @GET
+    @Path("{sourceKey}/refreshCache")
+    @Produces(MediaType.APPLICATION_JSON)
+    public JobExecutionResource refreshCache(@PathParam("sourceKey") final String sourceKey) {
+        Source source = getSourceRepository().findBySourceKey(sourceKey);
```
pavgra: The method should be available only when app is in secured mode and only for Admin role

PR #892, opened 2019-03-01
Only warm cache for sources with priority > 0.

chrisknoll:


pavgra:
Wouldn't this be shorter and easier to read?
```
sourceService.getSources()
            .stream()
            .filter(s -> s.daimons.stream().anyMatch(sd -> Objects.equals(sd.getDaimonType(), SourceDaimon.DaimonType.Results)) && s.daimons.stream().anyMatch(sd -> sd.getPriority() > 0))
            .forEach(s -> warmCache(s.sourceKey));
```

Issue #894, opened 2019-03-02
WebAPI recreates all linked entities when a user updates CC


pavgra:
When a user updates just name of a CC, the following occurs in logs:
```
Hibernate: delete from public.cc_cohort where cohort_characterization_id=?
Hibernate: delete from public.cc_analysis where cohort_characterization_id=?
Hibernate: insert into public.cc_cohort (cohort_characterization_id, cohort_id) values (?, ?)
Hibernate: insert into public.cc_analysis (cohort_characterization_id, fe_analysis_id) values (?, ?)
Hibernate: insert into public.cc_analysis (cohort_characterization_id, fe_analysis_id) values (?, ?)
Hibernate: insert into public.cc_analysis (cohort_characterization_id, fe_analysis_id) values (?, ?)
```

PR #895, opened 2019-03-02
Fixed update of CC

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/894


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -285,6 +275,7 @@ private void deleteParams(final CohortCharacterizationEntity entity, final Cohor
                 .stream()
                 .filter(parameter -> !nameToParamFromInputMap.containsKey(parameter.getName()))
                 .collect(Collectors.toList());
+        foundEntity.getParameters().removeAll(paramsForDelete);
```
wivern: Why not to apply the same method `getLinksToDelete` here?

Issue #893, opened 2019-03-01
migration #718 added conflicting pk


limadm:
Hi,
`V2.6.0.20181128150100__add_missing_pk.sql` in #718 added some primary keys, such as `ALTER TABLE ${ohdsiSchema}.cohort_inclusion ADD PRIMARY KEY (cohort_definition_id);`.

After the update, starting WebAPI gives `PSQLException: ERROR: could not create unique index "cohort_inclusion_pkey"` because my `cohort_inclusion` table is something like:
```
cohort_definition_id | rule_sequence | name        | description
                   2 |             0 | middle-aged | [NULL]
                   2 |             1 | male        | [NULL]
                   3 |             0 | Dx          | [NULL]
                   3 |             1 | 18+         | [NULL]
                   3 |             2 | M/F         | [NULL]
```
To me it seems that `(cohort_definition_id, rule_sequence)` should be the primary key of cohort_inclusion, as that migration does not work on a populated database.

chrisknoll:
You are absolutely right, that table should never had a PK assigned to it.   The origional issue was talking about cohort characterization tables PK, and that was pushed over to these tables.

@pavgra, @aklochkova , in this case, what we may want to do is alter this migration script to remove the PK on cohort_inclusion, and in a seperate migration remove the PK if it exists, or alternatively, put something in release notes to tell people to drop the PK on this table.  I don't like conditionals in our migration scripts, as you know, but in this case, for people that have data in their tables, this migration will fail, and for those who don't they will succeed....and hence for 2.7, we don't know if we need to clean up the PK or not.

chrisknoll:
@limadm , when reviewing this further, it looks like you are using the 'example tables' that are installed into the webAPI schema instead of using the DDL service (/WebAPI/ddl/results) to create the results schema.

The example tables (that I describe here:  https://github.com/OHDSI/WebAPI/pull/902) where never intended to be used as the results schema, and when the devs attempted to create PK constraints on these tables, it introduced problems for people that were writing results to the webAPI schema.

So, the #902 pr is going to try to undo all the problems here, but for your environment, I **strongly** suggest you separate your CDM results schema from your WebAPI schema by setting up a different database instance, and in that separate database instance, you store your CDM/Vocabulary/Results tables.  In a future release of WebAPI we will be _automatically_ dropping those example tables so we do now want to see anything break in your environment, so please migrate those tables out of WebAPI.

chrisknoll:
[...]

PR #902, opened 2019-03-04
Changed pk definition for cohort_inclusion

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/893


pavgra:
@chrisknoll , today Inclusion Name is nullable and UI allows to persist empty Inclusion Names. But wouldn't it be better to make the Inclusion Name field non nullable and transfer PK from (cohort_definition_id, rule_sequence) to (cohort_definition_id, name)?

chrisknoll:
Few problems with that idea:
- Since we allowed empty rule names into the db, adding a PK later would lead to the same sort of non-unique key.
- I don't really like the idea  of a PK field being avarchar that the user can update at any time.
- I don't really see a need to declare a PK on this table since there's no FK that needs to be associated with those rows.  However, a FK back to cohort definition_id in cohort_definition on the cohort_inclusion cohort_definition_id (cohort_definition_id is a FK in the cohort_inclusion table) would make sense.

What problem are we trying to solve here, exactly?

chrisknoll:
And, I need to take a step back for a minute and describe what the cohort_inclusion table is:

The cohort_inclusion table is a legacy table before we had the results_schema DDL service installed, where we would create tables in the WebAPI database schema to mirror tables that would need to be created in the results schema.  This was one of the worst ideas I had, and I take full responsibility for the confusion.  We have documentation describing which tables should be copied over to the results schema, as you can see here: http://www.ohdsi.org/web/wiki/doku.php?id=documentation:software:webapi:multiple_datasets_configuration

This is the list:

cohort | heracles_analysis
-- | --
COHORT_FEATURES | feas_study_result
COHORT_FEATURES_DIST | feas_study_index_stats
COHORT_FEATURES_REF | heracles_heel_results
COHORT_FEATURES_ANALYSIS_REF | heracles_results
cohort_inclusion | heracles_results_dist
cohort_inclusion_result | ir_analysis_dist
cohort_inclusion_stats | ir_analysis_strata_stats
cohort_summary_stats | ir_analysis_result
feas_study_inclusion_stats | ir_strata

Basically, if you have a question int he WebAPI schema about any of these tables, just ignore them.  When we get to WebAPI 3.0, we're going to remove all of these 'result schema example tables' out of the WebAPI schema and clean up the WebAPI schema to only contain tables that are needed by webapi, and the ddl service will be the only source of the results schema DDL.

aklochkova:
@chrisknoll So it is ok to have duplicate rows in cohort_inclusion table?

chrisknoll:
No, not saying that. cohort_id and rule_id will be unique.  But this table exists on the results schema, which means it's hosted in things like Google Big Query, Impala, Hadoop, etc, I think support for constraints like primary keys or unique constraints is limited.

But, my point is with this PR is that you should ignore any table in the list of tables in my previous comments.  They are not used by WebAPI, they are just example tables for a results schema; But that is an obsolete idea now that we have the results schema. So please don't try to hook up foreign keys or unique indexes on any of those tables, it's a waste of time.

chrisknoll:
@aklochkova , @limadm, Some of these changes proposed is going to re-write published migrations, which is something we want to avoid. @limadm, I'm going to propose a specific fix for your issue that will move your WebAPI tables out to a results schema and then you will truncate the tables in WebAPI so that the migration can proceed.  In future releases, we will be dropping these tables to avoid this confusion.  But for this PR, we'll close it out.

PR #909, opened 2019-03-05
Fixes IR allows running even if previous execution is not completed yet

### LINKED ISSUE #898: IR allows running even if previous execution is not completed yet

wivern:
Fixes #898


pavgra:
@wivern , let's reject requests for generation if there is one already running for a given ID and Source

Issue #626, opened 2018-10-04
Ensure only non-snapsot references are in POM.xml on master


chrisknoll:
There is a delay in bulding webAPI due to maven attempting to resolve certain resources from snapshot. This is a log of the build:
```
------------------------------------------------------------------------
Building WebAPI 1.0.0-SNAPSHOT
------------------------------------------------------------------------
Downloading: http://repo.ohdsi.org:8085/nexus/content/repositories/snapshots/commons-io/commons-io/maven-metadata.xml
Downloading: http://repo.maven.apache.org/maven2/commons-io/commons-io/maven-metadata.xml
Downloading: http://repo.ohdsi.org:8085/nexus/content/repositories/thirdparty/commons-io/commons-io/maven-metadata.xml
Downloading: http://repo.ohdsi.org:8085/nexus/content/repositories/releases/commons-io/commons-io/maven-metadata.xml
            
            
            
Downloading: http://repo.springsource.org/release/commons-io/commons-io/maven-metadata.xml
Downloading: http://redshift-maven-repository.s3-website-us-east-1.amazonaws.com/release/commons-io/commons-io/maven-metadata.xml
Downloading: http://repo.odysseusinc.com/artifactory/community-libs-snapshot-local/commons-io/commons-io/maven-metadata.xml
            
Downloaded: http://repo.maven.apache.org/maven2/commons-io/commons-io/maven-metadata.xml (783 B at 0.1 KB/sec)
Downloading: http://repo.odysseusinc.com/artifactory/community-libs-release-local/commons-io/commons-io/maven-metadata.xml
            
            
Downloading: https://oss.sonatype.org/content/repositories/snapshots/commons-io/commons-io/maven-metadata.xml
            
            
            
Downloading: http://repo.ohdsi.org:8085/nexus/content/repositories/releases/com/odysseusinc/arachne/execution-engine-commons/1.14.0-SNAPSHOT/maven-metadata.xml
Downloading: http://repo.ohdsi.org:8085/nexus/content/repositories/thirdparty/com/odysseusinc/arachne/execution-engine-commons/1.14.0-SNAPSHOT/maven-metadata.xml
Downloading: http://repo.maven.apache.org/maven2/com/odysseusinc/arachne/execution-engine-commons/1.14.0-SNAPSHOT/maven-metadata.xml
[...]

PR #910, opened 2019-03-05
Update version to 2.7.0

wivern:


chrisknoll:
With this PR, let's get the other POM.xml updates for released versions of external deps:
- circe
- sqlrender
- standard analysis API
- Any other ody artifacts that are currently in SNAPSHOT.

anthonysena:
Agreed with @chrisknoll - linking this to #626 to ensure we don't forget that as well. Also, I'm going to change the target branch for this PR to the new 2.7 release candidate branch: v2.7.0-rc.

Issue #841, opened 2019-02-18
cohort (treatment) pathways execution fails on Redshift


leeevans:
Atlas/WebAPI cohort (treatment) pathways generation fails on Redshift with this error:

`java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.springframework.jdbc.BadSqlGrammarException: StatementCallback; bad SQL grammar [CREATE TABLE #split_overlapping_events DISTKEY(ELSE) AS SELECT CASE WHEN ordinal < 3 THEN first.id ELSE second.id END as id, CASE WHEN ordinal < 3 THEN first.event_cohort_index ELSE second.event_cohort_index END event_cohort_index, CASE WHEN ordinal < 3 THEN first. subject_id ELSE second.subject_id END subject_id, CASE ordinal WHEN 1 THEN first.cohort_start_date WHEN 2 THEN second.cohort_start_date WHEN 3 THEN second.cohort_start_date WHEN 4 THEN first.cohort_end_date END as cohort_start_date, CASE ordinal WHEN 1 THEN second.cohort_start_date WHEN 2 THEN first.cohort_end_date WHEN 3 THEN first.cohort_end_date WHEN 4 THEN second.cohort_end_date END as cohort_end_date FROM #collapsed_dates_events first JOIN #collapsed_dates_events second ON first.subject_id = second.subject_id AND first.cohort_start_date < second.cohort_start_date AND first.cohort_end_date < second.cohort_end_date AND first.cohort_end_date > second.cohort_start_date CROSS JOIN (SELECT 1 ordinal UNION SELECT 2 UNION SELECT 3 UNION SELECT 4) multiplier]; nested exception is java.sql.SQLException: [Amazon](500310) Invalid operation: syntax error at or near "ELSE" Position: 49; at org.ohdsi.webapi.common.generation.AnalysisTasklet.waitForFuture(AnalysisTasklet.java:62) at org.ohdsi.webapi.pathway.GeneratePathwayAnalysisTasklet.doTask(GeneratePathwayAnalysisTasklet.java:74) at `

The sqlrendered translation of the below SQL file from OHDISQL to Redshift SQL is producing the invalid DISTKEY(ELSE) clause in the target SQL. The DISTKEY translation seems to be confused by the 'SELECT CASE WHEN ELSE clause'.

https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/pathway/runPathwayAnalysis.sql

[...]

PR #912, opened 2019-03-05
Fixes to avoid erroneous translation of Pathways SQL for Redshift

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/841

Tested on Postgres, Impala & Oracle

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -91,47 +91,42 @@ into
 IF OBJECT_ID('tempdb..#split_overlay_events', 'U') IS NOT NULL
 DROP TABLE #split_overlay_events;
 
-CREATE TABLE #split_overlay_events (
-  id BIGINT,
-  event_cohort_index INT,
-  subject_id BIGINT,
-  cohort_start_date DATETIME,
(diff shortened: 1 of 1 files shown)


chrisknoll:
The approach here is a Create Table followed by an Insert. However, this tends to perform very poorly on MPP systems, where a CREATE TABLE AS SELECT is much more performant (usually by a factor of 100). 

The initial proposal was to wrap the query in a sub-query so that SqlRender would parse the columns properly.  Did this approach not work?

pavgra:
PR was updated:
https://github.com/OHDSI/WebAPI/issues/841#issuecomment-469793604

src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -91,39 +91,42 @@ into
...
-		WHEN 1 THEN
-		s.cohort_start_date
-		WHEN 2 THEN
-		f.cohort_end_date
-		WHEN 3 THEN
-		f.cohort_end_date
-		WHEN 4 THEN
-		s.cohort_end_date
-		END as cohort_end_date
+SELECT soe.*
```
anthonysena: Per review, we should explicitly list the columns from the sub-query in this SELECT statement.

PR #923, opened 2019-03-06
Fixes potential NPE in toJobExecutionResource (into master)

pavgra:


src/main/java/org/ohdsi/webapi/job/JobUtils.java
```diff
@@ -116,7 +116,7 @@ public static JobExecutionResource toJobExecutionResource(final JobExecution job
             map.put(rs.getString(12), value.getValue());//value);
             
         }
-        if (jobexec.getExecutionId() != null) {
+        if (jobexec != null && jobexec.getExecutionId() != null) {
```
chrisknoll: What is this null check required here, when on line 76, it performs the jobexec check, and inside the IF it reassigns the jobexec to `jobexec = toJobExecutionResource(jobExecution)`?  Could the `toJobExecutionResource(jobExecution)` result in a null? Is it a bug if `toJobExecutionResource` returns a null?

pavgra: You will reach line 76 only if `rs.next()` has something. But it is not guaranteed:

image

PR #928, opened 2019-03-07
Adds concept code to cc results

wivern:
Closes OHDSI/Atlas#1427

### DIFF
--- a/src/main/resources/resources/cohortcharacterizations/sql/queryResults.sql
+++ b/src/main/resources/resources/cohortcharacterizations/sql/queryResults.sql
@@ -16,7 +16,6 @@ select
        r.covariate_id,
        r.covariate_name,
        c.concept_name,
-       c.concept_code,
        r.time_window,
        r.concept_id,
        r.count_value,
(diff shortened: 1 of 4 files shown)


src/main/resources/resources/cohortcharacterizations/sql/queryResults.sql
```diff
@@ -32,4 +34,5 @@ select
        r.strata_name
 from @results_database_schema.cc_results r
   JOIN threshold_passed_ids tpi ON tpi.covariate_id = r.covariate_id
+  LEFT JOIN @vocabulary_schema.concept c on c.concept_id = r.concept_id
```
anthonysena: Can we use an inner join here? Looking at the data in our environment, it appears that the `concept_id` is always >=0 and never null which should allow for an inner join without data loss.

pavgra:
@anthonysena , I haven't received confirmation from @pbr6cornell that in fact there was no duplication of rows and it was just a matter of different concepts, that's why I haven't merged any of these PRs yet. Also, I wanted to check with @pbr6cornell whether it is more useful to show concept code or concept id

anthonysena:
Left a comment on OHDSI/Atlas#1427 but will also state here: from my investigation, I did not observe any duplication of data on the back-end. The `concept_id` for each covariate should be unique (the covariate suffix may differ) so hoping that simply displaying the concept_id in addition to the name would be a solution and I think might avoid a change to WebAPI. I'd appreciate your thoughts on this @pavgra while we await feedback from @pbr6cornell.

PR #929, opened 2019-03-07
Adds concept code to cc results

wivern:
Closes OHDSI/Atlas#1427

### DIFF
--- a/src/main/resources/ddl/results/cohort_characterizations.sql
+++ b/src/main/resources/ddl/results/cohort_characterizations.sql
@@ -11,7 +11,7 @@ CREATE TABLE @results_schema.cc_results
   strata_id BIGINT,
   strata_name VARCHAR(1000),
   time_window VARCHAR(255),
-  concept_id INTEGER,
+  concept_id INTEGER NOT NULL,
   count_value BIGINT,
   avg_value DOUBLE PRECISION,
   stdev_value DOUBLE PRECISION,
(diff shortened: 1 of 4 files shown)


src/main/resources/ddl/results/cohort_characterizations.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE @results_schema.cc_results
   strata_id BIGINT,
   strata_name VARCHAR(1000),
   time_window VARCHAR(255),
-  concept_id INTEGER,
+  concept_id INTEGER NOT NULL,
```
chrisknoll: not all statistics have an associated CONCEPT_ID, are you sure this is a NOT NULL column?

pavgra: If there is no matching concept id, it will be CONCEPT_ID = 0 (I believe that's the case with all CDM tables)

chrisknoll: Ok, to confirm then: CriteriaFeatures will have Concept_ID = 0?

pavgra: From `GenerateCohortCharacterizationTasklet.getCriteriaFeatureQuery`:
```
 private List<PreparedStatementCreator> getCriteriaFeatureQuery(Integer cohortDefinitionId, FeAnalysisWithCriteriaEntity analysis, FeAnalysisCriteriaEntity feature, String targetTable, CcStrataEntity strata) {

            Long conceptId = 0L;
            ...
```

anthonysena:
Can we close this one in favor of #928? I've approved that other one for the v2.7.0-rc branch. Thanks!

pavgra:
The fix will be merged back from RC to master when RC is finished

PR #960, opened 2019-03-13
Splits runHeracles.sql into parts and assemble them in Java

### LINKED ISSUE #385: Split runHeracles.sql into parts and assemble them in Java
Today we have a single enormously huge SQL file containing all analyses. The file is processed by SqlRender to cut out all not required analyses, which is absolutely inefficient and takes up to several minutes. What is more, managing and editing of such file is also not convenient.

The proposal is to split the file into partials, one per analysis or analyses group, and assemble them in Java. And only then pass to SqlRender. This would give us a big boost in performance.

wivern:
Closes #385


src/main/java/org/ohdsi/webapi/cohortanalysis/HeraclesQueryBuilder.java
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  private String[] buildAnalysisParams(CohortAnalysisTask task) {
+
+    String resultsTableQualifier = task.getSource().getTableQualifier(SourceDaimon.DaimonType.Results);
+    String cdmTableQualifier = task.getSource().getTableQualifier(SourceDaimon.DaimonType.CDM);
+
+    String cohortDefinitionIds = (task.getCohortDefinitionIds() == null ? "" : Joiner.on(",").join(
+            task.getCohortDefinitionIds()));
+    String analysisIds = (task.getAnalysisIds() == null ? "" : Joiner.on(",").join(task.getAnalysisIds()));
```
pavgra: Why not move the construction to shared helper?
`task.getAnalysisIds() == null ? "" : Joiner.on(",").join(task.getAnalysisIds())`

src/main/java/org/ohdsi/webapi/cohortanalysis/HeraclesQueryBuilder.java
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  public String buildHeraclesAnalysisQuery(CohortAnalysisTask task) {
+
+    return new Builder(task).buildQuery();
+  }
+
+  private String[] buildAnalysisParams(CohortAnalysisTask task) {
+
+    String resultsTableQualifier = task.getSource().getTableQualifier(SourceDaimon.DaimonType.Results);
```
pavgra: You have a mix of `getTableQualifier(SourceDaimon.DaimonType.*` and `SourceUtils`. I would propose to be consistent and go for `SourceUtils` only

pavgra:
@wivern , new Full Analysis doesn't produce the same results:
image

Looking into SQL for analysis id = 1101 I see the following difference.

Old version:
image

New version:
image

wivern:
@pavgra, My bad, missed that field in condition.

pavgra:
@wivern , the following query stucks for me third time so I cannot verify results:
```
INSERT INTO results.HERACLES_HEEL_results (
  cohort_definition_id,
  analysis_id,
  HERACLES_HEEL_warning
)
  SELECT or1.cohort_definition_id, or1.analysis_id,
    CAST(CONCAT('ERROR: ', cast(or1.analysis_id as VARCHAR), '-', oa1.analysis_name, '; ', cast(COUNT(DISTINCT stratum_1) AS VARCHAR), ' concepts in data are not in vocabulary') AS VARCHAR(255)) AS HERACLES_HEEL_warning
  FROM results.heracles_results or1
    INNER JOIN results.heracles_analysis oa1
      ON or1.analysis_id = oa1.analysis_id
    LEFT JOIN public.concept c1
      ON or1.stratum_1 = CAST(c1.concept_id AS VARCHAR)
  WHERE or1.analysis_id IN (
    2,
    4,
    5,
    200,
    301,
    400,
    500,
    505,
    600,
    700,
    800,
    900,
    1000,
    1609,
    1610
  )
        AND or1.stratum_1 IS NOT NULL
        AND c1.concept_id IS NULL
        and cohort_definition_id in (214)
  GROUP BY or1.cohort_definition_id,
    or1.analysis_id,
    oa1.analysis_name
```

Using the following query, I do not observe any locks in DB:
```
select pid,
       usename,
       pg_blocking_pids(pid) as blocked_by,
       query as blocked_query
from pg_stat_activity
where cardinality(pg_blocking_pids(pid)) > 0;
```

There is also no load on the DB - this is the single query running

Issue #590, opened 2018-09-17
Missing db/migrations for attribute mode_id of ddl/results/cohort_*


limadm:
At v2.5.0 (2cda7393), `ddl/results/cohort_*` tables have a new attribute:
`mode_id INT NOT NULL DEFAULT 0;`
This atttribute is not present in `db/migration/*`, causing errors when generating cohorts.

chrisknoll:
db/migration does not touch the webAPI schema. You need to manage the results schema manually.

This change was noted in 2.5.0 release notes under the 'Migration' section:
https://github.com/OHDSI/WebAPI/releases

limadm:
Hello!
My bad, I git-pulled the repository and didn't see the release notes.
There is
`CREATE TABLE ${ohdsiSchema}.cohort_inclusion_result`
(and other result tables) in
`db/migration/postgresql/V1.0.2.0__cohort_feasiblity.sql`
so at first I thought that adding these `ALTER TABLES` in a subsequent migration would be reasonable to ease environment setup while they are still there.

chrisknoll:
> PS: I did a small change in the exception reporting in order to see what was causing this.
> In GenerationJobExecutionListener:85 instead of StringUtils.left(exceptionMessage, 2000), I used StringUtils.abbreviateMiddle(exceptionMessage, "... [truncated] ...", 2000).
> 

That would make an excellent pull request.

Re: the legacy tables in webAPI, yes, it's confusing, we'd like to drop it, but someone may actually be using those so we need to give clear warning before we do that (most likely for a 3.0 release)

PR #972, opened 2019-03-15
Added handling for case if source daimon doesn't exist

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/590

### DIFF
--- a/src/main/java/org/ohdsi/webapi/util/SourceUtils.java
+++ b/src/main/java/org/ohdsi/webapi/util/SourceUtils.java
@@ -4,25 +4,64 @@ import org.apache.commons.lang3.ObjectUtils;
 import org.ohdsi.webapi.source.Source;
 import org.ohdsi.webapi.source.SourceDaimon;
 
+import java.util.Map;
+
+import static org.ohdsi.webapi.Constants.Params.CDM_DATABASE_SCHEMA;
+import static org.ohdsi.webapi.Constants.Params.RESULTS_DATABASE_SCHEMA;
+import static org.ohdsi.webapi.Constants.Params.TARGET_DATABASE_SCHEMA;
+import static org.ohdsi.webapi.Constants.Params.TEMP_DATABASE_SCHEMA;
+import static org.ohdsi.webapi.source.SourceDaimon.DaimonType.CDM;
+import static org.ohdsi.webapi.source.SourceDaimon.DaimonType.Results;
+import static org.ohdsi.webapi.source.SourceDaimon.DaimonType.Temp;
+
 public class SourceUtils {
     public static String getVocabularyQualifier(Source source) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/ohdsi/webapi/util/SourceUtils.java
```diff
@@ -4,25 +4,64 @@
...
     public static String getResultsQualifier(Source source) {
-        return source.getTableQualifier(SourceDaimon.DaimonType.Results);
+        return source.getTableQualifier(Results);
+    }
+
+    public static String getResultsQualifierOrNull(Source source) {
+        return source.getTableQualifierOrNull(Results);
+    }
+
+    public static String getSchema(Map<String, Object> jobParams, String qualifier){
```
chrisknoll: This method seems out of place: it doesn't have anything to do with Sources, it's more of a JobParams helper to pull out a value from a jobs param.

At the end we have:
`
throw new NullPointerException(String.format("DaimonType (\"%s\") not found in Source", daimonName));
`

But, the source was never inspective for the DaimonType, so why would it say it wasn't found?

aklochkova: Source is inspected for daimons in `org.ohdsi.webapi.service.CohortGenerationService#getJobParametersBuilder`. And I wrote a comment there about nulls

chrisknoll: @aklochkova , thanks, I looked at the comment, i was confused by the comment (proceed with nulls because it will be added to database), but taking a step back:  in `CohortGenerationService.getJobParamatersBuilder`, when getting the table qualfiers, it seems there's special handling for the Temp qualifier: it's either going to be form the TEMP daimon, or RESULTS daimon.  So, wouldn't it be enough to say:
```
builder.addString(TEMP_DATABASE_SCHEMA, SourceUtils.getTempQualifier(source, SourceUtils.getResultsQualifier(source)))
```
This will try to get the actual TEMP qualfier from the source, and if that fails, it will return the Results qualifier from the source.

I feel like that's all that's needed here, and none of the `getSchema` code that was added to source utils.

aklochkova: @chrisknoll And what happens if Results qualifier is empty?
Now it fails with Runtime error which is caught by proxy instance and no job is created. If job isn't created this failure case isn't processed correctly in the fronted side (by 'correctly' I mean job status is updated to "FAILED" and job list is updated)
My logic is to allow creating job with empty table qualifiers but to throw error further

chrisknoll: Ok, thanks for the explanation. Initially, we restricted the jobs from being launched on sources that don't have the required Daimons, but you're correct that it is something that should be validated on the server in the event someone POSTs a request for a generation 'blindly'.

PR #974, opened 2019-03-15
Adds Support of BQ keyfile json for authentication of EE jobs

### LINKED ISSUE #966: Support of BQ keyfile json for authentication of EE jobs

wivern:
Closes #966


src/main/java/org/ohdsi/webapi/util/DataSourceDTOParser.java
```diff
@@ -26,6 +27,9 @@ public static DataSourceUnsecuredDTO parseDTO(Source source) {
         if (Objects.equals(IMPALA, dto.getType()) && AuthMethod.KERBEROS == params.getAuthMethod()) {
             KerberosUtils.setKerberosParams(source, params, dto);
         }
+        if (Objects.equals(BIGQUERY, dto.getType())) {
+            dto.setKrbKeytab(source.getKrbKeytab());
```
pavgra: Why not rename `setKrbKeytab` to smth more relevant? In case of BigQuery we don't have Kerberos at all so the setter naming is confusing. Maybe `setKeyfile`?

chrisknoll:

Issue #751, opened 2019-01-18
Update Basic Security Configuration in OHDSI Wiki so that it works for recent release(s)


t-abdul-basser:
### Expected Behavior

When attempting to configure a WebAPI deployment to support Atlas with security features, following the [Basic Security Configuration](http://www.ohdsi.org/web/wiki/doku.php?id=documentation:software:webapi:basic_security) tutorial in the OHDSI Wiki results in a successful deployment.

### Actual Behavior

As per #571, security is not configured in resulting deployment. Authentication fails with 
* error message in Atlas login form ("Bad Credentials") and 
* console errors raised for the `/login` service .

### Steps to Reproduce

1. Follow tutorial

t-abdul-basser:
Updated [Basic Security Configuration](http://www.ohdsi.org/web/wiki/doku.php?id=documentation:software:webapi:basic_security) so that 
    `- security.enabled=true`
 is replaced with
    `- security.provider=AtlasRegularSecurity`

anthonysena:
@t-abdul-basser - thanks again for making these edits to the WebAPI wiki. Recently I ported over the documentation to here: https://github.com/OHDSI/WebAPI/wiki/Basic-Security-Configuration and also expanded the [security documentation on the wiki](https://github.com/OHDSI/WebAPI/wiki/Security-Configuration). I think we should be set but please re-open if you see any issues in the documentation. Thanks!

Issue #686, opened 2018-10-31
WebAPI Wiki


anthonysena:
I'd like to propose that we move the content from our [current wiki](http://www.ohdsi.org/web/wiki/doku.php?id=documentation:software:webapi) to GitHub to keep it closer to the code. 

We also need help to update it to reflect the new options for configuration, setup, etc. Putting this out here to see who may be interested in helping with this effort.

konstjar:
@anthonysena I'm with you on this

anthonysena:
Per our Atlas/WebAPI call, we should also aim to have a list of SQL "templates" for flyway migrations for the various dialects.

anthonysena:
Closing this as it was done as part of the 2.7.2 release.

PR #980, opened 2019-03-19
Wiki and readme updates

anthonysena:
Aims to start work on updating the GitHub wiki per #686 and incorporates contributions from the OHDSI wiki from @t-abdul-basser per #751.

Some notes on this PR:

- I'm proposing to add a file "sample_settings.xml" that contains the 3 profiles for use when performing a maven build. I also thought of making a sample per DB dialect but thought having a single file would make subsequent updates easier.
- I've tried to add additional details for setting up & running WebAPI via Tomcat as well as the source/source_daimon configuration. Would appreciate a review and feedback to ensure that I haven't introduced any confusion on how this should be done. 
- Updates the README format to make it more consistent with the format of the OHDSI Methods Libraries.


anthonysena:
Thanks @t-abdul-basser! Appreciate your additions to the wiki - we're going to bring in this PR but we are doing it slightly differently since the wiki content is in its own git repo...stay tuned!

Issue #727, opened 2018-12-20
Improve init sql scripts


SBylemans:
I propose the initialization scripts (such as [init_concept_hierarchy.sql](https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/ddl/results/init_concept_hierarchy.sql)) make use of an indexed temp table and use this indexed temp table to perform the latest `LEFT JOIN`.

I recently ran into an issue where inserting the values for the Drug part of the concept_hierarchy took forever. After evaluation, it was clear that the latest `LEFT JOIN` in this process caused the issue (`... LEFT JOIN @vocab_schema.concept atc1 ON atc3_to_atc1.atc1_concept_id = atc1.concept_id;`). The selecting query resulted in +- 2.650.000 results and the concept table of the vocabulary schema had +- 5.200.000 results. The SQL code wouldn't even finish after running half a day.

The code below would run in under 10 minutes.

# Proposed solution for all init.sql files
```
CREATE TEMP TABLE drug_insert_temp (
  concept_id int,
  rxnorm_concept_name varchar(255),
  treemap varchar(255),
  rxnorm_ingredient_concept_name varchar(255),
  atc5_concept_name varchar(255),
  atc3_concept_name varchar(255)
);

INSERT INTO drug_insert_temp
SELECT
    rxnorm.concept_id,
    rxnorm.concept_name AS rxnorm_concept_name,
    'Drug'              AS treemap,
    rxnorm.rxnorm_ingredient_concept_name,
    atc5_to_atc3.atc5_concept_name,
    atc3_to_atc1.atc3_concept_name
  FROM (
         SELECT
           c1.concept_id,
           c1.concept_name,
           c2.concept_id   AS rxnorm_ingredient_concept_id,
           c2.concept_name AS RxNorm_ingredient_concept_name
         FROM vocabulary.concept c1
           INNER JOIN vocabulary.concept_ancestor ca1 ON c1.concept_id = ca1.descendant_concept_id
                                                         AND c1.domain_id = 'Drug'
           INNER JOIN vocabulary.concept c2 ON ca1.ancestor_concept_id = c2.concept_id
                                               AND c2.domain_id = 'Drug'
                                               AND c2.concept_class_id = 'Ingredient'
       ) rxnorm
    LEFT JOIN (
                SELECT
                  c1.concept_id      AS rxnorm_ingredient_concept_id,
                  max(c2.concept_id) AS atc5_concept_id
                FROM vocabulary.concept c1
[...]

PR #982, opened 2019-03-21
Introduced temp tables in init sql script

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/727

### DIFF
--- a/src/main/resources/ddl/results/init_concept_hierarchy.sql
+++ b/src/main/resources/ddl/results/init_concept_hierarchy.sql
@@ -84,6 +84,8 @@ SELECT
 FROM #condition_insert_temp LEFT JOIN @vocab_schema.concept soc ON condition_insert_temp.concept_id = soc.concept_id;
 
 truncate table #condition_insert_temp;
+
+IF OBJECT_ID('tempdb..#condition_insert_temp', 'U') IS NOT NULL
 drop table #condition_insert_temp;
 
 /********** DRUG **********/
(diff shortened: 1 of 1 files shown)


src/main/resources/ddl/results/init_concept_hierarchy.sql
```diff
@@ -71,19 +67,33 @@ LEFT JOIN (
...
+  condition_insert_temp.concept_name AS snomed_concept_name,
+	CAST('Condition' AS VARCHAR(20)) AS treemap,
+	condition_insert_temp.pt_concept_name,
+	condition_insert_temp.hlt_concept_name,
+	condition_insert_temp.hlgt_concept_name,
+	soc.concept_name AS soc_concept_name
+FROM #condition_insert_temp LEFT JOIN @vocab_schema.concept soc ON condition_insert_temp.concept_id = soc.concept_id;
+
+truncate table #condition_insert_temp;
+drop table #condition_insert_temp;
```
wivern: SqlRender recognizes table drop using this rule `IF OBJECT_ID('@table', 'U') IS NOT NULL DROP TABLE @table`
and w/o 'if object_id` illegal code may be generated.

wivern:
@aklochkova On which databases did you test init scripts?

aklochkova:
@wivern on postgresql

pavgra:
@aklochkova , could you provide performance comparison using classic relational DB (e.g. Postgres) and something like Impala / Redshift. I don't see performance improvement while observing the results of `EXPLAIN` in PG.

anthonysena:
Closing this out since we're unclear how these changes improve performance. I'm going to close this out. Thanks!

PR #988, opened 2019-03-26
Fixing R package installation at Windows

pavgra:


chrisknoll:
@wivern, please flush the toilet.

PR #1003, opened 2019-04-01
Fixes issue with wrong drivers path for PLE / PLP

pavgra:


src/main/resources/resources/estimation/r/runAnalysis.R
```diff
@@ -25,6 +25,8 @@ tryCatch({
         cohortTable <- Sys.getenv("COHORT_TARGET_TABLE")
         driversPath <- Sys.getenv("JDBC_DRIVER_PATH")
 
+        if (driversPath == "") driversPath = NULL
+
```
wivern: it could be `driversPath <- if (driversPath == "") NULL else driversPath`

chrisknoll:

PR #1005, opened 2019-04-01
Perfomance improvements for extracted CC analysis generation

### LINKED ISSUE #975: Extract Cohort Characterization analysis to external package
There is a need to run CC (and other new analyses types) via Arachne / R, therefore we would like to extract the code which builds SQL / runs the analyses into a common place. Earlier we discussed that it might be good to have a single analyses builder library and agreed that Hydra is a good candidate for this (@pbr6cornell , @chrisknoll, @anthonysena). Therefore, the proposed design for CC is:

1. Create a separate repo for CC analysis. Make it an R package
2. Extract code that generates SQL from `GenerateCohortCharacterizationTasklet` to a Java class placed into the R package
3. Include the package as dependency into WebAPI and replace the original implementation with a reference to the moved code
4. Create R stuff wrapping the Java class and submitting the generated SQL to DB + extracting generated results from DB

image

To guarantee reproducibility this is going to be accompaniated by a Packrat in Arachne:

image

wivern:
Closes #975


chrisknoll:

Issue #984, opened 2019-03-22
Pathways results take a long time to load


spbima:
Pathways results take a long time to load (approximately 1 minute)

![pathway-long-time](https://user-images.githubusercontent.com/804934/54821532-7c503900-4cb3-11e9-89ab-f15457db9b16.jpg)

[pathway-long-time.txt](https://github.com/OHDSI/WebAPI/files/2996360/pathway-long-time.txt)

chrisknoll:
Agreed. this is a 2.8 issue.  The core of the issue is that we are pulling patent-level pathways out of the database and procesisng pathways in memory. This needs to be summarized and stored in the DB and the report needs to just query a result table.

PR #1008, opened 2019-04-02
Moved collection of pathway generation results to its generation step

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/984


chrisknoll:
Hi, @aklochkova , you beat me to the punch on this one.

I've developed some optimizations for producing the path sequences on the database side. I believe this PR maintains that the patient-level event records are pulled into memory and the 'chains' are assembled there, summarized, and then written back into the database. 

I'd like to know if you'd be willing to accept those changes, of if you'd prefer we finish the work done here, and then revisit it with other performance updates.  I would rather not do double-work here, but I was too slow to get the ball rolling on this update.

Let me know, and if you'd like to work together on this PR, I'll post the ideas here, and possibly introduce a branch to your pr.

chrisknoll:
Hi, @aklochkova ,
  Just got off the call with Pavel/Anthony, and what I'd like to do is push up a branch based on your PR that just makes a few adjustments to the work you've already done.

What I'd like to leverage:
- You've removed the 'chain building' from the report results endpoint, and replaced that to a query to a results table. I want to keep that logic, just change the results table
- You've isolated the chain building into a tasklet, but I'd like to replace that with a server side aggregation to eliminate the DB-> java pull and Java-> db Push of the chains.
- you are storing the set of comboIDs in a results table so people can go from the comboID to the name of the cohorts in the combo. 

What I'm proposing to change:
- The results schema isn't going to be a 'varchar() chain' pathway. Rather, we're going to introduce a patway results table with generation_id, cohort_id, count and then 10 or 15 columns for each individual step.  This does introduce a new limitation on the pathways (it only stores up to 10 or 15 sequences (depending on what we decide, it's just more columns one way or the other), however, having each sequence in a dedicated column allows people to query the results directly and do their own aggregation, such as:
```
select generation_id, cohort_id, path_1, path_2, path_3, SUM(count_value) from pathway_results_summary where generation_id = X
```

And they can join the path_1/path_2/path_3 to the combo_lookup table to get the name of the combo for the generation.   Storing the paths as 3-16-4-21-5 is harder to 'consume' from our researchers perspective.

So, I'll put together a branch based on your latest PR with the proposed changes, and I hope we can collaborate about the final details that can make it into master and close this issue.

Please let me know if there's anything you'd like to discuss.

-Chris

aklochkova:
Hello, @chrisknoll 
I don't have any objections

chrisknoll:
Ok, this is ready for final review.  When this is merged, please squash-and-merge so that all this looks like 1 commit (I do not mind if @aklochkova has all the commit authorship here).  Thank you.

chrisknoll:
Java migration only deals with webAPI schema.  The results are in the CDM's results schema which we do not manage.

I was thinking about supplying some migration scripts, but the gap is creating the _code table requires to get the distinct code_ids in the _results, and then create the map.  However, it may be just simpler to instruct people to re-execute their studies.

I agree, giving a migration tool would be nice, I just am unsure if it would be somethign we have to bundle with webAPI or if we create a separate component that is a migration assistant.

anthonysena:
From my testing, these changes look good. 

In terms of migrating results: I think that including a note in the release notes to advise users to re-generate any 2.7.0 pathway analyses is the right balance here. They'll see a big benefit in terms of retrieving results which should be incentive to re-generate the analysis.

chrisknoll:
Ok, if either @wivern or @pavgra could submit their approval (if they agree with the re-generate approach for getting the results), that would be appreciated.  If we want to look at a more user-friendly approach to dealing with migration at the application level, we can open up a separate issue.

pavgra:
@chrisknoll ,

> Java migration only deals with webAPI schema

Why? What limits you to WebAPI schema only?

> dealing with migration at the application level, we can open up a separate issue.

This is the current PR which will invalidate all Pathway results once it gets merged - why separate issue?

chrisknoll:
> Why? What limits you to WebAPI schema only?
Good question. I'm assuming when you say 'Java Migration' you are talking about Flyway scripts.  Flyway migrations are bound to one specific DB schema, and the migration state is stored in a local table called 'schema_version'.  So, if we were to apply java migrations across multiple database schemas, we'd need to have a 'schema_version' type of table on each schema that we want to track.  

> This is the current PR which will invalidate all Pathway results once it gets merged - why separate issue?
In the past, when we've made updates to the results schema where the existing results may be invalidated or incompatable, just simply re-executing the analysis has been the recommendation.

Thinking on this further tho:  we keep histories of generations (and I don't think there's a way to clear them out via the UI). So, again, I'm not disagreeing with you that the function would be useful.  If you have team-cycles to spend on the problem, then that is great.  At the moment, we have people waiting upwards of 5 minutes to render a single report, so the interest is to get this performance fix done and if they have to wait for one additional generation so that we can save them 5 minutes per report load, I think they'd be fine with it.

If you are going to look into java migrations, you may want to also take a look at what I reproted here: https://github.com/flyway/flyway/issues/2319.  I wasn't able to get to the core of the issue, however, I did set a breakpoint on 2 separate flyway Beans: `flyway' and 'flywayInitializer` that are defined in org.ohdsi.webapi.FlywayConfig.  Oddly, the FlywayConfig bean fired the migrate, and then at a separate point in time, the breakpoint on the flywayInializer triggered. So, the 'double execution' I reported I think is a problem with how we have set up the Java configuration for Flyway.

chrisknoll:
@pavgra : I've had a change of heart, and I'd like to try to address this.  I can try, or if you could allocate one of your team to do this, that's fine with me, but I'd like this to be addressed quickly (people are really suffering the current performance):

We should do this via java migrations with the following limitations:
- The results tables will need to be created up front.  We can't expect the user/password specified in the source to have permissions to create tables (and it is recommended that they do not have create permissions on the results schema..only the temp schema).
- This is only going to migrate those results that can be found in the soruce/source_daimon tables _at the time of migration_.  Once the migration is marked complete, it won't re-run if they add a new source.

With these limitations in mind, this is how the migration script should work:

There should be 1 script (which I can provide) that can do a bulk insert into pathway_analysis_paths across all generations on a single source by reading the data in `pathway_analysis_events`.  We should _not_ try to re-use the analysis script for this, because the analysis script may get updated in the future, and the migration script has to have a _locked_ in implementation at the time of the migration.  the migration script can't change over time like the analysis scripts can.

The second table: pathway_analysis_codes can only be created by loading up the design for the associated generation (to get the names of the cohorts in the design).  Again, due to how migrations should not change over time, I think we need to resort to raw JSON parsing on the design table to get the cohort names (since the structure of the design may change over time, we need to lock in a parsing of the JSON _at the time of the migration_.  So, the second step of the migration is to loop through each generation for each analysis for each source, and execute the insert into pathway_analysis_codes for the distinct combo_ids that exist in each generation (which also reads from the pathway_analysis_events just like populating pathway_analysis_paths).

With these 2 steps complete, all the results should be migrated and no one will have any problems looking at previously-generated results.

Thoughts?

chrisknoll:
@pavgra , I've added migration scripts, as requested.

src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
```diff
@@ -0,0 +1,149 @@
...
+				int code = rs.getInt("combo_id");
+				List<PathwayEventCohort> eventCohorts = getEventCohortsByComboCode(design, code);
+				String names = eventCohorts.stream()
+								.map(PathwayEventCohort::getName)
+								.collect(Collectors.joining(","));
+				 return new PathwayCode(code, names, eventCohorts.size() > 1);
+			});
+    }
+		
+    private List<PathwayEventCohort> getEventCohortsByComboCode(PathwayAnalysisEntity pathwayAnalysis, int comboCode) {
```
pavgra: Why not to place the method into `PathwayService`? It doesn't contain any tasklet-specific stuff

pavgra:
@chrisknoll , this is what I get during migration:
```
PreparedStatementCallback; uncategorized SQLException for SQL [INSERT INTO synpuf_53_test_results.pathway_analysis_codes (pathway_analysis_generation_id, code, name, is_combo)
        VALUES (?,?,?,?);]; SQL state [HY000]; error code [500051]; [Simba][ImpalaJDBCDriver](500051) ERROR processing query/statement. Error Code: 0, SQL state: TStatus(statusCode:ERROR_STATUS, sqlState:HY000, errorMessage:AnalysisException: Target table 'synpuf_53_test_results.pathway_analysis_codes' is incompatible with source expressions.
Expression ''5543'' (type: STRING) is not compatible with column 'pathway_analysis_generation_id' (type: BIGINT)
), Query: INSERT INTO synpuf_53_test_results.pathway_analysis_codes (pathway_analysis_generation_id, code, name, is_combo)
        VALUES ('5543','2','Metformin','0');.
```

src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
chrisknoll: Because the implementation of PathwayService can change over time, and I need to lock in the implementation at the time the migration was written.

PR #1010, opened 2019-04-02
Fixes performance issues with PLE & PLP generations retrieval

pavgra:


chrisknoll:
@wivern,  please flush the toilet.

PR #1011, opened 2019-04-02
 Adds filter to do cache invalidation in case of threads re-usage

pavgra:


chrisknoll:
@wivern,  please flush the toilet.

PR #1015, opened 2019-04-03
Fixes SQL execution for BQ when keyfile is uploaded

pavgra:


chrisknoll:

PR #1023, opened 2019-04-05
Split EE result files storage into 2 tables and avoid contents retrieval on each request

pavgra:


src/main/resources/db/migration/sqlserver/V2.7.1.20190405124500__split_output_files.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ${ohdsiSchema}.output_file_contents (
+  output_file_id INTEGER,
+  file_contents VARBINARY(max),
+  CONSTRAINT output_file_contents_pkey PRIMARY KEY (output_file_id)
+)
+GO
+
+ALTER TABLE ${ohdsiSchema}.output_file_contents
+  ADD CONSTRAINT fk_ofc_of_id FOREIGN KEY (output_file_id)
+  REFERENCES ${ohdsiSchema}.output_files (id)
```
wivern: Message    : Column 'ohdsi.output_files.id' is not the same data type as referencing column 'output_file_contents.output_file_id' in foreign key 'fk_ofc_of_id'.

chrisknoll:

PR #1028, opened 2019-04-07
Proposed changes to issue 984 PR

chrisknoll:
@aklochkova , here are my proposed changes to your PR:

### New pathway results table structure.
There's 3 tables that are storing the pathway results:  pathway_analysis_codes, pathway_analysis_paths, pathway_analysis_stats.  The _codes table was changed to store names longer than 255 chars: when creating the combo names, this could get quite long, so i changed it to varchar(2000).  The pathway_analysis_paths table was added to store each path as a separate column. I moved away from your introduction of a 'chains' term, because these are all paths, and we don't need additional terminology that could be confusing.  The _stats table remains unchanged because the existing 'runPathwayAnalysis' depends on that code.  The runPathwayAnalysis was already producing the counts for cohortCounts and totalPahtwayCounts, so I removed some redundancy there.

I have included the results schema DDL in the DDL service, so you may need to alter your test env result schema to account for the new tables.
 
### Generate pathway counts via db query (patient-level data no longer returned)
With this change, we're going to have a limit of a max-depth of paths of 10. we can easily change this, but now we are able to do easier analyitics on these tables because we don't have to handle parsing out a 'a-b-c' path text from the table. The 'a-b-c' path format is built int he service for consumption by the front end.

### Re-structured result queries to fetch codes, stats and pathways separately.
The pathway results was combining joins of the paths, codes and stats all together and sorting it out on the client.  I changed this to select each subset of data separately, and assemble it in memory.  I believe it's more efficient this way.

###  Other changes.
I renamed the ChainTasklet to PathwayStatisticsTasklet because it's doing a collection of work with statistics, and I didn't want to introduce a new term 'Chain'. Hopefully you don't mind the switch.

I did not try to make the Tasklet follow the conventions defined in some of the other tasklets (such as extending AnalysisTasklet, which would allow to to be canceled).  But, based on the implementation you started with (and I extended), it doesn't look like the ChainTasklet ever was designed to be cancel-able.  Should it?  

I hope you can accept these changes without too many modifications.  Running on my local env, the performance gains was huge, and less taxing on our DB server too.


chrisknoll:
@aklochkova , if you could give your feedback today on this, i'd like to move this forward into a final PR.  Thanks!

src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java
```diff
@@ -377,7 +375,7 @@ public JobExecutionResource generatePathways(final Integer pathwayAnalysisId, fi
                         sourceService
                 )
         );
-        TransactionalTasklet chainsTasklet = new ChainsTasklet(getSourceJdbcTemplate(source), getTransactionTemplate(), source, 
+        TransactionalTasklet chainsTasklet = new PathwayStatisticsTasket(getSourceJdbcTemplate(source), getTransactionTemplate(), source,
```
aklochkova: change 'chain' to 'pathwayStatistics' everywhere

src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java
```diff
@@ -422,47 +420,41 @@ public PathwayAnalysisResult getResultingPathways(final @PathwayAnalysisGenerati
...
-        }
-
-        cohortPathwaysList.forEach(cp -> cp.setPathwaysCounts(pathwaysCounts.get(cp.getCohortId())));
-
-        result.setCodes(codes);
-        result.setCohortPathwaysList(cohortPathwaysList);
-        return result;        
-    };
+		private final RowMapper<PathwayCode> codeRowMapper = (final ResultSet resultSet, final int arg1) -> { 
+			return new PathwayCode(resultSet.getInt("code"), resultSet.getString("name"), resultSet.getInt("is_combo") == 0 ? false : true);
```
aklochkova: `resultSet.getInt("is_combo") != 0`

src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java
```diff
@@ -422,47 +420,41 @@ public PathwayAnalysisResult getResultingPathways(final @PathwayAnalysisGenerati
...
+		private final RowMapper<CohortPathways> pathwayStatsRowMapper = (final ResultSet rs, final int arg1) -> { 
+			CohortPathways cp = new CohortPathways();
+			cp.setCohortId(rs.getInt("target_cohort_id"));
+			cp.setTargetCohortCount(rs.getInt("target_cohort_count"));
+			cp.setTotalPathwaysCount(rs.getInt("pathways_count"));
+			return cp;
+		};		
+
+		private final ResultSetExtractor<Map<Integer, Map<String, Integer>>> pathwayExtractor = (final ResultSet rs) -> { 
+			Map<Integer, Map<String, Integer>> cohortMap = new HashMap<>();  // maps a chortId to a list of pathways (which is stored as a Map<String,Integer>
```
aklochkova: cohortId

src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java
```diff
@@ -473,15 +465,44 @@ public String findDesignByGenerationId(@PathwayAnalysisGenerationId final Long i
...
+			);
+			Map<Integer, Map<String, Integer>> pathwayResults = 
+							getSourceJdbcTemplate(source).query(pathwayResultsPsr.getSql(), pathwayResultsPsr.getOrderedParams(), pathwayExtractor);
+			
+			cohortStats.stream().forEach((cp) -> {
+				cp.setPathwaysCounts(pathwayResults.get(cp.getCohortId()));
+			});
+			
+			PathwayAnalysisResult result = new PathwayAnalysisResult();
+			result.setCodes(new HashSet(pathwayCodes));
```
aklochkova: diamond operator

src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java
```diff
@@ -473,15 +465,44 @@ public String findDesignByGenerationId(@PathwayAnalysisGenerationId final Long i
...
+			result.setCohortPathwaysList(new HashSet(cohortStats));
```
aklochkova: diamond operator

chrisknoll:
Thanks for your review @aklochkova , I believe I hit all your points.

src/main/java/org/ohdsi/webapi/util/CancelableJdbcTemplate.java
```diff
@@ -110,11 +110,12 @@ public String getSql() {
...
-            rowsAffected[i] = query.getUpdateCount();
-          } else if (!suppressApiException) {
-            throw new InvalidDataAccessApiUsageException("Invalid batch SQL statement: " + getSql());
-          }
+					if (!query.execute()) {
+						rowsAffected[i] = query.getUpdateCount();
+					}
+					else if (!suppressApiException) {
+						throw new InvalidDataAccessApiUsageException("Invalid batch SQL statement: " + getSql(current));
+					}
```
aklochkova: please align this code block

aklochkova: @chrisknoll just one little note, everything else is ok

chrisknoll: I think this is just tabs v. spaces. I'll check the formatting on my local IDE, I believe the indentation is correct (if using tabs).

chrisknoll: yes, this is just tabs showing up very indented in git.  The other formatting uses 8 spaces (?!).  I've seen other places where they use 2 tab indents (not 2 spaces per tab...2 tabs..) it's odd.  But we'll look closer at code formatting in a future PR.

PR #1029, opened 2019-04-08
Cascade type was changed to "MERGE" for children (design field) of FeAnalysisWithCriteriaEntity (#785)

### LINKED ISSUE #785: Feature Analyses update optimizations
Per @pavgra, we'd like to revisit #784 to get a better understanding of what is happening under the hood with the Hibernate session and transactions. The current implementation fixes the issue reported in #783 but is not the desired long-term solution.

ssuvorov-fls:
fixes #785

### DIFF
--- a/src/main/java/org/ohdsi/webapi/feanalysis/domain/FeAnalysisWithCriteriaEntity.java
+++ b/src/main/java/org/ohdsi/webapi/feanalysis/domain/FeAnalysisWithCriteriaEntity.java
@@ -11,7 +11,7 @@ import java.util.Objects;
 @Entity
 public abstract class FeAnalysisWithCriteriaEntity<T extends FeAnalysisCriteriaEntity> extends FeAnalysisEntity<List<T>> implements FeatureAnalysisWithCriteria<T, Integer> {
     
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/feanalysis/domain/FeAnalysisWithCriteriaEntity.java
```diff
@@ -11,7 +11,7 @@
 @Entity
 public abstract class FeAnalysisWithCriteriaEntity<T extends FeAnalysisCriteriaEntity> extends FeAnalysisEntity<List<T>> implements FeatureAnalysisWithCriteria<T, Integer> {
     
-    @OneToMany(targetEntity = FeAnalysisCriteriaEntity.class, fetch = FetchType.EAGER, mappedBy = "featureAnalysis", cascade = CascadeType.ALL)
+    @OneToMany(targetEntity = FeAnalysisCriteriaEntity.class, fetch = FetchType.EAGER, mappedBy = "featureAnalysis", cascade = CascadeType.MERGE)
```
pavgra: If we previously had `CascadeType.ALL`, wouldn't the switch to only `MERGE` be too limited? I mean, don't we need e.g. `CascadeType.REMOVE`?

ssuvorov-fls: fixed

chrisknoll:

PR #1030, opened 2019-04-08
Fixes Error doesn't sent while creating Incidence Rate

### LINKED ISSUE #948: Error isn't sent while creating Incidence Rate
Steps to reproduce:

1. Create new Incidence Rate
2. Press 'Save'
3. Return to Incidence Rates' list and ensure that yours IR wasn't created (probably due to issue with sequences in database)

ER:
error is sent to frontend

AR:
error isn't sent
image

wivern:
Fixes #948


pavgra:
@wivern , which case does it solve? I see exception returned after an attempt to create IR using the wrong sequence even with current `master`

image

wivern:
@pavgra Very strange, this is what I get having wrong sequence with current master:

network
idea

wivern:
Anyway, the `ErrorMvcAutoConfiguration` removes default error page that might be confusing

pavgra:
I see. The thing was that I set IR sequence to produce duplicates. But your case is about duplicates in permissions tables

chrisknoll:

PR #1035, opened 2019-04-09
Fixes issue with import new estimation/prediction

wivern:
fixes OHDSI/Atlas#1665


chrisknoll:

PR #1039, opened 2019-04-12
updated getConceptRecordCount.sql to include new Achilles analysis_ids

AnthonyMolinaro:
Updated the query that populates rc and drc in ATLAS to include the Achilles analysis_ids for computing source_code_concept_id counts, unit_concept_id counts, and value_as_concept_id counts.  See following PR;

[source code analyses](https://github.com/OHDSI/Achilles/pull/393)
[unit and value analyses](https://github.com/OHDSI/Achilles/pull/395)


anthonysena:
Making a note that after a review, I think these changes work well on the WebAPI side. There will need to be a corresponding set of changes on the ATLAS UI to retrieve record counts for non-standard codes.

PR #1047, opened 2019-04-17
Handles CC skeleton as external package in Hydra

wivern:


chrisknoll:

PR #1048, opened 2019-04-17
Throws an error when Estimation import failed

### LINKED ISSUE #1032: Can't import new estimation from example
json representations of cohorts, excluded concept sets and estimation are in attach archive

wivern:
Closes #1032


anthonysena:
@wivern this all looks good from my side after testing. A couple of requests: 

- Could you apply the same approach for the Prediction feature? It will certainly have the same potential error if someone uploads an older prediction specification.
- Could you resolve the conflict error?

I'll then approve and we can merge this in. Thanks!

PR #1052, opened 2019-04-18
Sensitive info is removed from source files with media type "text"

### LINKED ISSUE #1046: log.txt shows connection info
I noticed that in the log.txt that I downloaded for failed estimation analysis job that the connection info (User id and password) is being shown in clear text.

ssuvorov-fls:
fixes #1046


src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
```diff
@@ -1,22 +1,76 @@
...
 
-  @Override
-  public AnalysisResultFile filterSensitiveInfo(AnalysisResultFile source, Map<String, Object> variables, boolean isAdmin) {
+    private List<String> sensitiveExtensions;
 
-    if (MediaType.TEXT_PLAIN.equals(source.getMediaType())) {
-      final String value = filterSensitiveInfo(new String(source.getContents()), variables, isAdmin);
-      source.setContents(value.getBytes());
+    @Value("${sensitiveinfo.analysis.extensions}")
+    private String sensitiveAnalysisExtensions;
```
pavgra: Why not `@Value("#{'${sensitiveinfo.analysis.extensions}'.split(',')}")` or `@Value("${sensitiveinfo.analysis.extensions}") private String[] sensitiveAnalysisExtensions;`

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
```diff
@@ -1,22 +1,76 @@
...
+    public AnalysisResultFile filterSensitiveInfo(AnalysisResultFile source, Map<String, Object> variables, boolean isAdmin) {
+        // some txt files have media type "text" instead of MediaType.TEXT_PLAIN
+        if (isNeedFiltering(source)) {
+            final String value = filterSensitiveInfo(new String(source.getContents()), variables, isAdmin);
+            source.setContents(value.getBytes());
+        }
+        return source;
+    }
+
+    private boolean isNeedFiltering(AnalysisResultFile source) {
```
pavgra: Could you rename it to `isFilteringRequired`?

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
```diff
@@ -1,22 +1,76 @@
...
+        return MediaType.TEXT_PLAIN.equals(source.getMediaType()) || "text".equals(source.getMediaType()) || checkExtension(source);
+    }
+
+    private boolean checkExtension(AnalysisResultFile source) {
+        String extension = FilenameUtils.getExtension(source.getFileName());
+        for (String value : sensitiveExtensions) {
```
pavgra: Wouldn't it be simpler if you store `sensitiveExtensions` in a HashSet and use contains?

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
ssuvorov-fls: fixed, changed to array version

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
ssuvorov-fls: fixed

pavgra:
The result files, which you iterate over, are:

image
image

while the mentioned `log.txt` is in `3003_results.zip` -> `/results/log.txt`

How is it going to be filtered?

pavgra:
Also, a back-merge will be required

ssuvorov-fls:
added filtering of archive content

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
```diff
@@ -42,21 +50,41 @@ public void init() {
...
+
+    private AnalysisResultFile doFilterSensitiveInfo(AnalysisResultFile source, Map<String, Object> variables, boolean isAdmin) {
         if (isFilteringRequired(source)) {
             final String value = filterSensitiveInfo(new String(source.getContents()), variables, isAdmin);
             source.setContents(value.getBytes());
         }
         return source;
     }
 
+    private byte[] doFilterSensitiveInfo(Path path, byte[] bytes, Map<String, Object> variables, boolean isAdmin) {
```
pavgra: I would stick to single-source of truth: `bytes` represents info which is derived from `path`. So why cannot we just pass `path` to avoid potential contradictions and anomalies?

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileSensitiveInfoServiceImpl.java
```diff
@@ -66,4 +94,66 @@ private boolean checkExtension(AnalysisResultFile source) {
             return sensitiveExtensions.contains(extension);
         }
     }
+
+    private boolean isArchive(AnalysisResultFile resultFile) {
+        String extension = FilenameUtils.getExtension(resultFile.getFileName());
+        return EXTENSION_ZIP.equalsIgnoreCase(extension);
+    }
+
+    private AnalysisResultFile checkZipArchive(AnalysisResultFile resultFile, Map<String, Object> variables, boolean isAdmin) {
```
pavgra: Seems like a compute intense task. I suggest doing this processing one-time when results are received from EE. Not on each download

chrisknoll:
The migration could break as a result of future changes to the repository definition.  Instead of using the `repository.findAll()`, you should directly query the database to fetch the data _based on the db schema at the time the migration was written_ and process the data via raw operations, and not depend on any classes that may change implementation over time.

ssuvorov-fls:
@chrisknoll 
There is a problem with AnalysisResultFileContentSensitiveInfoServiceImpl.filterSensitiveInfo - i must pass there an AnalysisResultFileContent object, so migration will still depend on class that may implementation.

chrisknoll:
Ok, I think it may be possible to work around this by making copies or 'legacy' impelmentations of the Service interfaces that we can inject into the migration. But, that sounds like a lot of work.  We should be able to see errors in the migration script if something gets updated in the future, so we can address making the migrations work with the new changes when the need arises.

ssuvorov-fls:
@chrisknoll 
migration now uses raw operations

chrisknoll:
Excellent work, @ssuvorov-fls .  This will guard against future changes that may break migration.  Thanks!

chrisknoll:
Why wasn't this branch deleted?

PR #1060, opened 2019-04-19
Allow beans from dependencies with the same package name

### LINKED ISSUE #1059: Allow WebAPI extension by including Beans from dependencies
WebAPI could be easily customized if it would appreciate beans located in dependencies (but in the same namespace).

Proposal:
`@ComponentScan(basePackages = {"org.ohdsi.webapi"})`

pavgra:
fixes #1059


pavgra:
Seems like I was confused by issues with build and my local environment: the annotation is not required to support beans from dependencies

Issue #1058, opened 2019-04-19
Excessive ObjectMappers creation


pavgra:
There are multiple places in WebAPI where a new instance of ObjectMapper is initiated and configured. Not only such initializations often take place inside of REST endpoints which reduces performance, but also this means that multiple instances of ObjectMapper could serialize / deserialize study designs in a different way.

Proposal:
Introduce a single source of truth, bean with ObjectMapper creation and configuration. And autowire that single instance of ObjectMapper all across the application.

PR #1061, opened 2019-04-22
 Use autowired ObjectMapper instead of initializing separate ones

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/1058


chrisknoll:

PR #1062, opened 2019-04-23
Removed CC sql which was replaced by appropriate SQL in CC skeleton

pavgra:
Otherwise, it shadows required resource file


chrisknoll:

Issue #649, opened 2018-10-18
Geospaital functionality in Atlas: integration of AEGIS


gowthamrao:
### Geospatial functionality in Atlas: integration of Aegis

**Authors:** Gowtham Rao, Seng Chan You, Jaehyeong Cho, Andrew Williams, Robert Miller, Pavel Grafkin, Gregory Klebanov

**Motivation:** During the 2018 OHDSI Symposium, Washington DC USA - J Cho, SC You, K Kim, Y Soh, D Kim, RW Park - presented a software demonstration called _&#39;Application for Epidemiological Geographic Information System (AEGIS) - An open
source spatial analysis tool based on CDM&#39;_ .  See AEGIS under related work.

AEGIS, was built to support 5.x version of the OMOP CDM. This version did not have latitude and longitude in location table. AEGIS developers used observation and fact\_relationship table to design AEGIS using CDM 5.x. The OMOP CDM 6+ (released October 2018) has two location tables (location and location\_history). The location table has fields for latitude and longitude. These new fields may be used to represent precise location of persons, providers or care\_sites.

During the presentation, a decision was made to upgrade AEGIS to support CDM 6.x with new location table, and to evaluate if it was possible to integrate AEGIS like functionality, and the work of the OHDSI GIS workgroup into ATLAS.

**Background:** Spatial epidemiology is the description, analysis or surveillance of a populations health related factors such as medical service, diseases, in relation to other person level or area level factors like demographic, environmental exposure, behavioral determinants, socio-economic indicators, genetic and infectious risk factors. Two types of spatial epidemiology are discussed below.

Descriptive mapping, widely used in spatial epidemiology, is useful for establishing initial hypotheses about the patterns of incidence/prevalence in an area, or the correlation between exposure to specific factors and disease.

Cluster detection is a more advanced statistical method that may reveal geographic clusters, based on patterns and spatial correlation.

gowthamrao:
**Use-cases:**

- Hot spotting
https://www.scirp.org/journal/PaperInformation.aspx?PaperID=50032
https://www.newyorker.com/magazine/2011/01/24/the-hot-spotters
- Super-Utilizers
http://usa.thelancet.com/blog/2017-01-24-identifying-and-treating-%E2%80%9Csuper-utilizers%E2%80%9D
https://catalyst.nejm.org/superutilizer-care-teamwork-barriers-to-care/

gowthamrao:
**Requirement 1: Area shape selection:**
[...]

PR #1065, opened 2019-04-23
Added Location Region analyses to CC

pavgra:
image


anthonysena:
@pavgra can you link this to an existing issue or open a new one?

pavgra:
Related to https://github.com/OHDSI/WebAPI/issues/649

anthonysena:
From discussion with @pavgra and @gklebanov, I'm moving this to the 3.0 milestone since these changes utilize CDM 6.x functionality.

alondhe:
@anthonysena -- this looks like work that could be really useful; we're starting to need more location-based analytics. If we were to refactor it to work with CDM 5.3 / 5.4, do you think that'd be also useful for others?

anthonysena:
@alondhe - we made this PR a draft since it required tables in CDM v6.0 (specifically `location_history` but there may be others I'm missing). I'm not sure how easily this is ported to CDM v5.3/v5.4 but you could attempt to create some custom feature, as was done in this PR, to try it out in your environment.

PR #1070, opened 2019-04-25
Implement Caching.

chrisknoll:
Caching implemented by Ehcache, but supports JCache API so any provider can be plugged in.
Updated spring-boot to 1.5.20.
Removed unused methods from CDMResultsService.java.

I would like assistance/input with how the configuration can be overloaded by settings.xml.  I've already made the path to the ehcache.xml file a application.property, I do not know if that is enough:  Should the resource be coppied to the WAR file, or does the WebAPI process just use the path provided to read the local (local to the servlet engine) filesystem?  I'm not sure what the best approach is, and I don't have enough experience with maven plugin options to see how other options could work.  Need community feedback here.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jersey.version>2.14</jersey.version>
     <!-- Spring Boot manages spring.version as well -->
-    <spring.boot.version>1.5.2.RELEASE</spring.boot.version>
+    <spring.boot.version>1.5.20.RELEASE</spring.boot.version>
     <spring.batch.version>3.0.8.RELEASE</spring.batch.version>
     <flyway.version>4.2.0</flyway.version>
     <waffle.version>1.7.3</waffle.version>
@@ -160,6 +160,10 @@
 
     <hydra.externalPackage.prediction></hydra.externalPackage.prediction>
     <hydra.externalPackage.estimation></hydra.externalPackage.estimation>
+		
+		<!-- Cache Config -->
+		<spring.cache.jcache.config>classpath:config/ehcache.xml</spring.cache.jcache.config>
+			
   </properties>
   <build>
     <finalName>WebAPI</finalName>
@@ -231,7 +235,7 @@
         <configuration>
           <skipTests>true</skipTests>
         </configuration>
-</plugin>
+			</plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
@@ -406,7 +410,6 @@
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter</artifactId>
-      <version>${spring.boot.version}</version>
       <exclusions>
         <exclusion>
           <groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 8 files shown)


src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -149,6 +147,7 @@ protected PreparedStatementRenderer prepareGetConceptRecordCount(Integer[] ident
     @GET
     @Path("{sourceKey}/dashboard")
     @Produces(MediaType.APPLICATION_JSON)
+		@CacheResult(cacheName="datasources.dashboard")
```
pavgra: I think it would make sense to invalidate the cache when user clicks "Refresh cache" in the Configuration screen

pom.xml
```diff
@@ -433,18 +434,27 @@
...
       <artifactId>spring-boot-starter-jersey</artifactId>
-      <version>${spring.boot.version}</version>
     </dependency>
     <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-data-jpa</artifactId>
-      <version>${spring.boot.version}</version>
     </dependency>
+		<dependency>
+				<groupId>org.springframework.boot</groupId>
```
pavgra: Would be nice to have consistent indentation

chrisknoll: The additions are using tab-based indentation, while the existing file was space-based indentation.  If I reformatted the whole file, there would be complaints of white-space changes.  If I use spaces, then we're not following the decision to indent with tabs.  So, I went with tab-indentation, and it will have a mix of indentation in the file.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
chrisknoll: I'll need to look up how to access the cache directly. A little bit of it is hidden behind the `@CacheResult` annotation (it looks up the appropriate cache and uses the paramter of the function as the key.  It's probably a simple matter to 'clear cache' but if we expect to do 'selective removal' from a cache, we might need to think about providing other classes (like a CacheKeyGenerator) and use that mode of entering elements in a cache....since this is just starting to scratch the surface, i went with the quickest/straight forward approach for entering elements in the cache, but I'd rather have the long term approach implemented for this PR..i'm just not exactly sure what it is.

chrisknoll: @pavgra , i'm not sure what you're referring to as "Refresh Cache", there is a 'clear configuration cache' but I don't see a 'refresh cache'.  I don't know if I would bundle the configuration cache clear with the Datasources Cache.  Maybe a different sub-page under configuration needs to be introduced to display the set of caches, and let the user pick the cache to invalidate (by invoking a `/cache` endpoint that will clear that cache.

pavgra: image

pom.xml
pavgra: image

chrisknoll: I see. your initial comment wasn't clear. I can make sure the tab indentation is consistent between all tab indentation.

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
chrisknoll: I see it, it was something that was recently introduced, and I believe depends somewhat on enabled security.  I'll try to find which endpoint the refresh button calls back to, and see if I can implement.  The challenge here is that the caches use the source-key as one of the elements in the cache-key, so, it would be a matter of cycling through each cache (dashboard, treemap, person, etc) and removing only those cache elements which belong to the given source.  

This sounds like it'd be a maintenance nightmare, instead I think the approach would be to just have one button to clear all caches.

chrisknoll: Added a rest endpoint.  A separate UI PR can add it to the UI.

PR #1078, opened 2019-05-06
Added 'resultExclusions' configurable param

aklochkova:
Fixes https://github.com/OHDSI/ArachneExecutionEngine/issues/192


wivern:
But why not to exclude drivers on the EE side?
1. Paths to drivers is configurable and might differs from defaults.
1. I can't imagine case when someone wants to receive drivers with results.

chrisknoll:
Why wasn't the branch deleted?

pavgra:
Deleted

PR #1102, opened 2019-05-20
Fixes Sql injection vulnerability

wivern:


wivern:
> What about `timeWindows` and `domains` which can be injected even easier?

But `timeWindows` and `domains` are escaped by quoted strings

pavgra:
@wivern ,

> domainClauses.add(String.format("lower(ar.domain_id) = lower('%s')", domain));

What stops you from providing a domain = `' + <your custom sql> + '`?

Issue #1100, opened 2019-05-20
Broken import of PLP, PLE, CC, Pathway


aklochkova:
Steps to reproduce:
try to import PLP from attached file

Expected result:
PLP is imported

Actual result:
![image](https://user-images.githubusercontent.com/38942654/58020578-2eb84700-7b11-11e9-9775-aa6cc0984ae2.png)

the reason is -- lack of db migration which updates `CohortExpression` according to appearance `cdmVersionRange` field
hashcodes of persisted `CohortExpression` and incoming one are different, so `DesignImportService#findCohortByExpressionHashcodeAndPredicate` can't find already existing entity
[broken PLP.txt](https://github.com/OHDSI/WebAPI/files/3197830/broken.PLP.txt)

chrisknoll:
Question for you:

If the imported cohort definition hashes differently from the existing cohort definition (since the imported cohort definition will hash based on the `cdmVersionRange`, while the existing one did not) why doesn't that just result in a new cohort definition being imported?  And why does this result in a duplicate name error (from your screenshot)?

pavgra:
> why doesn't that just result in a new cohort definition being imported

Because there is already a cohort definition in DB having the same name as the imported cohort

chrisknoll:
> Because there is already a cohort definition in DB having the same name as the imported cohort

Ok, so if you have 2 different characterizations containing cohorts that have the same names (let's say, CC1 and CC2 both have a cohort called 'Cancer' but cancer is defined differently (ie: different hash)), will there be a problem importing CC1 (where the cohort is new and Cancer doesn't exist yet) and CC2 is imported (where now there is an existing cohort named 'Cancer' but has a different hash)?

aklochkova:
@chrisknoll yes, there will be a problem in this case because we will not find the 'Cancer' cohort with different hash and when we try to add this cohort the unique name constraint will raise

pavgra:
Since the hash is calculated based on the design field (so the cohorts actually differ) it would be reasonable to alter the name of the imported cohort to avoid clashes, e.g. "Cancer" -> "Cancer (1)"

chrisknoll:
[...]

PR #1104, opened 2019-05-20
Added migration for Cohort Expression update

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1100


chrisknoll:
@aklochkova , in your migraiton script, you have:
``
            CohortExpression expression = Utils.deserialize(details.getExpression(), CohortExpression.class);            
            details.setExpression(Utils.serialize(expression));
            details.updateHashCode();
            detailsRepository.save(details);
``
The deserialize->serialize step seems to be the exact same actions that is in `updateHashCode()`. Is it enough to just get the details, call `updateHashCode()`, and then `save()`?  Or am I missing something between the serialize/deserialize and the updateHashCode?

pavgra:
@chrisknoll , the `updateHashCode` should be sufficient

aklochkova:
@chrisknoll, @pavgra 
but I need to update the `expression` field in CohortDetails, because new field was introduced (`cdmVersionRange`), so it isn't enough only to update hashcode and call `save()`

PR #1106, opened 2019-05-20
Update SQL migration - fixes #1105

### LINKED ISSUE #1105: Unique constraint application in SQL Server
Per @aklochkova's detailed comment here: https://github.com/OHDSI/WebAPI/pull/1073#issuecomment-493748843, the update made to the SQL Server migration script that enforces unique names for various entities in WebAPI can potentially fail.

The changes I made for SQL Server as part of that pull request stemmed from the fact that creating, executing and then dropping stored procedures on SQL 2012 was failing in our environment. Instead, I revised that migration to use a set of UPDATE statements which works with the exception that @aklochkova described. 

For simplicity: I'd propose we revise the script so that instead of attempt to rename entities with a sequence number (i.e. adding a "(1)", "(2)", etc to the end of the entity name), we instead use the primary key from the table as the suffix. While this doesn't completely address the edge-case that @aklochkova described, I would expect that this would allow things to work 99% of the time.

anthonysena:
Re-applying the approach originally developed by @aklochkova and made a small modification to the names of the unique constraint. This addresses the problem nicely described here: https://github.com/OHDSI/WebAPI/pull/1073#issuecomment-493748843. 

@aklochkova - I'm not sure why but when I re-tested on our SQL Server instance with the original approach, it worked fine. It may have been something in my environment that caused my problems when testing your change. Thanks for keeping an eye on the changes and for coming up with a robust solution to resolve the duplicate names!

### DIFF
--- a/src/main/resources/db/migration/sqlserver/V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql
+++ b/src/main/resources/db/migration/sqlserver/V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql
@@ -1,198 +1,553 @@
 -- Alter Concept Set table
 
-UPDATE ${ohdsiSchema}.concept_set
-SET concept_set_name = u.concept_set_name + ' (' + CAST(u.rn AS varchar(15)) + ')'
-FROM 
-  ${ohdsiSchema}.concept_set, 
-  (
-    SELECT d.concept_set_id, d.concept_set_name, d.rn
-    FROM
-    (
-      select a.concept_set_name, cd.concept_set_id, ROW_NUMBER() OVER(PARTITION BY a.concept_set_name ORDER BY cd.concept_set_id ASC) rn
-      FROM
-      (
-        select concept_set_name, COUNT(*) cnt
-        FROM ${ohdsiSchema}.concept_set
-        group by concept_set_name having COUNT(*) > 1
-      ) a
-      INNER JOIN ${ohdsiSchema}.concept_set cd ON a.concept_set_name = cd.concept_set_name
-    ) d 
-  ) u
- WHERE u.concept_set_id = ${ohdsiSchema}.concept_set.concept_set_id 
-;
+CREATE PROCEDURE ${ohdsiSchema}.rename_cs_names AS
+BEGIN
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/sqlserver/V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql
```diff
@@ -1,198 +1,553 @@
...
- WHERE u.concept_set_id = ${ohdsiSchema}.concept_set.concept_set_id 
-;
-
-ALTER TABLE ${ohdsiSchema}.concept_set ADD CONSTRAINT uq_cs_name UNIQUE (concept_set_name);
+CREATE PROCEDURE ${ohdsiSchema}.rename_cs_names AS
+BEGIN
+    DECLARE @duplicate_names TABLE(id int IDENTITY (1, 1), duplicate_name varchar(100));
+    DECLARE @name_repeats TABLE (id int IDENTITY (1, 1), repeat_number int);
+    DECLARE @amount_of_duplicate_names int;
+    DECLARE @amount_of_constraints int;
```
aklochkova: remove the variable please, it is unused now (and the same for other tables)

src/main/resources/db/migration/sqlserver/V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql
```diff
@@ -1,198 +1,553 @@
...
+    DECLARE @constraint_title varchar(100);
```
aklochkova: remove the variable please, it is unused now (and the same for other tables)

src/main/resources/db/migration/sqlserver/V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql
```diff
@@ -1,198 +1,553 @@
...
+    DECLARE @schema_title varchar(100);
```
aklochkova: remove the variable please, it is unused now (and the same for other tables)

PR #1109, opened 2019-05-21
Fix databaseId in PLE R script(#1079)

AntonStepanof:
fixes https://github.com/OHDSI/ArachneExecutionEngine/issues/193


src/main/resources/resources/estimation/r/runAnalysis.R
```diff
@@ -30,6 +30,7 @@ tryCatch({
         resultsDatabaseSchema <- Sys.getenv("RESULT_SCHEMA")
         cohortsDatabaseSchema <- Sys.getenv("TARGET_SCHEMA")
         cohortTable <- Sys.getenv("COHORT_TARGET_TABLE")
+        databaseId <- Sys.getenv("DATA_SOURCE_NAME")
```
wivern: What if env var was not set?

AntonStepanof: I fixed that at https://github.com/OHDSI/ArachneExecutionEngine/pull/201 
Please check
I will close this one, because it is already at v2.7.2-rc

Issue #1108, opened 2019-05-21
Characterization execution failed on GCP


konstjar:
![image](https://user-images.githubusercontent.com/38942654/58086806-00e20980-7bc8-11e9-8ae1-1a2863b1baa5.png)

![image](https://user-images.githubusercontent.com/38942654/58086872-22db8c00-7bc8-11e9-9998-8ef6b488bc6e.png)

![image](https://user-images.githubusercontent.com/38942654/58086908-2e2eb780-7bc8-11e9-8f9f-fb6e99cbc7bb.png)

PR #1121, opened 2019-05-23
Added 'dataset' parameter to bigquery 

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1108


src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateCohortCharacterizationTasklet.java
```diff
@@ -110,7 +110,12 @@ protected void doBefore(ChunkContext chunkContext) {
                     .replaceAll("#", tempSchema + "." + sessionId + "_")
                     .replaceAll("tempdb\\.\\.", "");
         }
-        final String translatedSql = SqlTranslate.translateSql(sql, source.getSourceDialect());
+        final String translatedSql;
+        if (DBMSType.BIGQUERY.getOhdsiDB().equals(source.getSourceDialect())){
```
pavgra: Having `if` doesn't look like a solid solution. Cannot we avoid it?

PR #1123, opened 2019-05-23
Fixes IR execution on non-accessible IRs

wivern:
Fixes OHDSI/Atlas#1740

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
+++ b/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
@@ -42,6 +42,7 @@ import org.ohdsi.webapi.shiro.Entities.UserRepository;
 import org.ohdsi.webapi.shiro.annotations.DataSourceAccess;
 import org.ohdsi.webapi.shiro.annotations.SourceKey;
 import org.ohdsi.webapi.shiro.management.Security;
+import org.ohdsi.webapi.shiro.management.datasource.SourceAccessor;
 import org.ohdsi.webapi.source.Source;
 import org.ohdsi.webapi.source.SourceDaimon;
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -355,11 +362,15 @@ public IRAnalysisDTO saveAnalysis(final int id, IRAnalysisDTO analysis) {
...
   @Override
-  public JobExecutionResource performAnalysis(final int analysisId, final String sourceKey) {
+  @DataSourceAccess
+  public JobExecutionResource performAnalysis(final int analysisId, final @SourceKey String sourceKey) {
     Date startTime = Calendar.getInstance().getTime();
 
     Source source = this.getSourceRepository().findBySourceKey(sourceKey);
 
+    ExceptionUtils.throwNotFoundExceptionIfNull(source, String.format("There is no source with sourceKey = %s", sourceKey));
+    sourceAccessor.checkAccess(source);
```
pavgra: How can we reach the line if there is `this.sourcePermissionTemplates.put("ir:*:execute:%s:get", "Generate IR on Source with SourceKey = %s");` in `AtlasSecurity`? Could you provide test case?

wivern: This case is available with disabled or customized security. Anyway from my point of view we cannot rely on externalized setting and omit incoming parameters verification.
Test case is a good idea but I'm not sure what to test?

wivern: `throwNotFoundExceptionIfNull` is used to ensure requested source exists to prevent unexpected NPE.
And if security disabled `sourceAccessor` always passes.

PR #1128, opened 2019-05-27
fixes pathway results migration

wivern:
Fixes OHDSI/Atlas#1778


src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20190410103000__migratePathwayResults.java
```diff
@@ -49,9 +49,12 @@ public void savePathwayCodes(List<Object[]> pathwayCodes, Source source, Cancela
 			// clear existing results to prevent double-inserts
 			params = new String[]{"results_schema"};
 			values = new String[]{resultsSchema};
-			String truncateSql = SqlRender.renderSql("TRUNCATE TABLE @results_schema.pathway_analysis_codes;", params, values);
-			String translatedSql = SqlTranslate.translateSql(truncateSql, source.getSourceDialect()); // TRUNCATE could lead to multiple statements
-			Arrays.asList(SqlSplit.splitSql(translatedSql)).forEach(jdbcTemplate::execute);
+
+			// delete only codes that belongs current Atlas instance
```
chrisknoll: Could you explain what 'current atlas instance' means?  The results schema should be dedicated to a specific instance, but this sounds like you may have multiple WebAPI instances  writing to a single results schema. Is that the case here?

wivern: Yes, source with results schema could be shared between several Atlases.

chrisknoll: Then that's an invalid configuration.  If you have multiple WebAPIs referencing the same results schema, then you'll have identifiers (like cohort IDs) from one WebAPI conflicting with identifiers that are generated on a different WebAPI.  Perhaps this needs to be described in more detail in documentation, but it should be clear that each WebAPI instance needs to write to their own results schema.

As far as this PR goes, I'm not sure that what you are proposing is `correct` but I believe that the result on your own test environment got the desirerd result.  What I am not clear on is how you can establish that the 'codes that belong to current atlas' actually belong to the 'current atlas'.  What keeps the same 'code' from appearing across the different instances of Atlas?  We wouldn't have to worry about these concerns if we establish that each instance gets their own results schema.

wivern: I agree such configuration should not used in real environment.
But let's imaging someone can make a mistake resulting completely deletion of portion of data and making incomplete unexpected results of pathway analyses.
With replacement `TRUNCATE TABLE` to `DELETE FROM` we just providing a foolproof.
There is no differences for properly configured environments.

The idea to use different result schemes is really great. Thank you.

chrisknoll: Ok, thanks for the feedback.  deletes also work slower than truncates, but the difference is probably marginal.  Let's try to proceed with this approach.

chrisknoll:
Could you resolve the conflicts and we'll get this merged?

PR #1136, opened 2019-05-29
Build JAR file additionally to WAR

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/1135

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -268,6 +268,8 @@
         <version>2.2</version>
         <configuration>
           <failOnMissingWebXml>false</failOnMissingWebXml>
+          <attachClasses>true</attachClasses>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -268,6 +268,8 @@
         <version>2.2</version>
         <configuration>
           <failOnMissingWebXml>false</failOnMissingWebXml>
+          <attachClasses>true</attachClasses>
+          <classesClassifier>classes</classesClassifier>
```
chrisknoll: this is the default value.

pavgra:
@chrisknoll , I removed explicit `classesClassifier`. Anything else?

chrisknoll:
No, that was it. Thanks. Only thing left is to delete the branch.

PR #1159, opened 2019-05-31
Fix Oracle migration

wivern:


src/main/resources/db/migration/oracle/V2.7.1.20190405124500__split_output_files.sql
```diff
@@ -6,8 +6,7 @@ CREATE TABLE ${ohdsiSchema}.output_file_contents (
 
 ALTER TABLE ${ohdsiSchema}.output_file_contents
   ADD CONSTRAINT fk_ofc_of_id FOREIGN KEY (output_file_id)
-  REFERENCES ${ohdsiSchema}.output_files (id)
-  ON UPDATE NO ACTION ON DELETE CASCADE;
+  REFERENCES ${ohdsiSchema}.output_files (id);
```
pavgra: This changes behavior:
> If you omit this clause, then Oracle does not allow you to delete referenced key values in the parent table that have dependent rows in the child table.
https://docs.oracle.com/cd/B19306_01/server.102/b14200/clauses002.htm

PR #1160, opened 2019-05-31
Fixes AD groups import fails

### LINKED ISSUE #1155: AD groups import fails
image

wivern:
Fixes #1155


anthonysena:
@wivern @pavgra  I'd propose closing this in favor of #1162 since we should not be making more commits to the 2.7.2-rc brach.

PR #1162, opened 2019-05-31
Fixes failed AD group import

### LINKED ISSUE #1155: AD groups import fails

wivern:
Fixes #1155


src/main/resources/db/migration/postgresql/V2.8.0.20190531181956__alter_job-execution-params_string-val.sql
```diff
@@ -0,0 +1,121 @@
+DROP VIEW ${ohdsiSchema}.cc_generation;
+DROP VIEW ${ohdsiSchema}.estimation_analysis_generation;
+DROP VIEW ${ohdsiSchema}.pathway_analysis_generation;
+DROP VIEW ${ohdsiSchema}.prediction_analysis_generation;
+DROP VIEW ${ohdsiSchema}.user_import_job_history;
+
+ALTER TABLE ${ohdsiSchema}.batch_job_execution_params ALTER string_val TYPE VARCHAR;
```
anthonysena: We should not change the `batch_*` tables as these are provided by Spring Batch. It is also unclear why these changes are required here?

ssuvorov-fls: fixed

PR #1163, opened 2019-05-31
Hibernate startsWith to use escape - Fixes #1153

### LINKED ISSUE #1153: Import of estimation from v2.7.0 fails in v2.7.2
### Scenario

I've exported an estimation created in Atlas v2.7.0: [ple-2_7_0.txt](https://github.com/OHDSI/WebAPI/files/3240979/ple-2_7_0.txt) and am attempting to import this into a different target Atlas running v2.7.2. In my target environment, I have concept sets and cohort definitions with the same names as in the attached design.

### Expected Result

The design imports without issue

### Actual result

The design import fails with the following stack trace: [ple-2_7_0_import_stack_trace.txt](https://github.com/OHDSI/WebAPI/files/3240995/ple-2_7_0_import_stack_trace.txt). For brevity, here is the relevant part of the error:

`
"message": "Violation of UNIQUE KEY constraint 'uq_cs_name'. Cannot insert duplicate key in object 'dbo.concept_set'. The duplicate key value is ([OHDSI EU 2019] Negative controls).",
`

anthonysena:
Aims to correct the problems detailed in #1153 whereby entity names are not identified using the `startsWith` hibernate operator. 

As a note, I attempted to include the "%" as part of the `@Query` declaration like this:

`@Query("SELECT cs FROM ConceptSet cs WHERE cs.name LIKE ?1 + '%' ESCAPE '\\'")`

However, PostgreSQL did not like this so I instead add the "%" to the end of the string in the `NameUtils.getNameWithSuffix`  call instead.

I have not tested this for all entities but I was able to import the estimation example provided in #1153 which properly imported the estimation and the dependent cohort/concept set definitions without issue with these fixes.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/util/NameUtils.java
+++ b/src/main/java/org/ohdsi/webapi/util/NameUtils.java
@@ -22,7 +22,7 @@ public final class NameUtils {
     public static String getNameWithSuffix(String dtoName, Function<String, List<String>> getNamesLike){
         StringBuilder builder = new StringBuilder(dtoName);
 
-        List<String> nameList = getNamesLike.apply(dtoName);
+        List<String> nameList = getNamesLike.apply(dtoName.replace("[", "\\[") + "%");
         Pattern p = Pattern.compile(Pattern.quote(dtoName) + " \\(([0-9]+)\\)");
         nameList.stream()
                 .map(n -> {
--- a/src/main/java/org/ohdsi/webapi/util/NameUtils.java
+++ b/src/main/java/org/ohdsi/webapi/util/NameUtils.java
@@ -22,7 +22,7 @@ public final class NameUtils {
     public static String getNameWithSuffix(String dtoName, Function<String, List<String>> getNamesLike){
         StringBuilder builder = new StringBuilder(dtoName);
 
-        List<String> nameList = getNamesLike.apply(dtoName.replace("[", "\\[") + "%");
+        List<String> nameList = getNamesLike.apply(formatNameForLikeSearch(dtoName) + "%");
         Pattern p = Pattern.compile(Pattern.quote(dtoName) + " \\(([0-9]+)\\)");
         nameList.stream()
                 .map(n -> {
@@ -39,5 +39,9 @@ public final class NameUtils {
                 .ifPresent(cnt -> builder.append(" (").append(cnt + 1).append(")"));
 
         return builder.toString();
+    }
+    
+    public static String formatNameForLikeSearch(String name) {
+        return name.replace("[", "\\[").replace("]", "\\]").replace("%", "\\%").replace("_", "\\_");
     }
 }
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/repository/CcRepository.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/repository/CcRepository.java
@@ -11,6 +11,7 @@ import org.springframework.data.repository.query.Param;
 public interface CcRepository extends EntityGraphJpaRepository<CohortCharacterizationEntity, Long> {
     Optional<CohortCharacterizationEntity> findById(final Long id);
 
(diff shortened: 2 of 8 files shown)


anthonysena:
@pavgra - just a note for awareness: I've tested this set of changes against SQL Server and PostgreSQL for the following design imports: cohort definition, characterization, pathways, estimation and prediction. All worked without issue and without duplicating entities.

chrisknoll:
Just for information, if we need to do concatenation in the hibernate query expression, you can do it like this:
````
@Query("select p from Product p where p.name = ?1 or p.desc like CONCAT('%',?1,'%') ")
```

Found here: https://stackoverflow.com/questions/23326642/query-annotation-using-like-1

I think the way you have it is fine, however, so I'm not suggesting any changes, just the FYI about how it can be performed.

src/main/java/org/ohdsi/webapi/util/NameUtils.java
```diff
@@ -22,7 +22,7 @@ public static String getNameForCopy(String dtoName, Function<String, List<String
     public static String getNameWithSuffix(String dtoName, Function<String, List<String>> getNamesLike){
         StringBuilder builder = new StringBuilder(dtoName);
 
-        List<String> nameList = getNamesLike.apply(dtoName);
+        List<String> nameList = getNamesLike.apply(dtoName.replace("[", "\\[") + "%");
```
pavgra: What about other MS SQL special chars?
`%`, `_`, `]` - https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-2017

anthonysena: So in testing with multiple combinations of concept set names with `%`, `_`, `]` on SQL Server, the code in the PR works without the need to explicitly use an escape for these characters. I'm a bit puzzled about why only the `[` requires the escape but from my testing, it appears to be the only one.

anthonysena: I'm also linking this issue to https://github.com/OHDSI/WebAPI/issues/585 since it makes me LIKE (pun intended) PostgreSQL even more.

pavgra: @anthonysena , here are results from PG. The special chars definitely require escaping:
image
image

anthonysena: Correct - however, Hibernate appears to handle the special characters in PG without the need to explicitly escape them. I've re-done the same tests for those special characters against PG and it worked without issue.

chrisknoll: yeah, what @anthonysena says makes sense:  when you run the input through the hibernate Query Expression, does it handle those escapes for us?  I'm betting it does. Let me do some research to see if that is the case.

-Chris

pavgra: Could you explain the following to me then?
image

chrisknoll: I found this from 2014:
https://andy722.wordpress.com/2014/04/07/escaping-special-characters-in-hibernate-queries/

And more recent articles doesn't seem to correct this, so, @anthonysena , looks like the paramaters passed into the LIKE expression are passed in as-is, and we'll have to implement a manual escape method for it.  The attached link shows an implementation, but we don't have to go that crazy. We just need something that will escape out those 3 elements.

chrisknoll: > Could you explain the following to me then?
That's a good example showing that we need to handle escaping out the special characters from our entity names.  Thanks.

anthonysena: Thanks @chrisknoll and @pavgra for the attention to these details! I'll implement escaping for all of the potential special characters, re-test and we should be set.

PR #1164, opened 2019-06-03
added logging of error during login

AntonStepanof:
related to https://github.com/OHDSI/Atlas/issues/1183


src/main/java/org/ohdsi/webapi/shiro/lockout/LockoutWebSecurityManager.java
```diff
@@ -25,7 +25,7 @@ public LockoutWebSecurityManager(LockoutPolicy lockoutPolicy) {
 
     @Override
     protected void onFailedLogin(AuthenticationToken token, AuthenticationException ae, Subject subject) {
-
+        log.error("Failed to login: {}", ae.getMessage(), ae);
```
pavgra: I would not treat an attempt to login with incorrect credentials as application exception. And this code, if I am not mistaken, will do exactly this, right? If you need logging, why not using `log.debug`?

AntonStepanof: ok, updated

Issue #1110, opened 2019-05-21
Negative results in Cohort Characterisation


konstjar:
Why do we use negative numbers for Cohort Characterization jobs?

![image](https://user-images.githubusercontent.com/4144173/58097976-43fca680-7be1-11e9-9c5f-aee2c3f9b8c9.png)

chrisknoll:
I believe that is a hash value, we don't have any direct control over what the hash is generating, we just show the value (so you can reference it against other design executions and determine the difference). You're not able to know what the actual difference is based on the hash value, only that there is a difference between them.

pavgra:
I suggest just prettifying the hashes to something more human-readable like `V1`, `V2`, etc. So to do the following transformation at UI:

333333 --> V4
123123 --> V1
789789 --> V3
456456 --> V2
123123 --> V1

anthonysena:
@pavgra - this was closed for 2.7.2 but we have a PR open that is linked to this issue. So, I'm going to re-open this issue and note it for 2.8 but perhaps this warrants a separate issue for the remaining PR instead of linking to this one?

PR #1165, opened 2019-06-03
webapi-1110: save CC sourceInfo before completion 

acumarav:
move saving source info for the CC  job into a separate transaction
#Closes OHDSI/WebAPI#1110


acumarav:
@pavgra  Pathways sourceInfo save change is added

acumarav:
This change was already applied within PR #1174

PR #1169, opened 2019-06-04
Cohort definitions were added to export/import

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/1821

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
+++ b/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
@@ -324,6 +324,7 @@ public class IRAnalysisService extends AbstractDaoService implements GeneratesNo
       IncidenceRateAnalysisDetails details = new IncidenceRateAnalysisDetails(newAnalysis);
       newAnalysis.setDetails(details);
       details.setExpression(analysis.getExpression());
+      clearCohorts(details);
     }
     else {
       newAnalysis.setDetails(null);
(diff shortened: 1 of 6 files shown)


src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -332,7 +333,25 @@ public IRAnalysisDTO createAnalysis(IRAnalysisDTO analysis) {
     return getAnalysis(createdAnalysis.getId());
   }
 
-  @Override
+    private void clearCohorts(IncidenceRateAnalysisDetails details) {
```
pavgra: Cannot we encapsulate this in custom deserializer for IRAnalysisDTO?

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -316,25 +324,110 @@ public IRAnalysisDTO createAnalysis(IRAnalysisDTO analysis) {
...
+          IncidenceRateAnalysis analysis = this.irAnalysisRepository.findOne(id);
+          ExceptionUtils.throwNotFoundExceptionIfNull(analysis, String.format(NO_INCIDENCE_RATE_ANALYSIS_MESSAGE, id));
+
+          try {
+              IncidenceRateAnalysisExpression expression = objectMapper.readValue(
+                      analysis.getDetails().getExpression(), IncidenceRateAnalysisExpression.class);
+
+              // Cohorts are not stored in expression now - create lists of cohorts from
+              // lists of their ids
+              fillCohorts(expression.outcomeIds, expression.outcomeCohorts);
```
anthonysena: Question: do we need to retrieve the cohort expression for every cohort reference in the IR analysis? I think this is only necessary when we export the analysis unless I'm missing something in the approach designed here.

pavgra: @anthonysena , I agree with you - will update the PR

pavgra:
Rebuild according to review in https://github.com/OHDSI/WebAPI/pull/1172

PR #1173, opened 2019-06-05
Added test for entities copy/import/create

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1100


pom.xml
```diff
@@ -50,10 +50,12 @@
 
     <!-- Kerberos properties -->
     <kerberos.timeout>60</kerberos.timeout>
+<!--    <kerberos.configPath>D:\projects\Odysseus\WebAPI\src\main\resources\krb5.conf</kerberos.configPath>-->
```
pavgra: I guess this should be removed?

src/test/java/org/ohdsi/webapi/test/entity/BaseTestEntity.java
```diff
@@ -0,0 +1,73 @@
...
+    public static class TestInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+        @Override
+        public void initialize(ConfigurableApplicationContext context) {
+            JdbcDatabaseContainer container = new PostgreSQLContainer()
+                    .withDatabaseName("ohdsi_webapi")
+                    .withUsername("ohdsi")
+                    .withPassword("ohdsi");
+
+//            container = new MSSQLServerContainer();
```
pavgra: Do we need to store commented out lines?

src/test/java/org/ohdsi/webapi/test/entity/estimation/BaseEstimationTestEntity.java
```diff
@@ -0,0 +1,91 @@
...
+    protected EstimationController esController;
+    @Autowired
+    protected EstimationRepository esRepository;
+    @Autowired
+    protected EstimationService esService;
+    @Autowired
+    protected CohortDefinitionDetailsRepository cdRepository;
+    protected Estimation firstIncomingEntity;
+    protected EstimationDTO firstSavedDTO;
+    protected static final String ES_SPECIFICATION =
```
pavgra: Please put into a resource file

src/test/java/org/ohdsi/webapi/test/entity/prediction/BasePredictionTestEntity.java
```diff
@@ -0,0 +1,43 @@
...
+public abstract class BasePredictionTestEntity extends BaseTestEntity {
+    @Autowired
+    protected PredictionController prController;
+    @Autowired
+    protected PredictionAnalysisRepository prRepository;
+    @Autowired
+    protected PredictionService prService;
+    protected PredictionAnalysis firstIncomingEntity;
+    protected PredictionAnalysisDTO firstSavedDTO;
+    protected static final String PR_SPECIFICATION =
```
pavgra: Resource file pls

src/test/resources/application-test.properties
```diff
@@ -21,8 +21,24 @@ jobservice.endpoint.job=${jobservice.endpoint}/{instanceId}
 jobservice.endpoint.jobexecution=${jobservice.endpoint.job}/execution/{executionId}
 jobservice.endpoint.jobexecution.alt=${jobservice.endpoint}/execution/{executionId}
 
-
-
+security.provider=AtlasRegularSecurity
+
+#mssql properties
+#      
+#spring.profiles.active=webapi-mssql
```
pavgra: Do we need this?

src/test/resources/container-license-acceptance.txt
```diff
@@ -0,0 +1 @@
+mcr.microsoft.com/mssql/server:2017-CU12
```
pavgra: Do we need this?

src/test/resources/ple-example-for-import.txt
```diff
@@ -0,0 +1,6897 @@
+{
```
pavgra: Why `txt` and not `json`?

pom.xml
```diff
@@ -742,6 +742,23 @@
       <version>2.8.7</version>
       <type>jar</type>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <version>1.11.3</version>
```
pavgra: I would move the repeated common version into a property: `<version>${org.testcontainers.version}</version>`

src/test/resources/application-test.properties
aklochkova: should tests be only for postgres?

pom.xml
aklochkova: yes, was pushed by mistake, reverted back

src/test/java/org/ohdsi/webapi/test/entity/cohortcharacterization/create/TestCCCreate.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    public void testCreateWithDuplicateName() {
+        //Action
+        try {
+            ccController.create(firstIncomingDTO);
+            fail();
+        } catch (Exception e) {
+
+        //Assert
+            assertTrue(e.getCause().getCause().getCause().getCause().getMessage().contains("uq_cc_name"));
```
pavgra: Wow, `.getCause().getCause().getCause().getCause()` doesn't look nice

pavgra: Move `"uq_cc_name"` to constant pls

src/test/resources/container-license-acceptance.txt
aklochkova: it is needed for starting mssql test container

src/test/java/org/ohdsi/webapi/test/entity/cohortdefinition/create/TestCDCreate.java
```diff
@@ -0,0 +1,23 @@
...
+    @Test
+    public void testCreateWithDuplicateName() {
+        //Action
+        try {
+            cdService.createCohortDefinition(firstIncomingDTO);
+            fail();
+        } catch (Exception e) {
+
+        //Assert
+            assertTrue(e.getCause().getCause().getMessage().contains("uq_cd_name"));
```
pavgra: Pls get a look at comment above

pavgra:
Could you also describe how to run this using MS SQL?

chrisknoll:
Hey, Everyone,
 I really appreciate introducing these test cases, especially for those test cases where we are testing database interactivity.  I would really appreciate if we could get some testing documentation together so that if anyone else wants to contribute new tests (I would love to do this), that setting up the test environment is understood and the process of introducing new tests is documented.

Few things to cover:
1. The testing container:  these are all JUnit test cases, however we're leveraging an underlying database test container from this artifact:
```
      <groupId>org.testcontainers</groupId>
      <artifactId>testcontainers</artifactId>
      <version>1.11.3</version>
```

Can we get some high-level information on this?

2. the container seems to be initialized with:
```
            JdbcDatabaseContainer container = new PostgreSQLContainer()
                    .withDatabaseName("ohdsi_webapi")
                    .withUsername("ohdsi")
                    .withPassword("ohdsi");

            container.start();    
```

When this container is started, is there a physical Posttgresql instance backing it? In the case of SQL server, does a local install of soemthing like SQLExpress need to be setup/configured?  Or is it that in both cases the XXXContainer() class that is instantiated as a JdbcDatabaseContainer is a 'fake', and any JDBC interation to the container doesn't actually execute any database updates/commands, or is it an emulator?  I just would like to understand the setup-burden of the developer if they want to run the tests.

I also recall @pavgra showing how a database instance is set up in one of the Odysseus apps via an XML configuration where the database is instantiated off test data in XML form. I'm guessing that what we are doing in this PR is a different approach. and I wanted to know which direction do we see going long term?

pavgra:
@chrisknoll , thank you for the feedback, all valid. Let me mark this PR as `[WIP]` and I'll let you know as we finish internal reviews

pavgra:
1. Could you get a look at H2 which would allow us running tests w/o Docker? Although, it wouldn't be a 1:1 replacement for production Postgres, but should be enough to verify functionality
2. Also, as @chrisknoll mentioned, would be nice to establish database population approach which we used in Arachne via `com.github.springtestdbunit`. E.g.: https://github.com/OHDSI/ArachneCentralAPI/blob/develop/src/test/java/com/odysseusinc/arachne/portal/api/v1/controller/AnalysisControllerTests.java

aklochkova:
@pavgra about p.1: 
I tried 2 in-memory dbs -- H2db and HSQLDB, both in PostgreSql-compatible mode, but they fail during flyway migrations due to not fully syntax support. And I can't disable Flyway in config because it is integrated in java code.

pavgra:
@aklochkova , could you log the list of not supported (failing) expressions?

chrisknoll:
@aklochkova,
I"m running through the same hurdles as you with setting up some test cases for circe, and I'm starting with in-menory DB of H2 in postgresql compatibility.  I still way behind your competency level with the testing framework, but here's my suggestion about getting the in-memory H2 working:

- Copy the contents of `resources/db/migration/postgresql` into a new folder `/db/migration/h2`.
- Adjust your test profile to specify `classpath:db/migration/h2` as the `flyway.locations` property.

Then try to migrate and adjust the non-supported expressions (the list that @pavgra requested will be the one).  I'm happy to help with this if you want to set up this groundwork in your branch, and I can look into fixes as well.

The pain of this is that we'l have 'yet another dialect' to support for migrations, and I know that's painful, however we're planning on dropping support for everything except postgresql, so that will just leave it to be h2 and postgresql when we get to that point (if we want to maintain a h2 set of migrations)

Other thing I'll add to this thread:  I've been very resistant to adding additional dependencies on the dev cycle that would over-complicate the setup (such as depending on a message queue service to be installed, or have docker infrastructure).  However, after looking at [TestContainers](https://www.testcontainers.org/) , I'm ready to reverse that position  I'm going to be working on 2 approaches in circe: a H2 (postgresql compatability mode) branch, and one that leverages test containers PostgreSQL wrappers.  I think the best approach would be to use full-fledged postgresql instances in our DB testing and not 'emulators' like H2, but I was worried about the additional work of spinnign up instances, making tests completely encapsulated, etc.  But, after reading up on it, I think DBUNIT + Postgresql testcontainers will be worth the additional headache of getting a docker environment configured for local developers (and, to be clear, with the BIOS-level virtualization support and Hyper-V dependencies that Docker requires, it _is_ a headache, but it only really impacts executing tests, which I think we can manage).  @anthonysena : I did plan on discussing these points at a future WebAPI WG call when I have enough information to present.

pavgra:
@chrisknoll, I am very concerned with both options: either another set of migrations which would require additional effort for extending and managing or Docker which limits a lot of Windows-based developers. Of course, if we figure out that there are no other options, we don't have a choice. But to figure this out, @aklochkova , please provide the list of the expressions non conforming H2 dialect as was requested, and we'll continue the discussion from there.

chrisknoll:
> managing or Docker which limits a lot of Windows-based developers.

With build 1809 of Windows 10, there's 'process isolation' options available that does not require hyper-V.  I also have a powershell script which will download Docker Engine as a service (I am not up to 1809 on my Win10 workstation, but theoretically it will work).  

Read about it [here](https://www.kauffmann.nl/2019/03/04/how-to-install-docker-on-windows-10-without-hyper-v/). 

That's all I have to say on the Docker aspect.

pavgra:
I went through a bunch of constructs which were failing in H2. Applied is the patch with the changes:
[Adopt_Postgres_migrations_to_H2.patch.zip](https://github.com/OHDSI/WebAPI/files/3536767/Adopt_Postgres_migrations_to_H2.patch.zip)

But I was not able to workaround lack of support of `setval` function which we used for tweaking sequences. And the function cannot be replaced with `ALTER SEQUENCE ... RESTART WITH ...` because we calculate the new values on the fly

pavgra:
Please follow the testing approach defined in https://github.com/OHDSI/WebAPI/pull/1318/files

src/test/java/org/ohdsi/webapi/test/entity/TestCopy.java
```diff
@@ -0,0 +1,68 @@
+package org.ohdsi.webapi.test.entity;
+
+import static org.junit.Assert.assertEquals;
+import static org.ohdsi.webapi.test.TestConstants.COPY_PREFIX;
+import static org.ohdsi.webapi.test.TestConstants.NEW_TEST_ENTITY;
+
+import junitparams.Parameters;
+import org.junit.Test;
+
+public abstract class TestCopy extends TestCreate {
```
pavgra: Does the `TestCopy` inherit `TestCreate` from a logical / semantical perspective? If yes - how & why? If no - why would you use inheritance instead of composition (e.g. via interfaces with default methods)?

src/test/java/org/ohdsi/webapi/test/entity/TestImport.java
```diff
@@ -0,0 +1,95 @@
+package org.ohdsi.webapi.test.entity;
+
+import static org.junit.Assert.assertEquals;
+import static org.ohdsi.webapi.test.TestConstants.NEW_TEST_ENTITY;
+import static org.ohdsi.webapi.test.TestConstants.SOME_UNIQUE_TEST_NAME;
+
+import org.junit.Test;
+
+public abstract class TestImport extends TestCopy {
```
pavgra: The same question as for the `TestCopy.java`

src/test/java/org/ohdsi/webapi/test/entity/TestImport.java
```diff
@@ -0,0 +1,95 @@
...
+
+    protected abstract Integer getDtoId(Object dto);
```
pavgra: Is there really a separate logic required for each type? Shouldn't these DTOs implement something like `CommonEntityDTO` altogether?

src/test/java/org/ohdsi/webapi/test/entity/TestImport.java
```diff
@@ -0,0 +1,95 @@
...
+
+    protected abstract Object getEntity(int id);
```
pavgra: Why do you use `Object`-s and casts everywhere? Why don't you use generics?

src/test/java/org/ohdsi/webapi/test/entity/TestCreate.java
```diff
@@ -0,0 +1,53 @@
+package org.ohdsi.webapi.test.entity;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.ohdsi.webapi.test.TestConstants.NEW_TEST_ENTITY;
+
+import junitparams.JUnitParamsRunner;
```
pavgra: Where is an entry in pom.xml's dependencies for the JUnitParams?

src/test/java/org/ohdsi/webapi/test/entity/TestCopy.java
aklochkova: I can't do this via interfaces with default methods because `@SpringBootTest` and `@Test` annotations work only with classes. I agree that there is no logical sense in inheritance here but I don't see another way

chrisknoll:
Hello, everyone,
  I was browsing our open PRs (21!) and was trying to see where we could be focusing our attention on closing some of these out, and I came across this one.  

I recognize this PR has been very long-running, and there's bit a bit of back-and-forth about how these tests will be implemented...I believe it finally settled on an embedded PG instance.

The reason for my comment here, is that there's an open PR that we're planning on merging here: https://github.com/OHDSI/WebAPI/pull/1528, that may have some impact on the work done in this PR.

The primary change is how  the embedded database is instantiated and a small refactor to the AbstractInMemoryTest.java class (to use the embedded database in a class rule).  The primary issue I was having was that the database (via the `@ClassRule`) was being re-initialized (ie: new database ports, executing flyway scripts, and other setup operations for the spring container).  The chage in the PR makes the embedded instance a 'static instance' that only gets set up once, eliminating the need to `@DirtyContext` after each test (because the datasource environment vars are reset so the `@Ccomponents`need to be reset).  

The only thing this means for the individual tests is that they should reset the database state to a pre-test state that will allow the test to run.  This is what is being done in circe-be ([example](https://github.com/OHDSI/circe-be/blob/a9f12090270c3bffd46fd5a98e02ca48ccdec44a/src/test/java/org/ohdsi/circe/cohortdefinition/CohortGeneration_5_0_0_Test.java#L297)), and this is what I'm doing in the PathwayAnalysis tests ([example](https://github.com/OHDSI/WebAPI/blob/d928d1bc9b6d1ec3c3cc26b4ef1a6b3c2a23de62/src/test/java/org/ohdsi/webapi/pathway/PathwayAnalysisTest.java#L105)).

Doing it this way means that we don't have to wait for the DB or SpringContext to be completely reset between tests. This was saving me a lot of time when running my tests.  Note: i think there's still a need for `@DirtyContext` in tests:  in the case of SecurityEnabled, there's certain beans and filters that need to be reset so the simplest approach is to dirty the context.  But, I do want to avoid creating one database per test class.

I'd like to focus on getting this PR completed so that we can get better test coverage in WebAPI.  Please let me know if there's any way I can help with that effort.

src/test/java/org/ohdsi/webapi/test/WebApiIT.java
```diff
@@ -2,58 +2,56 @@
...
 @RunWith(SpringRunner.class)
-@SpringBootTest(classes = WebApi.class)
+@SpringBootTest(classes = WebApi.class, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @ActiveProfiles("test")
-@Ignore//do not run this test
-public abstract class WebApiIT {
-    
+@DbUnitConfiguration(databaseConnection = {"primaryDataSource"})
+@TestExecutionListeners({DependencyInjectionTestExecutionListener.class, DbUnitTestExecutionListener.class, DirtiesContextTestExecutionListener.class})
+@Ignore
```
YaroslavTir, line 28: Is this necessary to have ignore annotation here?

src/test/java/org/ohdsi/webapi/test/ITEntity.java
```diff
@@ -0,0 +1,66 @@
...
+        VocabularyServiceIT.class,
+        ConceptSetEntity.class,
+        CohortDefinitionEntity.class,
+        CCEntity.class,
+        IREntity.class,
+        PathwayEntity.class,
+        EstimationEntity.class,
+        PredictionEntity.class        
+})
+public class ITEntity extends AbstractShiroTest {
```
YaroslavTir, line 35: could not run tests get an exception:
Message    : Connection to 54.209.111.128:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.

src/test/java/org/ohdsi/webapi/test/entity/ConceptSetEntity.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Before
+    @Override
+    public void init() throws Exception {
+
+        TestCreate.super.init();
+    }
+
+    //region test methods
```
YaroslavTir, line 46: I don't see the advantage of these region comments, it is quite clear that all those tests inherited from an interface.

src/test/java/org/ohdsi/webapi/test/WebApiIT.java
aklochkova: no, it's old code, will remove it

chrisknoll:
Other than the build warning related to the version of DBUnit, this PR looks good (we can tackle some of the other issues later) so this PR can merge in whenever you are ready.

chrisknoll:
Other than the build warning related to the version of DBUnit, this PR looks good (we can tackle some of the other issues later) so this PR can merge in whenever you are ready.

aklochkova:
@chrisknoll please note I've fixed `dbunit` dependency
I think it would be better to fix other tests in a separate task because initially this one was just about adding new tests and not refactoring the whole bunch

chrisknoll:
Fine with me! I have no further comments. Thank you all very much for your hard work.

PR #1188, opened 2019-06-07
Fixes Characterization generation fails with custom features

wivern:
Fixes OHDSI/Atlas#1869


anthonysena:
@wivern - left a note on the original issue: https://github.com/OHDSI/Atlas/issues/1869#issuecomment-499958133. Just wanted to see if you had reproduced the same database connection issue on your side during testing?

PR #1193, opened 2019-06-10
move transactional to implementation

AntonStepanof:
fixes https://github.com/OHDSI/Atlas/issues/1879

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
+++ b/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
@@ -303,11 +303,13 @@ public class IRAnalysisService extends AbstractDaoService implements GeneratesNo
   }
 
   @Override
+  @Transactional
   public int getCountIRWithSameName(final int id, String name) {
     return irAnalysisRepository.getCountIRWithSameName(id, name);
   }
 
   @Override
+  @Transactional
   public IRAnalysisDTO createAnalysis(IRAnalysisDTO analysis) {
     Date currentTime = Calendar.getInstance().getTime();
 
@@ -334,6 +336,7 @@ public class IRAnalysisService extends AbstractDaoService implements GeneratesNo
   }
 
   @Override
+  @Transactional(readOnly = true)
   public IRAnalysisDTO getAnalysis(final int id) {
 
     return getTransactionTemplate().execute(transactionStatus -> {
@@ -390,6 +393,7 @@ public class IRAnalysisService extends AbstractDaoService implements GeneratesNo
     }
 
     @Override
+    @Transactional
   public IRAnalysisDTO saveAnalysis(final int id, IRAnalysisDTO analysis) {
     Date currentTime = Calendar.getInstance().getTime();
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -572,6 +580,7 @@ public IRAnalysisDTO copy(final int id) {
 
 
   @Override
+  @Transactional
   public Response export(final int id) {
```
wivern: I'd like to suggest add source access checking and filtering during export.

wivern: In addition to above comment, there is an issue when some of sources became unavailable. During such case export analysis to CSV also becomes unavailable. Probably, it's better to export results for working sources and notify a user that some sources doesn't responsible.

AntonStepanof: I added check for availability sources

AntonStepanof: also added check for permissions

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -785,4 +799,19 @@ private void fillCohortIds(List<Integer> ids, List<CohortDTO> cohortDTOS) {
...
+
+  private boolean isSourceAvailable(Source source) {
+    boolean sourceAvailable = true;
+    try {
+      sourceService.checkConnection(source.getSourceKey());
+    } catch (Exception e) {
+      log.error("cannot get connection to source with key {}", source.getSourceKey(), e);
+      sourceAvailable = false;
+    }
+    if (!sourceAccessor.hasAccess(source)) {
```
wivern: Not needed to detect connection availability when source access is prohibited, don't it?

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -596,6 +605,9 @@ public Response export(final int id) {
       for (ExecutionInfo execution : executions)
       {
         Source source = execution.getSource();
+        if (!isSourceAvailable(source)) {
```
wivern: But why not simply filter out executions with unaccessible source?
Smth like `executions.stream().filter(e -> this.isSourceAvailable(e.getSource()).collect(...)` ?

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
AntonStepanof: yes, fixed

AntonStepanof: yes, updated

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
AntonStepanof: updated

wivern:
dceab168-aded-4d78-b98c-4bf7a79ed0b8
idea

PR #1195, opened 2019-06-11
Revise concept path in cohort results queries

anthonysena:
Per OHDSI/Atlas#1872, I've noted that the queries that are used to retrieve the cohort reporting results are using the `concept_hierarchy.level4_concept_name` column from the `concept_hierarchy` table. However, this column is not populated for the following:

1. Drug Era: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/ddl/results/init_concept_hierarchy.sql#L144
2. Measurement: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/ddl/results/init_concept_hierarchy.sql#L205
3. Observation: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/ddl/results/init_concept_hierarchy.sql#L230
4. Procedure: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/ddl/results/init_concept_hierarchy.sql#L255

So, I've removed (commented out) the `concept_hierarchy.level4_concept_name` references from the corresponding cohort results queries so that we eliminate the "NA" value that was at the beginning of the concept_path display as mentioned in OHDSI/Atlas#1872.

For `sqlProcedureTreemap.sql`, I noted that the `CONCAT()` operation was inconsistent in using `isNull` while the others use `coalesce` so I've adjusted this accordingly.

### DIFF
--- a/src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
+++ b/src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
@@ -1,6 +1,7 @@
 select  concept_hierarchy.concept_id,
 	CONCAT(
-		coalesce(concept_hierarchy.level4_concept_name,'NA'), '||',
+                -- Note: concept_hierarchy.level4_concept_name is null so removing from this query
+		--coalesce(concept_hierarchy.level4_concept_name,'NA'), '||', 
 		coalesce(concept_hierarchy.level3_concept_name,'NA'), '||',
 		coalesce(concept_hierarchy.level2_concept_name,'NA'), '||',
 		coalesce(concept_hierarchy.level1_concept_name,'NA'), '||',
--- a/src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
+++ b/src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
@@ -1,11 +1,11 @@
 select  concept_hierarchy.concept_id,
 	CONCAT(
-                -- Note: concept_hierarchy.level4_concept_name is null so removing from this query
-		--coalesce(concept_hierarchy.level4_concept_name,'NA'), '||', 
-		coalesce(concept_hierarchy.level3_concept_name,'NA'), '||',
-		coalesce(concept_hierarchy.level2_concept_name,'NA'), '||',
-		coalesce(concept_hierarchy.level1_concept_name,'NA'), '||',
-		coalesce(concept_hierarchy.concept_name,'NA')
+                -- Note: concept_hierarchy.level4_concept_name is null 
+                -- and so it is not included in this query
+		isNull(concept_hierarchy.level3_concept_name,'NA'), '||',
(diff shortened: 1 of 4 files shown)


src/main/resources/resources/cohortresults/sql/procedure/sqlProcedureTreemap.sql
```diff
@@ -1,9 +1,9 @@
 select  concept_hierarchy.concept_id,
 	CONCAT(
-	  isNull(concept_hierarchy.level3_concept_name,'NA'), '||',
-	  isNull(concept_hierarchy.level2_concept_name,'NA'), '||',
-	  isNull(concept_hierarchy.level1_concept_name,'NA'), '||',
-	  isNull(concept_hierarchy.concept_name,'NA')
+	  coalesce(concept_hierarchy.level3_concept_name,'NA'), '||',
```
pavgra: `isNull` should be used since SqlRender provides translation for it, not for `coalesce`. E.g. for BigQuery `isNull` replaced with `IFNULL`, for Oracle - `NVL`

src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
```diff
@@ -1,6 +1,7 @@
 select  concept_hierarchy.concept_id,
 	CONCAT(
-		coalesce(concept_hierarchy.level4_concept_name,'NA'), '||',
+                -- Note: concept_hierarchy.level4_concept_name is null so removing from this query
+		--coalesce(concept_hierarchy.level4_concept_name,'NA'), '||',
```
pavgra: Why do we need to keep commented out lines? The note is good but why keep code?

src/main/resources/resources/cohortresults/sql/procedure/sqlProcedureTreemap.sql
anthonysena: @pavgra are you suggesting that all treemap queries in this PR should all use isNull in place of coalesce?

src/main/resources/resources/cohortresults/sql/drugera/sqlDrugEraTreemap.sql
anthonysena: I'll remove the code and revise the comment so a reader can understand why level4 is not included

src/main/resources/resources/cohortresults/sql/procedure/sqlProcedureTreemap.sql
pavgra: I'd say that all `coalesce`-s should be replaced with `isNull` not only in this PR but everywhere for the reasons described above

anthonysena: Fair enough - I'd say let's open another issue if you'd prefer to see `isNull` used in place of `coalesce `across the code base. For the scope of this PR, I've revised the queries to use `isNull`

anthonysena:
@pavgra - I've updated to reflect your feedback. Please let me know if you have any further concerns. Thanks!

PR #1196, opened 2019-06-11
Ensure unique feature analysis names upon import - fixes #1192

### LINKED ISSUE #1192: Characterization import fails with custom features specified
Using this design: [cc-import-fails-with-custom-feature.txt](https://github.com/OHDSI/WebAPI/files/3267366/cc-import-fails-with-custom-feature.txt), attempt to import this design into an Atlas instance more than one time.

The second attempt will fail with the following error: 

> "Violation of UNIQUE KEY constraint 'uq_fe_name'. Cannot insert duplicate key in object 'dbo.fe_analysis'. The duplicate key value is (Age>18 & Gender=Male).",

Related to #1100

anthonysena:
Adds checks to ensure feature analysis names are not duplicated upon import of a cohort characterization design. This fixes issue #1192. I've tested this against SQL Server and PostgreSQL.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
@@ -1,18 +1,16 @@
 package org.ohdsi.webapi.cohortcharacterization;
 
-import org.ohdsi.webapi.cohortcharacterization.domain.CohortCharacterizationEntity;
+import java.util.List;
 
 public class CcImportEvent {
 
-    private CohortCharacterizationEntity entity;
+    private List<Integer> savedAnalysesIds;
 
-    public CcImportEvent(CohortCharacterizationEntity entity) {
-
-        this.entity = entity;
+    public CcImportEvent(List<Integer> savedAnalysesIds) {
+        this.savedAnalysesIds = savedAnalysesIds;
     }
 
-    public CohortCharacterizationEntity getEntity() {
-
-        return entity;
+    public List<Integer> getSavedAnalysesIds() {
+        return savedAnalysesIds;
     }
 }
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
@@ -4,6 +4,7 @@ import java.util.List;
 
 public class CcImportEvent {
 
+    // should keep list of ids to prevent error for duplication of permissions
     private List<Integer> savedAnalysesIds;
 
     public CcImportEvent(List<Integer> savedAnalysesIds) {
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
@@ -26,6 +26,7 @@ import org.ohdsi.webapi.common.DesignImportService;
 import org.ohdsi.webapi.common.generation.AnalysisGenerationInfoEntity;
 import org.ohdsi.webapi.common.generation.GenerationUtils;
 import org.ohdsi.webapi.feanalysis.FeAnalysisService;
+import org.ohdsi.webapi.feanalysis.domain.FeAnalysisCriteriaEntity;
 import org.ohdsi.webapi.feanalysis.domain.FeAnalysisEntity;
 import org.ohdsi.webapi.feanalysis.domain.FeAnalysisWithCriteriaEntity;
 import org.ohdsi.webapi.job.GeneratesNotification;
@@ -330,11 +331,11 @@ public class CcServiceImpl extends AbstractDaoService implements CcService, Gene
         updateStratas(entity, persistedCohortCharacterization);
 
         importCohorts(entity, persistedCohortCharacterization);
-        importAnalyses(entity, persistedCohortCharacterization);
+        List<Integer> savedAnalysesIds = importAnalyses(entity, persistedCohortCharacterization);
 
         final CohortCharacterizationEntity savedEntity = saveCc(persistedCohortCharacterization);
 
-        eventPublisher.publishEvent(new CcImportEvent(savedEntity));
+        eventPublisher.publishEvent(new CcImportEvent(savedAnalysesIds));
 
         return savedEntity;
     }
@@ -623,7 +624,8 @@ public class CcServiceImpl extends AbstractDaoService implements CcService, Gene
         stat.setMax(rs.getDouble("max_value"));
     }
 
-    private void importAnalyses(final CohortCharacterizationEntity entity, final CohortCharacterizationEntity persistedEntity) {
+    private List<Integer> importAnalyses(final CohortCharacterizationEntity entity, final CohortCharacterizationEntity persistedEntity) {
+        List<Integer> savedAnalysesIds = new ArrayList<>();
         final Map<String, FeAnalysisEntity> presetAnalysesMap = buildPresetAnalysisMap(entity);
 
         final Set<FeAnalysisEntity> entityAnalyses = new HashSet<>();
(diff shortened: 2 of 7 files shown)


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -631,12 +631,14 @@ private void importAnalyses(final CohortCharacterizationEntity entity, final Coh
             switch (analysis.getType()) {
                 case CRITERIA_SET:
                     FeAnalysisWithCriteriaEntity criteriaAnalysis = (FeAnalysisWithCriteriaEntity)analysis;
+                    criteriaAnalysis.setName(NameUtils.getNameWithSuffix(criteriaAnalysis.getName(), this::getFeNamesLike));
```
pavgra: I would add here the same logic as we have for Cohorts: first, check if such analysis already exists (search for analysis with the "design" field having the same value as imported analysis) and only then, if such analysis doesn't exist, create it. Otherwise, if analysis exists, do not duplicate entities

anthonysena: @pavgra - I've pushed up a commit https://github.com/OHDSI/WebAPI/pull/1196/commits/c9160cb200313a9ad78c1f6da9f1b61eaddfa426 which will address the CUSTOM_FE duplication. 

As discussed offline, handling the CRITERIA_SET type is a bit more complex since this data structure represents 1 or more entries in the `fe_analysis_criteria`. If someone from your team can look at this case and provide a solution, I'd be happy to review & test.

AntonStepanof: added logic for CRITERIA_SET

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -701,5 +724,13 @@ private void invalidateGenerations() {
             return null;
         });
     }
+    
+    private List<String> getFeNamesLike(String name) {
+        return analysisService.getNamesLike(name);
+    }
+    
+    private Optional<FeAnalysisEntity> getFeByDesignAndName(final FeAnalysisWithStringEntity withStringEntity, final String feAnalysisName) {
```
pavgra: What is the reason for the function? - why don't you reference `analysisService.findByDesignAndName` directly?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
```diff
@@ -1,18 +1,16 @@
 package org.ohdsi.webapi.cohortcharacterization;
 
-import org.ohdsi.webapi.cohortcharacterization.domain.CohortCharacterizationEntity;
+import java.util.List;
 
 public class CcImportEvent {
 
-    private CohortCharacterizationEntity entity;
+    private List<Integer> savedAnalysesIds;
```
pavgra: Why?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -622,29 +624,50 @@ private void gatherForDistribution(final CcDistributionStat stat, final ResultSe
...
 
         for (final FeAnalysisEntity analysis : entity.getFeatureAnalyses()) {
             switch (analysis.getType()) {
                 case CRITERIA_SET:
-                    FeAnalysisWithCriteriaEntity criteriaAnalysis = (FeAnalysisWithCriteriaEntity)analysis;
-                    entityAnalyses.add(analysisService.createCriteriaAnalysis(criteriaAnalysis));
+                    FeAnalysisWithCriteriaEntity<? extends FeAnalysisCriteriaEntity> criteriaAnalysis = (FeAnalysisWithCriteriaEntity) analysis;
+                    List<? extends FeAnalysisCriteriaEntity> design = criteriaAnalysis.getDesign();
+                    Optional<FeAnalysisEntity> entityCriteriaSet = analysisService
+                            .findByCriteriaListAndName(design, criteriaAnalysis.getName());
```
pavgra: Why do you include the name into the search? We do not do that for cohorts import

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
AntonStepanof: fixed

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
AntonStepanof: I keep list of saved ids for prevent this error: Can't create permission - it already exists
For case when analysis already present in db.

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
AntonStepanof: updated

src/main/java/org/ohdsi/webapi/feanalysis/FeAnalysisServiceImpl.java
```diff
@@ -174,6 +176,41 @@ public void deleteAnalysis(FeAnalysisEntity entity) {
...
+        List<String> currentList = curCriteriaList.stream().map(FeAnalysisCriteriaEntity::getExpressionString).collect(Collectors.toList());
+        List<String> newList = newCriteriaList.stream().map(FeAnalysisCriteriaEntity::getExpressionString).collect(Collectors.toList());
+        return CollectionUtils.isEqualCollection(currentList, newList);
+    }
+
+    private Optional<FeAnalysisEntity> findByDesignAndPredicate(final String design, final Predicate<FeAnalysisEntity> f) {
+        List<FeAnalysisEntity> detailsFromDb = analysisRepository.findByDesign(design);
+        return detailsFromDb
+                .stream()
+                .filter(v -> Objects.equals(v.getDesign(), design))
```
pavgra: In what case this expression won't return `true`?

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
pavgra: Could you put this comment into the code, please?

AntonStepanof: added comment

src/main/java/org/ohdsi/webapi/feanalysis/FeAnalysisServiceImpl.java
AntonStepanof: yes, I removed this filter

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcImportEvent.java
AntonStepanof: added comment

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -622,29 +624,49 @@ private void gatherForDistribution(final CcDistributionStat stat, final ResultSe
...
+                    }
                     break;
                 case PRESET:
                     entityAnalyses.add(presetAnalysesMap.get(analysis.getDesign()));
                     break;
                 case CUSTOM_FE:
-                    entityAnalyses.add(analysisService.createAnalysis(analysis));
+                    FeAnalysisWithStringEntity withStringEntity = (FeAnalysisWithStringEntity) analysis;
+                    Optional<FeAnalysisEntity> findAnalysisResult = analysisService.findByDesignAndName(withStringEntity, withStringEntity.getName());
+                    if (findAnalysisResult.isPresent()) {
```
pavgra: This `if` statement is very similar to the one in CRITERIA_SET case. Cannot we create a templated function for this?

AntonStepanof: updated

AntonStepanof: updated

PR #1204, opened 2019-06-20
Additional fixes SQL/Oracle for #1105

anthonysena:
Updates the Flyway migration script `V2.8.0.20190424150601__add-unique-name-constraint-to-entities.sql `for Oracle/SQL Server to use a `varchar(255)` otherwise the script can fall into an infinite loop.

This PR does not address a very relevant comment from @chrisknoll noted here: https://github.com/OHDSI/WebAPI/issues/1105#issuecomment-504029253 since I'm not sure if this will happen practically when migrating. That said, I fully support using the `LEFT` operator as proposed by Chris but could use some assistance in implementing this on Oracle/PostgreSQL.

Tagging @pavgra @aklochkova @chrisknoll for awareness and feedback.


anthonysena:
Per Chris' additional https://github.com/OHDSI/WebAPI/issues/1105#issuecomment-504056035, let's roll with this solution provided that there are no concerns with the approach taken in this PR. Thanks!

PR #1208, opened 2019-06-21
updates arachne-commons version

wivern:


anthonysena:
@pavgra @wivern - sorry to push a change directly to the v2.7.2-rc branch but the alpha reference is not on the OHDSI Nexus repo. Let's revisit how we want to get this component tagged for the final release of v2.7.2

PR #1211, opened 2019-06-21
Fix overlap logic.

chrisknoll:
This PR is intended for post 2.7.2 release, so it's being attached to the 2.8.0 backlog.

This PR incorporates the fix for identifying overlapping event cohorts contributed by @mick-iqvia and @hms1.  To those collaborators: if you could pull this branch and run your local tests to determine if it identifies the overlaps properly.

This PR contains the following changes:
- Allow event cohort periods that start within the target cohort era.  Prior version required the event cohort era to start and end in the target cohort era.
- When collapsing, ensure start-end dates span at least a day.
- Use count_big when counting cohort events and pathways.

Note about what was discussed on prior WG calls:  this PR does not incorporate the new option for 'minimum segment length', since this PR is just going to focus on correctly identifying overlaps.  The pieces are in place to support the minimum segment length options, and that will be handled in a separate PR.  This PR is structured this way as to be included in a 2.7.3 without needing any changes to Atlas UI (which will be required when we add the new 'minimum segment length' option).  We had discussed switching 'collapse window' with 'minimum segment length' but the collapse window has very specific behavior that are desirable for our own research teams, so we are going to support both options (with adequate documentation).

For an example of the collapse window/multiple overlap fixes, i ran our own pathway analysis of T2DM medications over a T2DM patient population, an this was the resulting sunburst:

image

As you can see, by specifying a collapse window of 99999 days, all events were collapse to the same start/end dates.  This shows how this population had up to 4 distinct event cohorts (demonstrated by a single band having 4 different color stripes in it (note: it is very small slice of the sunburst)) in the same path segment, demonstrating that the new logic for finding overlapping periods is working.  Additional review would be welcome if there is any corner-cases that may remain undiscovered that should be addressed.

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -21,7 +21,7 @@ FROM (
 	  e.cohort_end_date
 	FROM @target_cohort_table e
 	  JOIN ( @event_cohort_id_index_map ) ec ON e.cohort_definition_id = ec.cohort_definition_id
-	  JOIN @target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_end_date <= t.cohort_end_date AND t.subject_id = e.subject_id
+	  JOIN @target_cohort_table t ON t.cohort_start_date <= e.cohort_start_date AND e.cohort_start_date <= t.cohort_end_date AND t.subject_id = e.subject_id
 	WHERE t.cohort_definition_id = @pathway_target_cohort_id
 ) RE;
 
@@ -52,7 +52,7 @@ replacements AS (
   SELECT orig.subject_id, orig.cohort_date, FIRST_VALUE(cohort_date) OVER (PARTITION BY group_idx ORDER BY ordinal ASC ROWS UNBOUNDED PRECEDING) as replacement_date
   FROM grouped_dates orig
 )
-SELECT *
+SELECT subject_id, cohort_date, replacement_date
 INTO #date_replacements
 FROM replacements
 WHERE cohort_date <> replacement_date;
@@ -77,6 +77,11 @@ FROM #raw_events event
 ;
 
 /*
+The collapsed dates (or the raw event cohort dates) may have intervals where start == end, so these should be expanded to cover a minimum of 1 day
+*/
+update #collapsed_dates_events set cohort_end_date = dateadd(d,1,cohort_end_date) where cohort_start_date = cohort_end_date;
+
+/*
 Split partially overlapping events into a set of events which either do not overlap or fully overlap (for later GROUP BY start_date, end_date)
 
 e.g.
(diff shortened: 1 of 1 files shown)


mick-iqvia:
Hi Chris,  I have run the updated sql through the tests I had in place before, most do not match because of the adjustment of end date by 1 day, but the combo ids match and so I am not sure that that difference is important.

Where there is a material difference is for events that overlap by 1 day, the new treatment does not see this as an overlap whilst the old one does. This is probably intentional but I thought I would point it out.

For example this test, where target cohort id is 9
{code}
cohort_definition_id |subject_id |cohort_start_date |cohort_end_date
int                             |bigint       |date                      |date
9                               |200           |2015-01-01          |2015-01-03
1                               |200           |2015-01-01          |2015-01-02
2                               |200           |2015-01-02          |2015-01-03
{code}
previous ouput
{code}
pathway_analysis_generation_id |target_cohort_id |combo_id |subject_id |ordinal |cohort_start_date |cohort_end_date
bigint                         |int              |bigint   |bigint     |int     |date              |date
1                                |9                 |1        |200        |1       |2015-01-01        |2015-01-01
1                                |9                 |3        |200        |2       |2015-01-02        |2015-01-02
1                                |9                 |2        |200        |3       |2015-01-03        |2015-01-03
{code}
new
{code}
pathway_analysis_generation_id |target_cohort_id |combo_id |subject_id |ordinal |cohort_start_date |cohort_end_date
bigint                         |int              |bigint   |bigint     |int     |date              |date
1                              |9                |1        |200        |1       |2015-01-01        |2015-01-02
1                              |9                |2        |200        |2       |2015-01-02        |2015-01-03
{code}

chrisknoll:
Hi, @mick-iqvia ,
  Yes, that result of not identifying the 1 day overlap with that input was intended.   This goes back to our discussion about inclusive dates, but when I was working through the logic this is why I went with the result you see:

If you take a timeline and split it at a specific date (which happens when we have some other start/end date between the start and end of the timeline:

```
|-----------------X------------|
```

And at that point X, you split the timeline into two segments:
```
|-----------------|
                  |------------|
``` 

Which would look like the following data:
cohort_definition_id |subject_id |cohort_start_date |cohort_end_date
-|-|-|-
int                             |bigint       |date                      |date
9                               |200           |2015-01-01          |2016-01-01
1                               |200           |2015-01-01          |2015-07-01
2                               |200           |2015-07-01          |2016-01-01

You can see that a single timeline split on the date 2015-07-01 should not lead to an overlap with itself, so that's why this PR's implementation doesn't.

What would you say if I added a step to the algorithm that when it reads in the cohort dates, it adds 1 to all end dates, leading your input data set to be the following:

cohort_definition_id |subject_id |cohort_start_date |cohort_end_date
-|-|-|-
int                             |bigint       |date                      |date
9                               |200           |2015-01-01          |2015-01-04
1                               |200           |2015-01-01          |2015-01-03
2                               |200           |2015-01-02          |2015-01-04

This would yield an overlap of one day with the span 2015-01-02 -> 2015-01-03?

I can push up that change now, and if there's a problem with it, it can be rolled back.

mick-iqvia:
Hi @chrisknoll , 

I am not sure about the change you are suggesting, but please make the call. 

From my perspective I think we need to make the algorithm as simple as possible (conceptually) so that users can understand the behaviour. I think the main problem we have is that start and end dates are inclusive and collapse windows lead to events sharing start and end dates which imply lots of single day overlaps. 

However I think that as you are supporting people actualy using this feature so are best placed to make the decision based on whatever works best for them, and what you have done already seems pretty reasonable.

chrisknoll:
Hi, @mick-iqvia ,
  I've padded the dates so you should get the result you expect when no collapse window is set.

You are correct: the people we're supporting rely on the collapse window behavior to 'smooth' out  events that are within days of each other, so the intent is to prefer not seeing the overlaps, but now they will need to specify a collapse window > 1 in order to collapse those events that start/end on the same day in the raw data (which is what they are already doing, they heavily depend on the behavior of collapse window to get that result, so this should work).

I'm hoping that this latest change will suit your needs and maintain the capability that our user base is familiar with.

-Chris

mick-iqvia:
Thanks @chrisknoll.

I can confirm that the results or running my tests are now the same as they were before (aprt from end date increased by 1).

Just to be clear, I do not have a string opinion on which version you choose. Perhaps @hms1 has a view.

anthonysena:
Linking to #779. @chrisknoll @mick-iqvia: question - do we still need pull request #808? Just trying to connect the dots a bit here.

chrisknoll:
This PR will replace #808

anthonysena:
These changes are now incorporated into v2.7.3 (#1226) and master (#1287) so I'm going to close this out.

PR #1212, opened 2019-06-24
Updates version of SqlRender

wivern:
Updates SqlRender version to 1.6.2-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -492,7 +492,7 @@
     <dependency>
       <groupId>org.ohdsi.sql</groupId>
       <artifactId>SqlRender</artifactId>
-      <version>1.6.1-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -492,7 +492,7 @@
     <dependency>
       <groupId>org.ohdsi.sql</groupId>
       <artifactId>SqlRender</artifactId>
-      <version>1.6.1-SNAPSHOT</version>
+      <version>1.6.2-SNAPSHOT</version>
```
pavgra: Why snapshot?

wivern: Since this issue https://github.com/OHDSI/SqlRender/issues/174 fix goes to SNAPSHOT.

PR #1217, opened 2019-06-27
Backmerge v2.7.2-rc into master

anthonysena:
Merge in updates for 2.7.2 release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,10 +6,10 @@
   <groupId>org.ohdsi</groupId>
   <artifactId>WebAPI</artifactId>
   <packaging>war</packaging>
-  <version>2.7.2</version>
+  <version>2.8.0-SNAPSHOT</version>
   <name>WebAPI</name>
   <properties>
-    <webapi.version>2.7.2</webapi.version>
+    <webapi.version>2.8.0</webapi.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jersey.version>2.14</jersey.version>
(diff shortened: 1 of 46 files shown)


pom.xml
```diff
@@ -6,15 +6,14 @@
   <groupId>org.ohdsi</groupId>
   <artifactId>WebAPI</artifactId>
   <packaging>war</packaging>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>2.7.2</version>
```
pavgra: Why not `2.7.3-SNAPSHOT`?

chrisknoll: if this is going on master, it should probably be 2.8.0-SNAPSHOT. If we need to do a point release, we'll branch off the 2.7.2 release tag, and set the POM version accordingly.

anthonysena: Good catch here @pavgra. I agree with @chrisknoll here - putting 2.8.0-SNAPSHOT on master at this point makes sense. Should we need a 2.7.3, we'll branch from the 2.7.2 release tag. I've updated the PR accordingly. Thanks!

pavgra:
@anthonysena , seems like you've updated it only in one place while 2 are required

image

anthonysena:
Sorry for that - I've adjusted this so that the `webapi.version` is 2.8.0 while the `version` is 2.8.0-SNAPSHOT. The former is displayed through the /info endpoint and thought it would be confusing if the -SNAPSHOT reference was there. 

If you do the merge, please squash down these repetitive commits. Thanks

Issue #219, opened 2017-07-28
Cohort Definition List query joins to CohortDetails


chrisknoll:
### Expected behavior
Querying for cohort definition list should not include columns (especially large columns) that are not required for the cohort definition list view.

### Actual behavior
When querying the cohort definition list, the JPQL generated causes a inner join to CohortDefinitionDetails and pulls in the VARCHAR(MAX) column. This has negative performance impacts, and has even causes queries to hang on our EC2 sql server instances.

### Steps to reproduce behavior
1. Open console log in WebAPI
2. Open a browser and access /WebAPI/cohortdefinition
3. you will see a Hibernate: debug message with the query. Notice that the cohort_definition_details table is joined, and the expression column is returned.

chrisknoll:
Reopening issue due to regression.  Getting the cohort definition list is joining to the details table, which results in excessive bandwidth against the database.  The query reported when accessing /WebAPI/cohortdefinition:

```
select cohortdefi0_.id as id1_12_0_, cohortdefi1_.id as id1_13_1_, userentity2_.ID as ID1_49_2_, 
	userentity3_.ID as ID1_49_3_, cohortdefi0_.created_by_id as created_7_12_0_, cohortdefi0_.created_date as created_2_12_0_, 
	cohortdefi0_.modified_by_id as modified8_12_0_, cohortdefi0_.modified_date as modified3_12_0_, 
	cohortdefi0_.description as descript4_12_0_, cohortdefi0_.expression_type as expressi5_12_0_, 
	cohortdefi0_.name as name6_12_0_, cohortdefi1_.expression as expressi2_13_1_, 
	cohortdefi1_.hash_code as hash_cod3_13_1_, userentity2_.last_viewed_notifications_time as last_vie2_49_2_, 
	userentity2_.LOGIN as LOGIN3_49_2_, userentity2_.NAME as NAME4_49_2_, 
	userentity3_.last_viewed_notifications_time as last_vie2_49_3_, userentity3_.LOGIN as LOGIN3_49_3_, 
	userentity3_.NAME as NAME4_49_3_ 
from ohdsi.cohort_definition cohortdefi0_ 
inner join ohdsi.cohort_definition_details cohortdefi1_ on cohortdefi0_.id=cohortdefi1_.id 
left outer join ohdsi.SEC_USER userentity2_ on cohortdefi0_.created_by_id=userentity2_.ID 
[...]

PR #1226, opened 2019-07-09
[2.7.3] Fixes for issue #219 and #779 for rc2.7.3

chrisknoll:
Fixes for #219 and #779 for merge into rc2.7.3 branch.

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -8,9 +8,6 @@ WHERE pathway_analysis_generation_id = @generation_id AND target_cohort_id = @pa
 * SELECT 1 AS cohort_definition_id, 1 AS cohort_index UNION ALL ...
 */
 
-IF OBJECT_ID('tempdb..#raw_events', 'U') IS NOT NULL
-DROP TABLE #raw_events;
-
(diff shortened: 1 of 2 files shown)


src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -200,12 +175,12 @@ SELECT
   target_count.cnt AS target_cohort_count,
   pathway_count.cnt AS pathways_count
 FROM (
-  SELECT CAST(COUNT(*) AS INT) cnt
+  SELECT COUNT_BIG(*) cnt
```
anthonysena: I think we need to keep the CAST() here based on this change for Impala: 

https://github.com/OHDSI/WebAPI/commit/687d0123e7c56079c89635a44d934b9c98163626#diff-02c181aadb1a447251b457feccfa48eaR205

chrisknoll: Ok, I'll update that commit with the change.

anthonysena:
@chrisknoll - tested again and I'm facing an issue with a pathway analysis with > 1 target cohort in the definition. Here is the error I've received:

`There is already an object named '#combo_events' in the database.`

Likely an issue with this script: `src/main/resources/resources/pathway/runPathwayAnalysis.sql `

chrisknoll:
Good catch @anthonysena .  I've updated the PR and force-pushed the branch (after testing with multiple T's).  Please delete your local branch before pulling down the PR branch. I will also update the main PR for master branch with this update.

PR #1230, opened 2019-07-15
 lack client cache directives

ssuvorov-fls:
fixes https://odysseusdataservices.atlassian.net/browse/ATLP-672


src/main/java/org/ohdsi/webapi/shiro/management/FilterTemplates.java
```diff
@@ -69,6 +69,7 @@
     AUTHZ("authz"),
     CORS("cors"),
     SSL("ssl"),
+    NO_CACHE("no_cache"),
```
wivern: Let's don't mix snake_case with camelCase

Issue #1232, opened 2019-07-16
Exporting IR analysis results leads to net::ERR_CONTENT_DECODING_FAILED 500


chrisknoll:
Expected Result:
After generating an IR analysis, yous hould be albe to export the results as CSV.

Actual Results:
Clicking export button begins to export the data, but Chrome reports net::ERR_CONTENT_DECODING_FAILED 500.

Googling this error turns up that the encoding of the response doesn't match the type specified in the response.

The rest endpoint that handled /{id}/export has `@Produces(MediaType.APPLICATION_JSON)`, which I beleive should be @Produces(MediaType.APPLICATION_OCTET_STREAM), since the response is an octet stream.

What is interesting about this is that while I can reliably reproduce this on our current production env (2.7.2), I'm unable to reproduce this problem (even with the apparently wrong MediType.APPLICATION_JSON) working off the 2.7.3 RC branch.  I'll have to dig into the logs on the prod server to see exactly what is going on.

YaroslavTir:
@chrisknoll You are right, it should not be an application-json media type. I guess the reason why it works is that we set this type programmatically in the method:
```
      response = Response
        .ok(baos)
        .type(MediaType.APPLICATION_OCTET_STREAM)
```
tbh I even can not reproduce this issue for version 2.7.2. I would remove this annotation, or put there correct type.

PR #1245, opened 2019-07-23
Remove produces annotation + minor improvements (#1232)

YaroslavTir:
`export` method was annotated with wrong media-type (MediaType.APPLICATION_JSON). We put correct media type inside of the method programmatically, so I just remove this redundant annotation

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
+++ b/src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
@@ -19,7 +19,9 @@ import com.cosium.spring.data.jpa.entity.graph.domain.EntityGraph;
 import com.cosium.spring.data.jpa.entity.graph.domain.EntityGraphUtils;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableMap;
 import com.opencsv.CSVWriter;
+import javax.ws.rs.core.HttpHeaders;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.IterableUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -584,11 +586,9 @@ public class IRAnalysisService extends AbstractDaoService implements GeneratesNo
   public Response export(final int id) {
 
     Response response = null;
-    HashMap<String, String> fileList = new HashMap<>();
-    HashMap<Integer, String> distTypeLookup = new HashMap<>();
 
-    distTypeLookup.put(1, "TAR");
-    distTypeLookup.put(2, "TTO");
+    Map<String, String> fileList = new HashMap<>();
+    Map<Integer, String> distTypeLookup = ImmutableMap.of(1, "TAR", 2, "TTO");
 
     try {
       IncidenceRateAnalysis analysis = this.irAnalysisRepository.findOne(id);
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -584,11 +586,9 @@ public IRAnalysisDTO copy(final int id) {
   public Response export(final int id) {
 
     Response response = null;
-    HashMap<String, String> fileList = new HashMap<>();
-    HashMap<Integer, String> distTypeLookup = new HashMap<>();
 
-    distTypeLookup.put(1, "TAR");
-    distTypeLookup.put(2, "TTO");
+    Map<String, String> fileList = new HashMap<>();
+    Map<Integer, String> distTypeLookup = ImmutableMap.of(1, "TAR", 2, "TTO");
```
chrisknoll: I don't see the benefit here of incorporating a new dependency on com.google.common.collect.ImmutableMap in order to do `ImmutableMap.of`  This is also less readable:

```
map.put(1,"Abc");
map.put(2,"bcd");
map.put(3,"cde");
map.put(4,"efg");
```

vs;

```
map = ImmutableMap.of(1,"Abc", 2,"bcd", 3,"cde", 4,"efg");
```

But I'm sure the 'readability' argument is a matter of style.  I don't think we need to change anything to ImmutableMap for purposes of addressing the `MediaType` annotation.

YaroslavTir: I agree with you that it does not have anything with MediaType, but I guess this is a good practice to make the code better with each commit. From my point of view, readability is one of the main aspects for the code and one line `ImmutableMap.of()` is better than multiline code. Developers of JDK also think that this way of initialization of the map is a good approach so added `Map.of("key1", "value")` method for Java 9, but we use Java 8, so guava is good alternative, and this dependency is already present in the project(ImmutableMap is the part of the guava) so I did not add a new dependency.

chrisknoll: > readability is one of the main aspects for the code and one line ImmutableMap.of() is better than multiline code.

I guess we can agree to disagree here:  I think if I never used newlines in all o fthe code so every file was a single line, you would not agree that it would make the code more 'readable'. My point above is that I can see 4 lines of code, and immediately recognize 4 values in the map, vs. i fyou put it into one line of code, it's not really obvious how many keys are we dealing with. Maybe  more concrete example:
```
map.put(1,"Here is some test, with commas, making it harder to pick out each value from a single line.");
map.put(2,"3,4, value, ,5, another value,6,7,8,9,10 example with commas and values, can you tell where the key/values split?");
map.put(3,"cde");
map.put(4,"efg");
```

vs:

```
map = ImmutableMap.of(1,"Here is some test, with commas, making it harder to pick out each value from a single line.", 2,"3,4, value, ,5, another value,6,7,8,9,10 example with commas and values, can you tell where the key/values split?", 3,"cde", 4,"efg");
```

Are you sure you want to make the claim that single-line is more readable?

Maybe if it was structured like this:

```
map = ImmutableMap.of(
  1,"Here is some test, with commas, making it harder to pick out each value from a single line.",
  2,"3,4, value, ,5, another value,6,7,8,9,10 example with commas and values, can you tell where the key/values split?", 
  3,"cde", 
  4,"efg"
);
```

That'd be much more readable, IMO.

YaroslavTir: `map = ImmutableMap.of(1,"Here is some test, with commas, making it harder to pick out each value from a single line.", 2,"3,4, value, ,5, another value,6,7,8,9,10 example with commas and values, can you tell where the key/values split?", 3,"cde", 4,"efg");`
yeah, This line can break anybody's brain. This is a good example, and of course, one line is not appropriate for this particular example, so it should be used wisely.

chrisknoll: > Developers of JDK also think that this way of initialization of the map is a good approach so added Map.of("key1", "value") method for Java 9

Funny, I actually looked this up: https://docs.oracle.com/javase/9/docs/api/java/util/Map.html

The way they implemetned, is that they have an overloadded method that handles up to 10 initializations.  The JDK devs are smoking something.  Personally, I would have designed the interface as fluent:
``
map.put(a,b).put(c,d).put(d.e)...
``

Personally, I woudln't point to what the JDK devs did for this interface as evidence that it's a smarter/improved approach.

Just my 2 cents!

pavgra:
fixed https://github.com/OHDSI/WebAPI/issues/1232

PR #1249, opened 2019-07-24
Heracles reports execution fails for impala (#1214)

### LINKED ISSUE #1214: Heracles reports execution fails for impala
Report "Conditions by index" - Treemap
Sql script and error message see in attachment
[impala.log](https://github.com/OHDSI/WebAPI/files/3324948/impala.log)

YaroslavTir:
Fixes #1214 
There is only regexp_like function for STRING arguments, but subquery returns INT value type, so I just cast it to the STRING


wivern:
We always have to keep in mind that queries should run on different DBMS.
I see cast to 'STRING'  data type in this PR but SqlServer doesn't provide such data type.
Basically queries written on the T-SQL dialect and translates to other dialects during execution with the magic of SqlRender library.

src/main/resources/resources/cohortresults/sql/cohortSpecific/byConcept/firstConditionRelativeToIndex.sql
```diff
@@ -24,7 +24,7 @@ FROM   (SELECT cohort_definition_id,
                               ORDER BY -1* CAST(stratum_1 AS INT)*30 ASC) AS
               count_value 
                    FROM  (
-                                             select stratum_1, max(count_value) as count_value
+                                             select CAST(stratum_1 AS STRING) AS stratum_1, max(count_value) as count_value
```
wivern: The original error text is `No matching function with signature: regexp_like(BIGINT, STRING)`
I guess this is somewhere near of `IS_NUMERIC` since Impala doesn't provide is_number function one's translated to the `regexp_like` which expects text value as an argument. 
It's better to apply cast directly to the function argument rather than using implicit cast from subquery.

YaroslavTir: yeah, you are right about String, I have found that in such cases we are using  varchar(11). 
` No matching function with signature: regexp_like(BIGINT, STRING)`
actually the original issue lay on the fact that we use `row_number()` `(1)` and `isNumeric()` `(2)` as stratum_1, both of them are not **string**, but then we have a numeric condition `(3)` for this field, so I would keep **int** there. Then we need **string** type for `regexp_like` function, so I cast it in (0)
```
select CAST(stratum_1 as STRING) as  stratum_1 -- (0) the outer query use this value for regexp_like, so it should be a STRING
from (select row_number() over (order by analysis_id) as stratum_1, 0 as count_value -- (1) int
      select CAST(isNumeric(stratum_1)  AS INT) as stratum_1, count_value int (ac) -- (2) int  (we kind of cast numeric to int)
     ) t1
where stratum_1 <= 10  -- (3) need int here
   or count_value > 0
group by stratum_1
```
But there is another question, since it always int, why do we need `regexp_like(stratum_1,'^([0-9]+\.?[0-9]*|\.[0-9]+)$')`?  the null check will be more then enough in this case.

wivern: The key word is `in this case` but there are some cases when null check isn't enough

PR #1253, opened 2019-07-29
kerberos auth support

### LINKED ISSUE #1240: Kerberos auth method has been lost
During merge of Atlas+ into community Atlas a Kerberos Auth Realm has been introduced:
https://github.com/OHDSI/WebAPI/pull/316/files#diff-cda19cf522452bd30c81e6333da1944d

But today the class is not presented in the codebase. Seems like it has been lost later during development. Therefore, there is a need to bring it back and **provide an example configuration** for how to setup it at **both local machine and server**

ssuvorov-fls:
fixes #1240


README.md
```diff
@@ -27,4 +27,51 @@ Documentation can be found a the [Web API Installation Guide](https://github.com
 ## License
 OHDSI WebAPI is licensed under Apache License 2.0
 
+## Kerberos
```
pavgra: I would not put such specific instructions into the main README. Why not create README in the `realms` or `shiro` package?

ssuvorov-fls: done

PR #1258, opened 2019-07-30
Made FoldingKey unique

aklochkova:
fixes https://github.com/OHDSI/Atlas/issues/1820


pom.xml
```diff
@@ -681,11 +681,6 @@
       <artifactId>arachne-common-utils</artifactId>
       <version>${arachne.version}</version>
     </dependency>
-    <dependency>
```
pavgra: Why have you removed the dep?

aklochkova: it is incorrectly duplicated, the original is on 709 line

pavgra: image

aklochkova: https://github.com/OHDSI/WebAPI/blob/447c2f5955c1348c9ca3f0837947d47ad9521e68/pom.xml#L709

aklochkova:
I noticed that after refresh jobs with the same folding key disappear.
Here you can see that Cohort 2 was generated and has terminated statuses in all sources (column 'Generated' is filled) but on Notifications screen statuses are wrong:
Screenshot_1
After refresh the block with Cohort 2 disappeared:
Screenshot_2
Cohort 2 and Pathway Analysis 2 have the same folding key.

And after my fix statuses are updated correct and after refresh the data about Cohort 2 remains:
Screenshot_3

pavgra:
@aklochkova , there is a time gap between analysis, e.g. cohort, being completed (see background with `Completed`) and notifications being updated (see popup with `Starting`):
image

anthonysena:
@olga-ganina - assigning you since we need to make sure that the comments here from @pavgra have been addressed. We need to test this with https://github.com/OHDSI/Atlas/pull/1989 which is also under review and will assign you so you can get this out to your team.

Issue #1268, opened 2019-07-31
Execution Engine jobs are not invalidated automatically


konstjar:
We have property
`<execution.invalidation.period>600000</execution.invalidation.period>`
that is not used at the moment.
It is expected that PLE/PLP execution jobs should be automatically invalidated after long time.

anthonysena:
This was closed with the 2.7.3 release.

PR #1272, opened 2019-08-01
Execution Engine jobs are not invalidated automatically (#1268)

YaroslavTir:
Fixes https://github.com/OHDSI/WebAPI/issues/1268
- add scheduler class ExecutionEngineAnalysisStatusInvalidationScheduler, that is active only in case `execution.invalidation.maxage` and `execution.invalidation.period` are set
- small improvements

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,7 @@
     <arachne.version>1.15.0-SNAPSHOT</arachne.version>
     <jersey-media-multipart.version>2.25.1</jersey-media-multipart.version>
     <execution.invalidation.period>600000</execution.invalidation.period>
+    <execution.invalidation.maxage>4</execution.invalidation.maxage>
     <execution.status.period>10000</execution.status.period>
 		<executionengine.url>https://localhost:8888/api/v1/analyze</executionengine.url>
 		<executionengine.token>Basic YWRtaW5Ab2R5c3NldXNpbmMuY29tOnBhc3N3b3Jk</executionengine.token>
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <arachne.version>1.15.0-SNAPSHOT</arachne.version>
     <jersey-media-multipart.version>2.25.1</jersey-media-multipart.version>
     <execution.invalidation.period>600000</execution.invalidation.period>
-    <execution.invalidation.maxage>4</execution.invalidation.maxage>
+    <execution.invalidation.maxage>12</execution.invalidation.maxage>
     <execution.status.period>10000</execution.status.period>
 		<executionengine.url>https://localhost:8888/api/v1/analyze</executionengine.url>
 		<executionengine.token>Basic YWRtaW5Ab2R5c3NldXNpbmMuY29tOnBhc3N3b3Jk</executionengine.token>
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
     <arachne.version>1.15.0-SNAPSHOT</arachne.version>
     <jersey-media-multipart.version>2.25.1</jersey-media-multipart.version>
     <execution.invalidation.period>600000</execution.invalidation.period>
-    <execution.invalidation.maxage>12</execution.invalidation.maxage>
+    <execution.invalidation.maxage.hours>12</execution.invalidation.maxage.hours>
     <execution.status.period>10000</execution.status.period>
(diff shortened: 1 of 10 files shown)


YaroslavTir:
> When you click on FAILED status of invalidated PLP execution, popup doesn't arise. Should be
> image

indeed, Only status was invalidated, but not a job. I fixed it with the last commit.

pavgra:
1. First, scheduler calls `invalidateExecutions -> jobInvalidator.invalidateJobExecutionById -> jobRepository.update`
2. `exit_code = Failed`  together with `exit_message = Invalidated by system` are properly persisted in database
3. Breakpoint in `org.springframework.batch.core.job.AbstractJob` line 341 (`finally` section of job execution) show exitStatus with empty message

src/main/java/org/ohdsi/webapi/JobInvalidator.java
```diff
@@ -1,19 +1,22 @@
...
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.support.TransactionTemplate;
 
-import javax.annotation.PostConstruct;
-import java.util.Calendar;
-
 @Component
 public class JobInvalidator {
 
+    public static final String INVALIDATED_BY_SYSTEM_EXIT_CODE = "Invalidated by system";
```
chrisknoll: Is this a 'code' or a message?

pavgra: Message. So I agree, would be good to rename the variable

pom.xml
```diff
@@ -143,6 +143,7 @@
     <arachne.version>1.15.0-SNAPSHOT</arachne.version>
     <jersey-media-multipart.version>2.25.1</jersey-media-multipart.version>
     <execution.invalidation.period>600000</execution.invalidation.period>
+    <execution.invalidation.maxage>4</execution.invalidation.maxage>
```
wivern: Could we provide some note about which units should be used for maxage?

wivern: As to me the 4 hours is to low to set as default, especially for estimation analysis. Could we increase it to 12 at least?

src/main/java/org/ohdsi/webapi/executionengine/service/ScriptExecutionServiceImpl.java
```diff
@@ -81,7 +79,15 @@
...
 
-    private List<ExecutionEngineAnalysisStatus.Status> FINAL_STATUES = ImmutableList.of(ExecutionEngineAnalysisStatus.Status.COMPLETED, ExecutionEngineAnalysisStatus.Status.COMPLETED);
+    private static List<ExecutionEngineAnalysisStatus.Status> INVALIDATE_STATUSES = ImmutableList.of(
+            ExecutionEngineAnalysisStatus.Status.RUNNING,
+            ExecutionEngineAnalysisStatus.Status.STARTED,
+            ExecutionEngineAnalysisStatus.Status.PENDING
+    );
+
+    private static List<ExecutionEngineAnalysisStatus.Status> FINAL_STATUES = ImmutableList.of(
+            ExecutionEngineAnalysisStatus.Status.COMPLETED,
```
wivern: I guess it should be FAILED here

src/main/java/org/ohdsi/webapi/JobInvalidator.java
```diff
@@ -29,15 +32,23 @@ public JobInvalidator(JobExplorer explorer, JobRepository repository, Transactio
...
             return null;
         });
     }
+
+    public void invalidateJobExecutionById(ExecutionEngineAnalysisStatus executionEngineAnalysisStatus) {
+        JobExecution job = jobExplorer.getJobExecution(executionEngineAnalysisStatus.getExecutionEngineGeneration().getId());
+        invalidationJobExecution(job);
+    }
+
+    private void invalidationJobExecution(JobExecution job) {
```
wivern: This is similar to the `CcServiceImpl.invalidateGenerations`, could we provide some code reuse?
E.g. via `AbstractDaoService`

pom.xml
YaroslavTir: Good point, this value used to be 4 hours, so I just return it.

src/main/java/org/ohdsi/webapi/JobInvalidator.java
YaroslavTir: good catch.

Issue #639, opened 2018-10-09
Owner of an entity should be able to grant permissions for the entity update / delete / etc to others


pavgra:
- Entity owner field should be introduced (I propose to put it into CommonEntity)
- There should be an additional button for Permissions management for each entity next to the entity's Save / Close / Delete button (let's move the button block to common component)
- The button should open permissions management popup, where:
  - Current owner should be displayed and the ownership can be transferred
  - Entity permissions (update / delete / execute / etc) can be granted to any Atlas user or Role. The block should consist of autocomplete for users, select for available entity-level permissions and list of permissions granted to users. Similarly to following Arachne popup (but with permission selector):
![image](https://user-images.githubusercontent.com/10010071/46640598-07d97180-cb3b-11e8-94f9-a253539fad4a.png)

fdefalco:
Bumping to 2.8 after discussion with @anthonysena - this is a feature that will be necessary for collaboration to be feasible in our environment

PR #1273, opened 2019-08-01
Adds ability for regular users to grant and revoke write permissions for their owned assets

pavgra:
Fixes https://github.com/OHDSI/WebAPI/issues/639


src/main/java/org/ohdsi/webapi/shiro/Entities/RoleRepository.java
```diff
@@ -2,20 +2,36 @@
...
+
+  @Query(
+    "SELECT r " +
+    "FROM RoleEntity r " +
+    "JOIN RolePermissionEntity rp ON r.id = rp.role.id " +
+    "JOIN PermissionEntity p ON rp.permission.id = p.id " +
+    "WHERE p.value IN :permissions " +
+    "GROUP BY r.id, r.name " +
+    "HAVING COUNT(p.value) = :permissionCnt"
+  )
```
wivern: Failed with:
```
com.microsoft.sqlserver.jdbc.SQLServerException: Column 'ohdsi.SEC_ROLE.system_role' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.
```

src/main/java/org/ohdsi/webapi/security/PermissionService.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public void removePermissionsFromRole(Map<String, String> permissionTemplates, Integer entityId, Long roleId) {
+
+        RoleEntity role = roleRepository.findById(roleId);
+        permissionTemplates.keySet()
+                .forEach(pt -> {
+                    String permission = getPermission(pt, entityId);
+                    PermissionEntity permissionEntity = permissionRepository.findByValueIgnoreCase(permission);
+                    if (permissionEntity != null) {
+                        RolePermissionEntity rp = rolePermissionRepository.findByRoleAndPermission(role, permissionEntity);
```
wivern: Why not to use `PermissionManager#removePermission(java.lang.Long, java.lang.Long)`?

pavgra: Because it deletes PermissionEntity-s and we want to delete RolePermissionEntity-s

Issue #1125, opened 2019-05-26
Application properties management


pavgra:
Today application properties are managed via pom.xml and the `application.properties` file works as a "proxy". I'd like to figure out what was the reasoning for the approach and suggest to migrate app configuration to `application.properties` and keep only dependencies in `pom.xml`.

@anthonysena , @fdefalco , @chrisknoll

chrisknoll:
Discussed here: http://forums.ohdsi.org/t/webapi-properties/219

I would prefer an approach that allows specifying application configuration settings independent of a build step.  In my experence, keeping the configuration separate from build means you are less likely to deploy QA settings into a Prod env or worse a PROD environment into a DEV env.

pavgra:
> I would prefer an approach that allows specifying application configuration settings independent of a build step

Let me confirm: so you are in support of extracting configuration properties from pom.xml and moving them into application.properties (application.yml), right?

chrisknoll:
I'm not clear on the tradeoffs of one approach vs. another. I'm just indicating that I would prefer that the `function` of our application configuration worked as:  each deployment environment has a set of configuration values that when the application starts up in that environment, it reads the application configuration local to that environment.  In this way, we do not bind the build to the environment, which in my experience leads to wrong configuration being deployed (usually leading to catastrophic results).

If you know of different options than what we use today (which currently is: pom.xml properties are copied out to application.properties, and application properties are read in by Spring to populate `@value` references), I'm sure we would all be happy to hear about what those other options are, and discuss the pros and cons of each approach.

pavgra:
> In this way, we do not bind the build to the environment

But neither you do today. Spring already `reads the application configuration local to that environment`. Or what I got wrong?

> If you know of different options than what we use today

I know that:
- Each thing should do its own function: Maven should resolve dependencies, `application.properties` should take care of app configuration. When responsibilities are mixed, things get messed, it is harder to understand and manage them
- Each tutorial / book / Spring documentation talks about application.properties as a primary way to manage app configuration

chrisknoll:
> But neither you do today. Spring already reads the application configuration local to that environment. Or what I got wrong?

[...]

PR #1297, opened 2019-08-15
add disable warming of CDM result cache property

YaroslavTir:
add cdm.result.cache.warming.disable property

### DIFF
--- a/src/main/resources/application-dev.properties
+++ b/src/main/resources/application-dev.properties
@@ -1,9 +1,0 @@
-security.provider=DisabledSecurity
-cdm.result.cache.warming.disable=true
-
-# Logging
-spring.jpa.show-sql=true
-spring.jpa.properties.hibernate.format_sql=true
-
-logging.level.org.hibernate.SQL=DEBUG
-logging.level.org.hibernate.type.descriptor.sql.BasicBinder=TRACE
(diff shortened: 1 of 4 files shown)


src/main/resources/application-dev.properties
```diff
@@ -0,0 +1,9 @@
+security.provider=DisabledSecurity
```
pavgra: I wouldn't do this. It is preferable to always have security on

pavgra: Also, I do not see such profile in pom.xml

src/main/resources/application.properties
```diff
@@ -205,3 +205,4 @@ hydra.externalPackage.prediction=${hydra.externalPackage.prediction}
 
 #Cache Config
 spring.cache.jcache.config=${spring.cache.jcache.config}
+cdm.result.cache.warming.disable=false
```
pavgra: Let's follow the currently existing pattern of props management (prop in `pom.xml` and proxy prop in `application.properties`). Although it may not look good, it would be even worse to have inconsistency. And we already planned application.properties refactoring - https://github.com/OHDSI/WebAPI/issues/1125

Btw, is it a good idea to have double inversion? `disabled=false`. Wouldn't it be easier to read `enabled` property?

YaroslavTir: > Let's follow the currently existing pa...
yeah, you are right.
> Btw, is it a good idea to have double inversion? disabled=false
hmm, yeah, I guess you are right. this looks some kind of peculiarly

PR #1304, opened 2019-08-21
Allow cache endpoint to be accessed in secured mode.

### LINKED ISSUE #1303: /cache/clear endpoint cannot be accessed in secure mode
No migrations were created which would define new permission and grant it to Atlas user role. Therefore, it is impossible to access the endpoint with enabled auth

chrisknoll:
Fixes #1303

### DIFF
--- a/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
@@ -103,6 +103,9 @@ public abstract class AtlasSecurity extends Security {
 
             .addRestPath("/permission/access/**/*", JWT_AUTHC) // Authorization check is done inside controller
 
+						//cache
+            .addRestPath("/cache/*")
+						
             .addProtectedRestPath("/**/*");
   }
 


chrisknoll:
After discussing on the call, the decision was that the group would like to change the functionality from being a un-restricted endpoint to a restricted endpoint.

@pavgra offered to add the migration scripts and permission assignments for this endpoint.

However, I believe he stated that he was going to add these new permissions assignments as part of 2.7.4 and a new UI elements for 2.8.  I feel this breaks semantic versioning.

@pavgra, I would urge you, regardless of any 'bad actor' behavior that could impact the performance of the application due to the unrestricted nature of this endpoint, to reconsider making functional changes (non-secured -> secured) as part of a patch-version update.  it's important that we respect the meaning of a semantic versioning, and the PR I'm putting forward here is in line with semantic versioning vs. what you are proposing.

pavgra:
Could you tell me how my proposal conflicts with the semver?

Currently:
- Non-secure mode: everyone can access
- Secure mode: doesn't work

With permission + role migration (for 2.7.4):
- Non-secure mode: everyone can access
- Secure mode: works for a person with proper permissions

chrisknoll:
Sure, and it would be good to put in writing what a semantic version change represents in this context:

The current design is (and I can state this, since I wrote it): the cache endpoint is not secured.  The current bug is that it is necessary (something that I was not aware of) to 'whitelist' endpoints for non-restricted access.  That's the missing piece.

You want to make it from non-secured to secured.  This is a basic functionality change.  The change is going from 'you don't need to worry about security roles and permissions' to 'you now need to worry about security roles and permissions'.  You can make a claim that you can make it appear transparant to the user, but the fact is _you are adding new permissions to WebAPI as a result of a *hotfix* semantic version update_.

If you don't like that just anyone can invoke an endpoint without restriction, you can change that behavior, but not as part of a patch...that's a functionality change that should go into 2.8.  If a function would enable 'bad actors' to do something in 2.7.1 due to a design limitation (the decision to allow anyone to do it) that capability should continue to exist in 2.7.2 etc until you are the opportunity to change the behavior in a minor release update (2.8).  There could be 'good actors' that depend on that behavior, and there should be some control about what sort of behavior we will allow to change as a result of a patch-version semantic version increase.

All this being said, I'm in agreement that the group would like to have this endpoint secured, and there needs to be more discussion (or you unilaterally decide) what the interface to the function is, what the default permissions on the function is, etc.

Just do that in 2.8.

pavgra:
> You want to make it from non-secured to secured. This is a basic functionality change

Have you seen the example in my previous comment?

It _doesn't work_ with security enabled. I will make it working. It _doesn't require any auth headers_ in non-secure mode. That won't change. What's wrong? Seems like we talked that through during 1-hour call and continue to spend time on the topic.

chrisknoll:
This pr fixes the issue.  If that's all you need, then just take this PR.

I know we spent an hour talking on the call, but the topic of semantic versioning came up, we spent some time talking about what a 2.7.4 would mean compared to 2.8.  I think everyone understood that 2.7.X branches are for patches, and are not part of any roadmap. 

But after all that, you say 'just to confirm, i'm going add new permission types in 2.7.4....it's like we didn't even have the conversation at all.

There's something that you're being very stubborn on:  the /cache endpoint was never intended to require authorization to use (again, _I wrote it_, so I am telling you how it should work, not the other way around) yet you are insisting that it should have been done differently.  Well, we can't go back in time, so we have to live with the _semantics_ of the endpoint ie: it is unsecured. 

Can we agree that, since I wrote it, that when I tell you 'it wasn't meant to be secured, the bug is that the rest endpoint was not marked as 'any access', and therefore this PR fixes this issue', that you can make your _semantic_ changes to the endpoint in 2.8?  You keep wanting to make this a 'secured vs. not secured' argument, but that was an argument that should have been made for the 2.7.0 release, not for some _patch_ of 2.7.4.

pbr6cornell:
Team:

I thought we have reached agreement on the ATLAS WG call to the following two items, both of which @pavgra took the action to lead:

1) Make a change in WebAPI for v2.7.4 that makes the cache/clear security enabled.  This involves adding permissions to this endpoint.   

2) Make a change to ATLAS for v2.8, which will likely involve adding a 'Clear server cache' button to the Configuration page, and make that button role-based security enabled (with default that only 'Admin' role can see button, but organizations can change).

I respect that @chrisknoll is trying to make sure we are clear in how we address bugfixes for minor releases and new feature enhancements for major releases, and I appreciate that all of us are trying to mature our development practices to get aligned on a common approach.  In this case, it feels like we are still repaying our technical debt from past efforts, and so I think we should implement the hotfix in 2.7.4, as agreed on the call, to support the portion of the community that is already deploying ATLAS/WebAPI with security enabled now and to close an open endpoint that could pose a security risk, and then we fully bake the new feature functionality of an Admin console button in 2.8 to encourage a broader portion of the community to implement security once we work it through it paces and address the outstanding 2.8 open issues.

PR #1314, opened 2019-08-27
upgrade zip4j dependency

YaroslavTir:
dependency itself was upgraded in the common library, here I just fixed code.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileContentSensitiveInfoServiceImpl.java
+++ b/src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileContentSensitiveInfoServiceImpl.java
@@ -1,7 +1,7 @@
 package org.ohdsi.webapi.executionengine.service;
 
 import com.odysseusinc.arachne.execution_engine_common.util.CommonFileUtils;
-import net.lingala.zip4j.core.ZipFile;
+import net.lingala.zip4j.ZipFile;
 import net.lingala.zip4j.exception.ZipException;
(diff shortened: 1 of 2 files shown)


JamesSWiggins:
Hi, this commit fixes this issue:  https://github.com/OHDSI/WebAPI/issues/1315.  Can it be merged into the releases tagged v2.7.2 and/or v2.7.3?  They aren't compiling right now.

pavgra:
@JamesSWiggins , that would introduce new dependency into the released version which is not the best way to address the problem. The situation here is that we used SNAPSHOT version of arachne-commons dependency in the 2.7.3 and the snapshot version has been received breaking changes. So, what we are going to do is to mark the arachne-commons as of 2.7.3 release date state with a specific tag and replace the SNAPSHOT dependency with the specific tag in 2.7.3. Will be done the upcoming week.

rohitkumar5058:
Hey , I just have this dependency in my pom:

`<dependency>
            <groupId>net.lingala.zip4j</groupId>
            <artifactId>zip4j</artifactId>
            <version>2.11.2</version>
   </dependency>`

 I get the following error: 
 symbol:   class Zip4jConstants
[ERROR]   location: package net.lingala.zip4j.util


then I was adding following additional dependency along with the above dependency and it worked:
`<dependency>
            <groupId>net.lingala.zip4j</groupId>
            <artifactId>zip4j</artifactId>
            <version>1.3.2</version>
            <scope>compile</scope>
        </dependency>`

But why we need to add this, they are same dependency, and the 1.3.2 version has vulnerabilities too, which gives Blackduck issue to my project. issue still exists, Can you please comment and help me on this?

Issue #1088, opened 2019-05-14
Re-use generated cohort across analyses


pavgra:
As we implemented self-sufficient analyses which can be executed on their own w/o non-transparent pre-requirements (e.g. a need to generate cohorts before executing IR), we can focus on performance and introduce repo of shared analyses results which could be utilized as cache and let us running e.g. cohort generation only when there are no results for a given cohort design (hash) yet.

chrisknoll:
I'm glad we are coming around to addressing a performance deficit that was introduced with these changes (eg: cohorts are always re-generated when executing IR, a behavior that did not exist prior).

There is one important detail that hasn't been addressed since we introduced generations based on 'design snapshots': cleanup.  part of this work should implement proper cleanup procedures so that valuable resources can be recovered after a design is no longer needed/referenced/wanted.  When proposing your enhancements, be sure to address resource cleanup in your proposal.

pavgra:
Requirements:
- Results for the re-usage must be stored in a database in relational, normalized format (e.g. **no** JSON / JSONB or other NoSQL / JCR data structures) since cohort records are used by different types of databases and environments (R / Java / SQL)
- The mechanism must ensure consistency at a database level due to usage by different environments. I.e. R script should not be able to remove half of the cohort records while Atlas would still think that results are cached and could be re-used
- The mechanism should be generalized enough to support extension for different types of results, i.e. not only cohort records
- The mechanism should support caching by design to allow the following workflows:
  - There are two different cohort definitions (two different records with different IDs) in the system which define the same design. If the first definition has already been generated and its results were persisted, there is no need to run generation from scratch for the second definition
  - Some analysis utilizes an older version of cohort, cohort which has been generated but later was changed - analysis should be able to use cache
- The mechanism should provide the ability to clean-up outdated results:
  - Manually erase results that have been cached before a certain date
  - Manually erase results that have been cached for an entity which was updated more than N times since the caching occurred
  - Scheduled erasure of cached results
- The caching should be unified for different entities: e.g. now we have isolated generation in IR, Pathways, CC, and non-isolated generation in Cohorts

anthonysena:
> The mechanism must ensure consistency at a database level 

I'd also add here that this mechanism should globally prevent a cohort generation when it is being used during an analysis (i.e. in IR, Pathways, CC). Presumably a user should not alter the contents of the target table that holds the `subject_id`s otherwise it could corrupt the analysis. I think that is what is implied here but would like it explicitly stated.

[...]

PR #1318, opened 2019-09-01
Re-use generated cohort across analyses

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/1088

depends on https://github.com/OHDSI/circe-be/pull/93, https://github.com/OHDSI/ArachneCommons/pull/202


src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
```diff
@@ -0,0 +1,117 @@
...
+
+        this.generationCacheService = generationCacheService;
+    }
+
+    public CacheResult computeIfAbsent(CohortDefinition cohortDefinition, Source source, CohortGenerationRequestBuilder requestBuilder, BiConsumer<Integer, String[]> sqlExecutor) {
+
+        CacheableGenerationType type = CacheableGenerationType.COHORT;
+        String designHash = generationCacheService.getDesignHash(type, cohortDefinition.getDetails().getExpression());
+
+        synchronized (monitors.computeIfAbsent(new CacheableResource(type, designHash, source.getSourceId()), cr -> new Object())) {
```
YaroslavTir: I am just curious is it possible to replace this `synchronized` just with 
```
monitors.compute(new Cacheable(), value -> {
   synchronized code;
   return new Object(); 
}
```

YaroslavTir: and also do run webApi in high availability mode? (2 and more nodes?) if so then this synchronization would be spread through all nodes.

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
```diff
@@ -72,30 +81,33 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
...
-                                        cleanupManager.cleanupTempTables();
-                                    }
+                                    CohortGenerationRequestBuilder generationRequestBuilder = new CohortGenerationRequestBuilder(
+                                        sessionId,
+                                        targetSchema,
+                                        COHORT_GENERATIONS_TABLE,
+                                        GENERATION_ID,
+                                        false
+                                    );
+                                    GenerationCacheHelper.CacheResult res = generationCacheHelper.computeIfAbsent(cd, source, generationRequestBuilder, (resId, sqls) -> {
```
YaroslavTir: the function name is not clear, Absence of what? we pass 3 arguments, but actually check the absence of the cache

src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+        Source source = sourceRepository.findBySourceId(sourceId);
+        GenerationCache generationCache = generationCacheRepository.findByTypeAndAndDesignHashAndSource(type, designHash, source, EntityGraphUtils.fromAttributePaths("source"));
+        GenerationCacheProvider provider = getProvider(type);
+        if (generationCache != null) {
+            String checksum = provider.getResultsChecksum(generationCache.getSource(), generationCache.getResultIdentifier());
+            if (Objects.equals(generationCache.getResultChecksum(), checksum)) {
+                return generationCache;
+            } else {
+                removeCache(generationCache.getType(), generationCache.getSource(), generationCache.getResultIdentifier());
```
YaroslavTir: I would not expect that get method removes something or validate something. I would recommend rename method like`getCacheOrEraseInvalid`

src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+        log.info(CACHE_CREATED, type, designHash);
+
+        return generationCache;
+    }
+
+    @Override
+    public void removeCache(CacheableGenerationType type, Source source, Integer resultIdentifier) {
+
+        // Cleanup cached results
```
YaroslavTir: I don't like such comments in the code.  If it is really necessary you can extract methods with the name `cleanupCacheResults` and `cleanupCacheRecords`.

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
```diff
@@ -0,0 +1,381 @@
...
+        isSqlExecuted.set(false);
+
+        res = generationCacheHelper.computeIfAbsent(
+                cohortDefinition,
+                source,
+                cohortGenerationRequestBuilder,
+                (resId, sqls) -> isSqlExecuted.set(true)
+        );
+
+        Assert.isTrue(!isSqlExecuted.get(), "Cohort results are retrieved from cache");
```
YaroslavTir: :)  `isFalse(isSqlExecuted, ... ) `

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
```diff
@@ -0,0 +1,381 @@
...
+        try {
+            System.setProperty("datasource.url", getDataSource().getConnection().getMetaData().getURL());
+            System.setProperty("flyway.datasource.url", System.getProperty("datasource.url"));
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
+        }
+    }
+
+    @Before
+    public void beforeTest() throws SQLException {
```
YaroslavTir: really minor thing, but the standard name for `@Before` method is `setUp`

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
```diff
@@ -0,0 +1,381 @@
...
+        File eunomiaZip = File.createTempFile("eunomia", "", tempDir.toFile());
+
+        try (InputStream is = GenerationCacheTest.class.getResourceAsStream(EUNOMIA_CSV_ZIP)) {
+            Files.copy(is, eunomiaZip.toPath(), REPLACE_EXISTING);
+        }
+
+        ZipFile zipFile = new ZipFile(eunomiaZip);
+        zipFile.extractAll(tempDir.toAbsolutePath().toString());
+
+        for (final File file : tempDir.toFile().listFiles()) {
```
YaroslavTir: `tempDir.toFile().stream()`

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationUtils.java
```diff
@@ -0,0 +1,71 @@
...
+
+      String[] names = new String[]{"iteration", "ruleName", "ruleDescription"};
+      List<InclusionRule> inclusionRules = request.getExpression().inclusionRules;
+      for (int i = 0; i < inclusionRules.size(); i++) {
+        InclusionRule r = inclusionRules.get(i);
+        String[] values = new String[]{((Integer) i).toString(), r.name, MoreObjects.firstNonNull(r.description, "")};
+
+        String inclusionRuleSql = SqlRender.renderSql(insertSql, names, values);
+        sqlBuilder.append(inclusionRuleSql).append("\n");
+      }
```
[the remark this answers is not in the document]
pavgra: Not really: where would you get the `i` from? (which is used as the value for `iteration` variable)

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
```diff
@@ -27,35 +33,67 @@
...
-        jobParams.getOrDefault(SESSION_ID, SessionUtils.sessionId()).toString(),
-        jobParams.get(TARGET_DATABASE_SCHEMA).toString(),
-        jobParams.get(TARGET_TABLE).toString(),
-        Boolean.valueOf(jobParams.get(GENERATE_STATS).toString())
+
+    Integer cohortDefinitionId = Integer.valueOf(jobParams.get(COHORT_DEFINITION_ID).toString());
+    Integer sourceId = Integer.parseInt(jobParams.get(SOURCE_ID).toString());
+    String targetSchema = jobParams.get(TARGET_DATABASE_SCHEMA).toString();
+    String sessionId = jobParams.getOrDefault(SESSION_ID, SessionUtils.sessionId()).toString();
+    Boolean generateStats = Boolean.valueOf(jobParams.get(GENERATE_STATS).toString());
```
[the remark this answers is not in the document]
pavgra: Then you would have a single record in the `batch_job_execution_params` table instead of multiple per-parameter records which 1. violates the 1st Normal Form + 2. limits you in joining the param records (which is currently done for PLE/PLP/Pathways, e.g. see `db/migration/oracle/V2.7.0.20190225165203__plp_gen_view.sql`)

src/main/java/org/ohdsi/webapi/generationcache/CohortGenerationCacheProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public String getDesignHash(String design) {
+
+        CohortDefinitionDetails cohortDetails = new CohortDefinitionDetails();
+        cohortDetails.setExpression(design);
+        return cohortDetails.calculateHashCode().toString();
+    }
+
+    // NOTE: should not be used directly! Only via GenerationCacheService
```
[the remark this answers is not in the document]
pavgra: I totally agree that a contract is better than any comment, but here it is a bit tricky - we expect that CacheProvider implements `nextResultIdentifier` method, it is required to pass `checkCachingWithPrefilledResults` test. But the method should not be called directly, otherwise `checkCachingWithEmptyResultSet` test will fail - we need to get MAX of IDs from results schema and generation caches list.

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
pavgra: I do not see such method:
https://docs.spring.io/autorepo/docs/spring/4.3.2.RELEASE/javadoc-api/org/springframework/util/Assert.html

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
pavgra: image

src/main/java/org/ohdsi/webapi/generationcache/CohortGenerationCacheProvider.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public String getResultsChecksum(Source source, Integer resultIdentifier) {
+
+        long startTime = System.currentTimeMillis();
+        PreparedStatementRenderer psr = new PreparedStatementRenderer(
+                source,
+                COHORT_CHECKSUM_SQL_PATH,
+                "@" + RESULTS_DATABASE_SCHEMA,
```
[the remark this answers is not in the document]
pavgra: Why better? Honestly, I do not think that it is either better readable or shorter

src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheServiceImpl.java
pavgra: Fixed

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
YaroslavTir: `Arrays.stream(tempDir.toFile().listFiles())`

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationUtils.java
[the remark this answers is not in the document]
YaroslavTir: true, you cannot do it directly using java 8, but guava provides a nice utils method for this
```
            Streams.mapWithIndex(inclusionRules.stream(), ImmutablePair::of)
                    .map(pair -> {
                                String index = pair.getRight().toString();
                                InclusionRule rule = pair.getLeft();
                                return new String[]{index, rule.name, MoreObjects.firstNonNull(rule.description, "")};
                            }
                    )
                    .map(values -> SqlRender.renderSql(insertSql, names, values))
                    .forEach(inclusionRuleSql -> sqlBuilder.append(inclusionRuleSql).append("\n"));
```
anyway it is up to you, I don't mind to have here foreach.

src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
pavgra: From the `ConcurrentHashMap.java` -> `compute`:
```
* Attempts to compute a mapping for the specified key and its
* current mapped value (or {@code null} if there is no current
* mapping).
```

So, looks that it will be called just once which is not a suitable behavior

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortGenerationUtils.java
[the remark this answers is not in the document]
pavgra: Longer. And discussible whether is easier to read

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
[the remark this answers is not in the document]
pavgra: Doesn't seem that we have any 2 tasklets with the same set of params so that the extraction code could be reused. And if it cannot be re-used, I'm very unsure if the rewrite is reasonable: currently, there is just 5 lines which are really straightforward. With your approach - much more code is requried and not the fact that the new code is better readable

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
pavgra: Longer. And discussible whether is easier to read

pom.xml
```diff
@@ -629,7 +632,7 @@
     <dependency>
       <groupId>org.ohdsi</groupId>
       <artifactId>circe</artifactId>
-      <version>1.8.1</version>
+      <version>${circe.version}</version>
```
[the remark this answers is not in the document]
pavgra: It's more a matter of convention which allows us to override the version via configuration / component utilizing the WebAPI

src/main/java/org/ohdsi/webapi/DataAccessConfig.java
```diff
@@ -115,7 +115,7 @@ public EntityManagerFactory entityManagerFactory() {
         factory.setJpaVendorAdapter(vendorAdapter);
         factory.setJpaProperties(getJPAProperties());
         factory.setPackagesToScan("org.ohdsi.webapi");
-        factory.setDataSource(primaryDataSource());
+        factory.setDataSource(dataSource);
```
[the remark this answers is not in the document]
pavgra: This is to utilize DI and to be able to override just a Primary Data Source bean so that the overriden bean is automatically utilized by the entityManagerFactory w/o need to override the entityManagerFactory as well (because currently, we reference the concrete implementation - the `primaryDataSource` method and overriden data source bean won't be taken into account)

pom.xml
```diff
@@ -768,6 +771,12 @@
       <version>2.8.7</version>
       <type>jar</type>
     </dependency>
+    <dependency>
+      <groupId>com.opentable.components</groupId>
```
anthonysena: Very cool trying to use OpenTable Embedded PostgreSQL Component: https://github.com/opentable/otj-pg-embedded

src/main/java/org/ohdsi/webapi/FlywayConfig.java
```diff
@@ -17,6 +18,7 @@
  *
  */
 @Configuration
+@ConditionalOnProperty(prefix = "flyway", name = "enabled", matchIfMissing = true)
```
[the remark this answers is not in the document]
pavgra: Yep, this was added when I tried to utilize H2 + Hibernate's `generate-ddl` feature to be able to disable Flyway. It is not actually used in the most recent version of tests but I thought that leaving such capability (to disable flyway) wouldn't harm

src/main/java/org/ohdsi/webapi/JobConfig.java
```diff
@@ -42,6 +43,7 @@
  */
 @Configuration
 @EnableBatchProcessing
+@DependsOn({"batchDatabaseInitializer"})
```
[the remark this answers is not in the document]
pavgra: This might be required if you decide to go for Hibernate's `generate-ddl` instead of Flyway scripts. The same story as in the previous comment

src/main/java/org/ohdsi/webapi/generationcache/CleanupScheduler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.commons.lang.time.DateUtils;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+import java.util.Date;
+import java.util.List;
+
+@Component
+public class CleanupScheduler {
```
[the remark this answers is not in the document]
pavgra: Spring invokes methods marked with `@Scheduled` annotation automatically

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
```diff
@@ -76,7 +74,7 @@ public void afterJob(JobExecution je) {
     Integer defId = Integer.valueOf(jobParams.getString("cohort_definition_id"));
     Integer sourceId = Integer.valueOf(jobParams.getString("source_id"));
     Source source = sourceService.findBySourceId(sourceId);
-    String cohortTable = jobParams.getString(TARGET_DATABASE_SCHEMA) + "." + jobParams.getString(TARGET_TABLE);
+    String cohortTable = jobParams.getString(TARGET_DATABASE_SCHEMA) + ".cohort";
```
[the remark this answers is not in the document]
pavgra: Actually, in case of the GenerationJobExecutionListener, the variable was always equal to `DEFAULT_COHORT_TABLE` (which is `cohort`) but to discover it you had to go through almost ten function calls. So, attempted to simplify it

src/main/java/org/ohdsi/webapi/FlywayConfig.java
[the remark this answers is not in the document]
chrisknoll: Having redundant/unused code in the code base could be harmful (indicating to other developers that there is 2 modes to alter the DB schema, when there should only be one approach (single source of truth, and all that)).  If it's not used, it should be removed.  @pavgra , you've made such statements in numerous areas of the code (both from a code duplication perspective and a single source of truth). Let's be consistent with these principles.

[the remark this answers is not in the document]
chrisknoll: If I understand this, it's not implementing a hibernate ddl generation, but rather allowing the choice to disable flyway, so this isn't a matter of introducing a new mechanism to instantiate the webapi schema, but rather letting the users control it.

While not causing direct harm, I think it should be understood that there's a defined mechanism in place to manage this function.  If it's not understood, you can easily expect that a person who didn't know what that setting did and left it disabled, they'd be confused why the database tables aren't being created for them, and they thought they were being 'safer' buy disabling things by default and enabling as needed.  

In addition, our hibernate anntoations may not even be complete so that the DDL from hibernate won't produce a complete schema. (ie: are we sure that every column annotation has the correct specification of what flavor of varchar() should be applied?  Is the same choice of varchar appropriate for all platforms (text vs. varchar(max) for example).

So, while it may not cause direct harm, it opens up a 'point of failure' and disables a critical application function (ensuring that the schema is up to date properly).

src/main/resources/resources/generationcache/cohort/resultsChecksum.sql
```diff
@@ -74,6 +74,6 @@ WITH partials AS (
   WHERE generation_id = @generation_id
 )
 SELECT AVG(
-  CAST(CONVERT(VARBINARY, CONCAT('0x', HASHBYTES(line)), 1) AS BIGINT)
+  CAST(CONVERT(VARBINARY, CONCAT('0x', HASHBYTES('MD5',line)), 1) AS BIGINT)
```
anthonysena: @pavgra - I'm still facing a problem after these changes with the code running on PDW. In my instance, this query is resulting in the following error:

````
110802;An internal DMS error occurred that caused this operation to fail. Details: Error converting data type varchar to varbinary.
````

Playing with this query a bit, I revised the last portion to:

````
SELECT AVG(
  CAST(CONVERT(VARBINARY, HASHBYTES('MD5',line), 1) AS BIGINT)
) as checksum
FROM partials
````

Removing the `CONCAT()` helped since the `HASHBYTES` function will return a value of the form `0x7412...`. However, I then faced a different issue:

````
Msg 110802, Level 16, State 1, Line 1
110802;An internal DMS error occurred that caused this operation to fail. Details: Arithmetic overflow error converting expression to data type bigint.
````

Continuing to debug, I have found this is due to the use of the AVG() function. It appears that even though the inner values are CAST to a BIGINT, PDW is still having trouble with this operation. Perhaps this should utilize a temp table approach to avoid the overflow? Meaning insert into temp and then perform the AVG?

Comment (pavgra) [reply src/main/resources/resources/generationcache/cohort/resultsChecksum.sql]: Looking at the word `overflow` I think that the `hashbytes` returns such HEX which represents integer bigger than BIGINT could store. To test the hypothesis, could you wrap the `hashbytes` into `SUBSTRING(HASHBYTES('MD5',line), 1, 8)`?

File: src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java [hunk 20]
```diff
@@ -72,30 +81,33 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
...
+                                                    transactionTemplate,
+                                                    source.getSourceDialect(),
+                                                    sessionId,
+                                                    SourceUtils.getTempQualifier(source)
+                                            );
+                                            cleanupManager.cleanupTempTables();
+                                        }
+                                    });
+                                    String sql = String.format(COPY_CACHED_RESULTS, SourceUtils.getTempQualifier(source), targetTable, cd.getId(), res.getSql());
+                                    cancelableJdbcTemplate.batchUpdate(stmtCancel, sql);
```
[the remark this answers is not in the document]
pavgra: Although, e.g. Postgres throws an exception when `batchUpdate` is canceled and the exception stops caching procedure, I've added code which explicitly checks if generation has been canceled and throws an exception which crashes the caching method (see commit 2716a77 below)

pavgra:
@wivern ,

> Probably we should immediately remove all caches related when source has been deleted.

Done in 26ce49d

anthonysena:
@pavgra testing is going well with this from my side. A question: when I am generating a CC (for example) and one of the cohorts is not yet generated, I noticed that the results are not written to the result schema's `cohort` table. Instead it appears that it relies on the `cohort_generations` table to get the persons in the cohort. Do I have this correct? I have no objections to this approach but wanted to confirm the behavior was expected.

anthonysena:
I was able to run through a number of tests for generating cohorts, CC and pathways without issue. However, IR gave me an issue:

````
2019-09-19 21:49:17.286 INFO taskExecutor-3 org.springframework.batch.core.launch.support.SimpleJobLauncher -  - Job: [SimpleJob: [name=irAnalysis]] launched with the following parameters: [{jobName=IR Analysis: 2: Optum (PDW - DEV) (OPTUM-PDW), analysis_id=2, source_id=1, sessionId=hj9tmhb5, target_table=temp_cohort_hj9tmhb5, time=1568944157063, jobAuthor=anonymous}]
2019-09-19 21:49:17.714 INFO taskExecutor-3 org.springframework.batch.core.job.SimpleStepHandler -  - Executing step: [irAnalysis.createCohortTable]
2019-09-19 21:49:18.796 INFO taskExecutor-3 org.springframework.batch.core.job.SimpleStepHandler -  - Executing step: [irAnalysis.generateCohort]
2019-09-19 21:49:19.042 ERROR taskExecutor-3 org.springframework.batch.core.step.AbstractStep -  - Encountered an error executing step irAnalysis.generateCohort in job irAnalysis
java.util.concurrent.CompletionException: java.lang.NullPointerException
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
	at org.ohdsi.webapi.generationcache.GenerationCacheHelper.computeCacheIfAbsent(GenerationCacheHelper.java:39)
	at org.ohdsi.webapi.cohortcharacterization.GenerateLocalCohortTasklet.lambda$null$1(GenerateLocalCohortTasklet.java:91)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1590)
	... 3 more
2019-09-19 21:49:19.718 INFO taskExecutor-3 org.springframework.batch.core.launch.support.SimpleJobLauncher -  - Job: [SimpleJob: [name=irAnalysis]] completed with the following parameters: [{jobName=IR Analysis: 2: Optum (PDW - DEV) (OPTUM-PDW), analysis_id=2, source_id=1, sessionId=hj9tmhb5, target_table=temp_cohort_hj9tmhb5, time=1568944157063, jobAuthor=anonymous}] and the following status: [FAILED]
````

Here is the IR analysis I was using for testing: [ir-issue-1088.txt](https://github.com/OHDSI/WebAPI/files/3633843/ir-issue-1088.txt).

**Comment by pavgra** [comment]
> confirm the behavior was expected

Yep, that's intended behavior. Once you click Generate cohort after that it will pull the cached results from the `cohort_generations`

**Comment by anthonysena** [comment]
@pavgra thanks for the fixes. From my testing, the caching is working well. As discussed offline, we'll need to create a migration script for the 2.8 release as part of incorporating this work into master. To that end, I started to draft up a migration script:

````
-- Historical generations table
IF OBJECT_ID('@results_schema.cohort_generations', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations
(
	generation_id int NOT NULL,
	SUBJECT_ID bigint NOT NULL,
	cohort_start_date date NOT NULL,
	cohort_end_date date NOT NULL
);


-- Link from all generations to the most recent generation for a cohort_definition_id
IF OBJECT_ID('@results_schema.cohort_generations_ref', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations_ref
(
	generation_id int NOT NULL,
	cohort_definition_id int NOT NULL
);

ALTER TABLE @results_schema.cohort_inclusion ADD COLUMN generation_id int NOT NULL;
ALTER TABLE @results_schema.cohort_inclusion_result ADD COLUMN generation_id int NOT NULL;
ALTER TABLE @results_schema.cohort_inclusion_stats ADD COLUMN generation_id int NOT NULL;

-- Migrate the data so we can drop the cohort_definition_id from the tables above

-- Generate the generation_id sequence based on the cohort definitions in the cohort table
INSERT @results_schema.cohort_generations_ref (generation_id, cohort_definition_id)
SELECT DISTINCT ROW_NUMBER() OVER (ORDER BY cohort_definition_id)
FROM @results_schema.cohort
;

-- TODO: Update the downstream tables to use that new generation_id in place of the cohort_definition_id 
-- Might this be easier to by creating a new table, dropping the old one and then renaming (or re-creating)
-- the target table

-- This part below may not be necessary based on how we approach the step above but leaving here so we ensure that the cohort_definition_id is dropped in these tables.
ALTER TABLE @results_schema.cohort_inclusion DROP COLUMN cohort_definition_id int NOT NULL;
ALTER TABLE @results_schema.cohort_inclusion_result DROP COLUMN cohort_definition_id int NOT NULL;
ALTER TABLE @results_schema.cohort_inclusion_stats DROP COLUMN cohort_definition_id int NOT NULL;
````

Just curious for your thoughts on mapping over the `generation_id`'s to the `cohort_definition_ids` to the `cohort_inclusion*` tables? I think that inserting the data into a new, temp table, dropping the old table and then re-creating the target table might be easiest but may take a while depending on the data size.

Also we should note how these new properties are used:

`
cache.generation.invalidAfterDays=${cache.generation.invalidAfterDays}
cache.generation.cleanupInterval=${cache.generation.cleanupInterval} 
`

**Comment by pavgra** [comment]
Hmm, @anthonysena , wouldn't something as simple as this work?

```
-- Historical generations table
IF OBJECT_ID('@results_schema.cohort_generations', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations
(
	generation_id int NOT NULL,
	SUBJECT_ID bigint NOT NULL,
	cohort_start_date date NOT NULL,
	cohort_end_date date NOT NULL
);

-- Link from all generations to the most recent generation for a cohort_definition_id
IF OBJECT_ID('@results_schema.cohort_generations_ref', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations_ref
(
	generation_id int NOT NULL,
	cohort_definition_id int NOT NULL
); 

INSERT INTO @results_schema.cohort_generations
SELECT cohort_definition_id as generation_id, subject_id, cohort_start_date, cohort_end_date
FROM @results_schema.cohort;

INSERT INTO @results_schema.cohort_generations_ref
SELECT cohort_definition_id as generation_id, cohort_definition_id as cohort_definition_id
FROM @results_schema.cohort;

ALTER TABLE @results_schema.cohort_censor_stats RENAME cohort_definition_id TO generation_id;
ALTER TABLE @results_schema.cohort_inclusion RENAME cohort_definition_id TO generation_id;
ALTER TABLE @results_schema.cohort_inclusion_result RENAME cohort_definition_id TO generation_id;
ALTER TABLE @results_schema.cohort_inclusion_stats RENAME cohort_definition_id TO generation_id;
ALTER TABLE @results_schema.cohort_summary_stats RENAME cohort_definition_id TO generation_id;
```

**Comment by anthonysena** [comment]
Thanks @pavgra - that does simplify the generation_ref side for sure. My only concern is the column rename operation as this is not as straight forward for other dialects. Or at least, it is not very simple for SQL Server which requires the use of a system stored proc `sp_rename`: https://docs.microsoft.com/en-us/sql/relational-databases/tables/rename-columns-database-engine?view=sql-server-2017. Not sure of the other DB platforms but they may also face some challenges.

**Comment by pavgra** [comment]
@anthonysena , then we can go with a table replacement approach (table renaming is [supported in PDW](https://docs.microsoft.com/en-us/sql/t-sql/statements/rename-transact-sql?view=aps-pdw-2016-au7)):
```
ALTER TABLE @results_schema.cohort_censor_stats RENAME TO old_cohort_censor_stats;

CREATE TABLE @results_schema.cohort_censor_stats (
  generation_id int NOT NULL,
  lost_count BIGINT NOT NULL
);

INSERT INTO @results_schema.cohort_censor_stats SELECT * FROM @results_schema.old_cohort_censor_stats;

DROP TABLE @results_schema.old_cohort_censor_stats;
```

anthonysena:
Looks like a good approach - here is the revised script:

````
-- Historical generations table
IF OBJECT_ID('@results_schema.cohort_generations', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations
(
	generation_id int NOT NULL,
	SUBJECT_ID bigint NOT NULL,
	cohort_start_date date NOT NULL,
	cohort_end_date date NOT NULL
);

-- Link from all generations to the most recent generation for a cohort_definition_id
IF OBJECT_ID('@results_schema.cohort_generations_ref', 'U') IS NULL
CREATE TABLE @results_schema.cohort_generations_ref
(
	generation_id int NOT NULL,
	cohort_definition_id int NOT NULL
); 

-- Establish the cohort generation references based on the cohort_definition_id's in the system
INSERT INTO @results_schema.cohort_generations
SELECT cohort_definition_id as generation_id, subject_id, cohort_start_date, cohort_end_date
FROM @results_schema.cohort;

INSERT INTO @results_schema.cohort_generations_ref
SELECT cohort_definition_id as generation_id, cohort_definition_id as cohort_definition_id
FROM @results_schema.cohort;

-- Migrate cohort_inclusion
ALTER TABLE @results_schema.cohort_inclusion RENAME TO @results_schema.old_cohort_inclusion;

CREATE TABLE @results_schema.cohort_inclusion (
  generation_id int NOT NULL,
  rule_sequence int NOT NULL,
  name varchar(255) NULL,
  description varchar(1000) NULL
);

INSERT INTO @results_schema.cohort_inclusion (generation_id, rule_sequence, name, description)
SELECT cohort_definition_id, rule_sequence, name, description 
FROM @results_schema.old_cohort_inclusion;

TRUNCATE TABLE @results_schema.old_cohort_inclusion;
DROP TABLE @results_schema.old_cohort_inclusion;

-- Migrate cohort_inclusion_result
ALTER TABLE @results_schema.cohort_inclusion_result RENAME TO @results_schema.old_cohort_inclusion_result;

CREATE TABLE @results_schema.cohort_inclusion_result (
  generation_id int NOT NULL,
  mode_id int NOT NULL,
  inclusion_rule_mask bigint NOT NULL,
  person_count bigint NOT NULL
);

INSERT INTO @results_schema.cohort_inclusion_result (generation_id, mode_id, inclusion_rule_mask, person_count)
SELECT generation_id, mode_id, inclusion_rule_mask, person_count
FROM @results_schema.old_cohort_inclusion_result
;

TRUNCATE TABLE @results_schema.old_cohort_inclusion_result;
DROP TABLE @results_schema.old_cohort_inclusion_result;

-- Migrate cohort_inclusion_stats
ALTER TABLE @results_schema.cohort_inclusion_stats RENAME TO @results_schema.old_cohort_inclusion_stats;

CREATE TABLE @results_schema.cohort_inclusion_stats (
  generation_id int NOT NULL,
  rule_sequence int NOT NULL,
  mode_id int NOT NULL,
  person_count bigint NOT NULL,
  gain_count bigint NOT NULL,
  person_total bigint NOT NULL
);

INSERT INTO @results_schema.cohort_inclusion_stats (generation_id, rule_sequence, mode_id, person_count, gain_count, person_total)
SELECT cohort_definition_id, rule_sequence, mode_id, person_count, gain_count, person_total
FROM @results_schema.old_cohort_inclusion_stats
;

TRUNCATE TABLE @results_schema.old_cohort_inclusion_stats;
DROP TABLE @results_schema.old_cohort_inclusion_stats;

-- Migrate cohort_summary_stats
ALTER TABLE @results_schema.cohort_summary_stats RENAME TO @results_schema.old_cohort_summary_stats;

CREATE TABLE @results_schema.cohort_summary_stats(
  generation_id int NOT NULL,
  mode_id int NOT NULL,
  base_count bigint NOT NULL,
  final_count bigint NOT NULL
);

INSERT INTO @results_schema.cohort_summary_stats (generation_id, mode_id, base_count, final_count)
SELECT cohort_definition_id, mode_id, base_count, final_count
FROM @results_schema.old_cohort_summary_stats
;

TRUNCATE TABLE @results_schema.old_cohort_inclusion_stats;
DROP TABLE @results_schema.old_cohort_inclusion_stats;

-- Migrate cohort_censor_stats
ALTER TABLE @results_schema.cohort_censor_stats RENAME TO @results_schema.old_cohort_censor_stats;

CREATE TABLE @results_schema.cohort_censor_stats (
  generation_id int NOT NULL,
  lost_count BIGINT NOT NULL
);

INSERT INTO @results_schema.cohort_censor_stats (generation_id, lost_count)
SELECT cohort_definition_id, lost_count 
FROM @results_schema.old_cohort_censor_stats;

TRUNCATE TABLE @results_schema.old_cohort_censor_stats;
DROP TABLE @results_schema.old_cohort_censor_stats;

````

PR #1340, opened 2019-10-02
Upgrade Impala driver to 2.6 (using JDBC 4.2)

pavgra:
fixes https://github.com/OHDSI/WebAPI/pull/1339

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -903,50 +903,15 @@
       <id>webapi-impala</id>
       <properties>
         <impala.enabled>true</impala.enabled>
-				<!-- Impala JDBC driver path -->
-				<impala.classpath>...path/to/impala/jdbc/drivers...</impala.classpath>
+        <impala.driver.version>2.6.15</impala.driver.version>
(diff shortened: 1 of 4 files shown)


wivern:
An issue with unsupported batch updates has gone.
Unfortunately there is a new issue now related with casts:
```
Error Code: 0, SQL state: TStatus(statusCode:ERROR_STATUS, sqlState:HY000, errorMessage:AnalysisException: Possible loss of precision for target table 'synpuf_53_temp.s9qhd84zcov_ref'.
Expression 'CAST(concat('observation during day -30 through 0 days relative to index: ', (CASE WHEN (concept.concept_name IS NULL) THEN 'Unknown concept' ELSE concept.concept_name END)) AS STRING)' (type: STRING) would need to be cast to VARCHAR(512) for column 'covariate_name'
```
5ed3b327-58e1-4f4c-9757-b99174f2a7bf

pavgra:
Doesn't seem to be reproducible from my side. Could you give it another shot and double-check that all your deps are up-to-date?

PR #1347, opened 2019-10-15
Safe casts in the WHERE expression

### LINKED ISSUE #1366: Heracles analyses fails on BigQuery
Direct CAST of stratum_1 value to int is not safe and may fail when the query planner decides to execute WHERE clause conditions in parallel. It usually happens on BigQuery.

acumarav:
BigQuery tends to execute expressions in WHERE clause with a very high parallelism level which makes behavior dramatically different compared to regular databases. There is a high probability of the bugs in case of CASTS with many 'AND' expressions. It is known and well documented feature of the Bigquery and it offers a bunch of SAFE_* methods to address this feature. As we cannot write code directly for the BigQuery I have to patch where part to sanitize such a values.
Closes #1366


wivern:
@acumarav Could you please check this on other databases rather than BigQuery?

acumarav:
@wivern - the latest changes included into this branch. Please have a look.

PR #1348, opened 2019-10-15
Safe casts in the WHERE expression rc-2.7.5

acumarav:
BigQuery tends to execute expressions in WHERE clause with a very high parallelism level which makes behavior dramatically different compared to regular databases. There is a high probability of the bugs in case of CASTS with many 'AND' expressions. It is known and well documented feature of the Bigquery and it offers a bunch of SAFE_* methods to address this feature. As we cannot write code directly for the BigQuery I have to patch where part to sanitize such a values.


olga-ganina:
Fixes #1366

PR #1354, opened 2019-10-25
Fixes AD groups import fails. User import job params were moved to separate table

### LINKED ISSUE #1155: AD groups import fails

ssuvorov-fls:
fixes #1155


src/main/resources/db/migration/sqlserver/V2.8.0.20190531181956__alter_job-execution-params_string-val.sql
```diff
@@ -0,0 +1,136 @@
...
+
+CREATE SEQUENCE ${ohdsiSchema}.user_import_sequence;
+GO
+
+CREATE TABLE ${ohdsiSchema}.user_import (
+  id INTEGER DEFAULT NEXT VALUE FOR ${ohdsiSchema}.user_import_sequence,
+  provider VARCHAR NOT NULL,
+  preserveRoles BIT NOT NULL ,
+  userRoles VARCHAR,
+  roleGroupMapping VARCHAR,
```
wivern: Same as above

src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java
```diff
@@ -27,7 +27,7 @@
 
     LdapTemplate ldapTemplate = getLdapTemplate();
     AndFilter filter = new AndFilter();
-    filter.and(getCriteria(OBJECTCLASS_ATTR, getGroupClasses()));
+//    filter.and(getCriteria(OBJECTCLASS_ATTR, getGroupClasses()));
```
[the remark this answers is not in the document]
ssuvorov-fls: fixed

src/main/resources/db/migration/sqlserver/V2.8.0.20190531181956__alter_job-execution-params_string-val.sql
```diff
@@ -0,0 +1,136 @@
...
+DROP VIEW ${ohdsiSchema}.user_import_job_history;
+
+CREATE SEQUENCE ${ohdsiSchema}.user_import_sequence;
+GO
+
+CREATE TABLE ${ohdsiSchema}.user_import (
+  id INTEGER DEFAULT NEXT VALUE FOR ${ohdsiSchema}.user_import_sequence,
+  provider VARCHAR NOT NULL,
+  preserveRoles BIT NOT NULL ,
+  userRoles VARCHAR,
```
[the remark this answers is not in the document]
ssuvorov-fls: fixed

src/main/resources/db/migration/sqlserver/V2.8.0.20190531181956__alter_job-execution-params_string-val.sql
ssuvorov-fls: fixed

wivern:
@ssuvorov-fls one thing confuse me a little:
With this, we've got a `UserImport` and `UserImportJob` entities that are completely the same excepting Job has abilities to be scheduled. UserImport could be treated as Job that executes immediately only once.
The next one misuse is both entities shares a job history.
And finally, UserImport is not visible from UI in any way unless when preparing to run a new one. I reckon it's not obvious.

wivern:
One more issue, hoping really final:
New users aren't visible in any groups after import finished.
But page reload makes them visible. I guess there're some UI changes required.

wivern:
Do we need a 'user_import` table? It seems one's not used anymore.

src/main/java/org/ohdsi/webapi/user/importer/UserImportController.java
```diff
@@ -85,17 +112,38 @@ public Response testConnection(@PathParam("type") String type) {
...
+    @Path("user/import")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public UserImportJobDTO importUsers(List<AtlasUserRoles> users,
+                                @QueryParam("provider") String provider,
+                                @DefaultValue("TRUE") @QueryParam("preserve") Boolean preserveRoles) {
+        LdapProviderType providerType = LdapProviderType.fromValue(provider);
+
+        Calendar calendar = GregorianCalendar.getInstance();
+        calendar.add(Calendar.SECOND, 5);
```
pavgra: What does this magic logic do?

ssuvorov-fls: It is timeout before the job will be started. By default current time is used but in this case job wil be treated as closed (see com.odysseusinc.scheduler.service.BaseJobServiceImpl.createJob and com.odysseusinc.scheduler.service.BaseJobServiceImpl.isClosed:163)

pavgra: I believe it would be reasonable to extract the lines into a function with a self-descriptive name and put the comment there. What do you think?

ssuvorov-fls: agree, getting job start date was extracted to separate function

PR #1356, opened 2019-11-07
migration of criteria set value was created (#1206)

### LINKED ISSUE #1206: Unify Feature Analysis types constant
Today we have `StandardFeatureAnalysisType` defining `CRITERIA_SET` and `CAST('CRITERIA' AS VARCHAR(255)) as fa_type` in SQLs

ssuvorov-fls:
fixes #1206

### DIFF
--- a/src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
+++ b/src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
@@ -1,0 +1,63 @@
+package org.ohdsi.webapi.db.migartion;
+
+import com.odysseusinc.arachne.commons.config.flyway.ApplicationContextAwareSpringMigration;
+import org.ohdsi.circe.helper.ResourceHelper;
+import org.ohdsi.sql.SqlRender;
+import org.ohdsi.webapi.service.AbstractDaoService;
+import org.ohdsi.webapi.source.Source;
+import org.ohdsi.webapi.source.SourceDaimon;
+import org.ohdsi.webapi.source.SourceRepository;
+import org.ohdsi.webapi.util.CancelableJdbcTemplate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.stereotype.Service;
+
+@Component
+public class V2_8_0_20191106092815__migrateEventFAType implements ApplicationContextAwareSpringMigration {
+    private final static String SQL_PATH = "/db/migration/java/V2_8_0_20191106092815__migrateEventFAType/updateFaType.sql";
+    private static final Logger log = LoggerFactory.getLogger(V2_8_0_20191106092815__migrateEventFAType.class);
+
+    private final SourceRepository sourceRepository;
+    private final V2_8_0_20191106092815__migrateEventFAType.MigrationDAO migrationDAO;
+
+    @Autowired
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
```diff
@@ -0,0 +1,63 @@
...
+                    return; // no results in this source
+                }
+
+                CancelableJdbcTemplate jdbcTemplate = migrationDAO.getSourceJdbcTemplate(source);
+
+                this.migrationDAO.updateColumnValue(source, jdbcTemplate);
+
+            }
+            catch(Exception e) {
+                log.error(String.format("Failed to update fa type value for source: %s (%s)", source.getSourceName(), source.getSourceKey()));
```
wivern: Why should we ignore an error?

src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
```diff
@@ -0,0 +1,63 @@
...
+        });
+    }
+
+    @Service
+    public static class MigrationDAO extends AbstractDaoService {
+        public void updateColumnValue(Source source, CancelableJdbcTemplate jdbcTemplate) {
+            String resultsSchema = source.getTableQualifier(SourceDaimon.DaimonType.Results);
+            String[] params = new String[]{"results_schema"};
+            String[] values = new String[]{resultsSchema};
+            String translatedSql = SqlRender.renderSql(ResourceHelper.GetResourceAsString(SQL_PATH), params, values);
```
wivern: Why not to read sql only once rather than reload it for each source?

src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
```diff
@@ -0,0 +1,63 @@
...
+        this.migrationDAO = migrationDAO;
+    }
+
+    @Override
+    public void migrate() throws Exception {
+        sourceRepository.findAll().forEach(source -> {
+            try {
+                String resultsSchema = source.getTableQualifierOrNull(SourceDaimon.DaimonType.Results);
+
+                if (resultsSchema == null) {
```
wivern: Why not to filter out sources without result schema?

ssuvorov-fls: CRUD function with filtering will have too long name, because it also involves WHERE-clause

src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
ssuvorov-fls: exception is now thrown outside

src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
ssuvorov-fls: done

src/main/java/org/ohdsi/webapi/db/migartion/V2_8_0_20191106092815__migrateEventFAType.java
ssuvorov-fls: filtering is added

PR #1363, opened 2019-11-19
🛩 Flyway mode_id migration and testing setup

psbrandt:
I don't expect this PR to get merged, but I just wanted to walk through my setup to write a new migration and test it. This is related to [this forum thread](https://forums.ohdsi.org/t/cohort-generation-in-atlas-with-generation-status-failed/5690).

### Migration Authoring

This was the easy part, although getting the behavior I wanted on Oracle was not trivial due to the ergonomics of working with that database. The SqlRender translator wasn't really much help either, since it doesn't support translating queries to DBMS metadata tables.

### Migration Testing

This was the challenging part. It doesn't seem that the current migrations are currently being tested, so I decided to set up some infrastructure to do so. I wrote three tests using [Testcontainers](https://www.testcontainers.org/), one for each supported database, which runs through all the Flyway migrations on each system. The Java migrations are not currently running, because they don't implement the Interfaces that Flyway uses to autodetect them. There is probably a way to manually configure them, but I didn't spend any time on that. The results for [Oracle](https://travis-ci.org/PheMA/ohdsi-webapi/builds/614202495#L3818), [Postgres](https://travis-ci.org/PheMA/ohdsi-webapi/builds/614202495#L4250), and [SQL Server](https://travis-ci.org/PheMA/ohdsi-webapi/builds/614202495#L4472) can be seen in the Travis logs.

I ended up skipping [a lot](https://github.com/OHDSI/WebAPI/compare/master...PheMA:flyway-mode_id-migration?expand=1#diff-600376dffeb79835ede4a0b285078036R851-R867) of tests, because I couldn't figure out the required setup to get them working.


chrisknoll:
Going to close this, tho the ideas put forward here are good. We do have an integrated testing (but postgres only) and in the future will probably only support Postgres DB for WebAPI (note: broader support for the CDM platforms is a separate thing).

Thanks very much for your input and suggestions.

PR #1365, opened 2019-11-21
Move call to proxy outside (#1364)

### LINKED ISSUE #1364: impossible to create a copy of PLP
1) Try to copy any plp(prediction)

expected:The copy has been created

actual:
front:
image

 back:
WebAPI/prediction/{id}/copyRequest Method: GETStatus Code: 500{"payload":{"cause":null,"stackTrace":[],"message":"An exception ocurred: org.springframework.core.convert.ConversionFailedException","localizedMessage":"An exception ocurred: org.springframework.core.convert.ConversionFailedException","suppressed":[]},"headers":{"id":"249a112a-4afe-32c3-1543-b30a203dac25","timestamp":1573718841739}}

as a result - the copy has been created, but it's unable for editing

ssuvorov-fls:
fixes #1364

### DIFF
--- a/src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
+++ b/src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
@@ -7,28 +7,37 @@ import org.apache.commons.lang3.StringUtils;
 import org.ohdsi.analysis.Utils;
 import org.ohdsi.circe.helper.ResourceHelper;
 import org.ohdsi.hydra.Hydra;
+import org.ohdsi.webapi.analysis.AnalysisCohortDefinition;
+import org.ohdsi.webapi.analysis.AnalysisConceptSet;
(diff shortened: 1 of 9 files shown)


src/main/java/org/ohdsi/webapi/prediction/PredictionServiceImpl.java
```diff
@@ -135,6 +144,9 @@ public PredictionAnalysis createAnalysis(PredictionAnalysis pred) {
         Date currentTime = Calendar.getInstance().getTime();
         pred.setCreatedBy(getCurrentUser());
         pred.setCreatedDate(currentTime);
+        // Fields with information about modifications have to be reseted
+        pred.setModifiedBy(null);
+        pred.setModifiedDate(null);
```
wivern: Why not to add the same for other analyses types (Estimation, Pathways etc.)?

ssuvorov-fls: done

Issue #699, opened 2018-11-16
Pathways failure on RedShift


anthonysena:
In working with @leeevans we encountered a situation where a simple pathways analysis was failing to execute against RedShift. After a few attempts we found that it was related to the way the data source was configured. Our initial configuration that was failing:

`jdbc:postgresql://<redshift server>:5439/<db>?user=<user>&password=<pwd>&OpenSourceSubProtocolOverride=true;
`

We changed this to:

`jdbc:redshift://<redshift server>:5439/<db>?user=<user>&password=<pwd>&ssl=true&sslfactory=com.amazon.redshift.ssl.NonValidatingFactory`

Linking this to #686 so that we can put this type of information in the setup guide for configuring a connection to RedShift.

Issue #592, opened 2018-09-18
For Postgres sources, use OpenSourceSubProtocolOverride


alondhe:
As discussed here https://github.com/OHDSI/Broadsea/issues/11 (shoutout to @chrisknoll for helping to solve it), when using Postgres sources, the Redshift driver may try to intercept a Postgres connection, resulting in Redshift errors.

We should document this:

If using a Postgres source, append this to your postgresql JDBC url:
`?OpenSourceSubProtocolOverride=true`

anthonysena:
Noting this for #686

anthonysena:
Wiki updated to reflect this information.

PR #1382, opened 2019-12-10
redshift driver was moved to the end of drivers list 

### LINKED ISSUE #1370: Startup Failure when running unit tests: operator does not exist: boolean = integer
Possibly related to the embedded Postgresql instance that WebAPI uses for unit tests, however, I am intermittently seeing the folliwing error in the log:

```
Hibernate: select userimport0_.id as id1_54_0_, rolegroupm1_.id as id1_48_1_, userimport0_.cron as cron2_54_0_, userimport0_.is_enabled as is_enabl3_54_0_, userimport0_.executed_times as executed4_54_0_, 
  userimport0_.frequency as frequenc5_54_0_, userimport0_.is_closed as is_close6_54_0_, userimport0_.last_executed_at as last_exe7_54_0_, userimport0_.recurring_times as recurrin8_54_0_, userimport0_.recurring_until_date as recurrin9_54_0_, userimport0_.start_date as start_d10_54_0_, userimport0_.preserve_roles as preserv11_54_0_, userimport0_.provider_type as provide12_54_0_, 
  rolegroupm1_.group_dn as group_dn2_48_1_, rolegroupm1_.group_name as group_na3_48_1_, rolegroupm1_.provider as provider4_48_1_, rolegroupm1_.role_id as role_id5_48_1_, rolegroupm1_.job_id as job_id6_48_1_, rolegroupm1_.job_id as job_id6_48_0__, rolegroupm1_.id as id1_48_0__ 
from public.user_import_job userimport0_ 
left outer join public.sec_role_group rolegroupm1_ on userimport0_.id=rolegroupm1_.job_id 
where userimport0_.is_enabled=1 and userimport0_.is_closed=0
2019-11-22 22:43:17.831 WARN main org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - SQL Error: 0, SQLState: 42883
2019-11-22 22:43:17.831 ERROR main org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - ERROR: operator does not exist: boolean = integer
```

I'm guessing that in the user import tables, there's a boolean for is_enabled or is_closed that is defined as a boolean, but it's using integers to check the flag (0 or 1).

Other times, I am able to execute the tests without issue. So it's a strange, almost race-condition feeling issue.  However, if it does get tripped, the application context fails to initialize, and the unit tests fail to run.

ssuvorov-fls:
fixes #1370

### DIFF
--- a/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
+++ b/src/main/java/org/ohdsi/webapi/DataAccessConfig.java
@@ -26,6 +26,9 @@ import org.springframework.transaction.support.TransactionTemplate;
 
 import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.SQLException;
 import java.util.Properties;
 
 /**
@@ -84,6 +87,23 @@ public class DataAccessConfig {
                 logger.info("error loading {} driver. {}", driverName, ex.getMessage());
             }
         }
+
+        // Redshift driver can be loaded first because it is mentioned in manifest file -
+        // put the redshift driver at the end so that it doesn't
+        // conflict with postgres queries
+        java.util.Enumeration<Driver> drivers =  DriverManager.getDrivers();
+        while (drivers.hasMoreElements()) {
+            Driver d = drivers.nextElement();
+            if (d.getClass().getName().contains("com.amazon.redshift.jdbc")) {
+                try {
+                    DriverManager.deregisterDriver(d);
+                    DriverManager.registerDriver(d);
+                } catch (SQLException e) {
+                    throw new RuntimeException("Could not deregister redshift driver", e);
+                }
+            }
+        }
+
         return ds;
         //return new org.apache.tomcat.jdbc.pool.DataSource(pc);
     }
--- a/src/test/java/org/ohdsi/webapi/AbstractInMemoryTest.java
+++ b/src/test/java/org/ohdsi/webapi/AbstractInMemoryTest.java
@@ -1,0 +1,71 @@
+package org.ohdsi.webapi;
+
+import com.opentable.db.postgres.junit.EmbeddedPostgresRules;
+import com.opentable.db.postgres.junit.SingleInstancePostgresRule;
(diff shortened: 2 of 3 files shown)


anthonysena:
@chrisknoll - can you verify this patch?

chrisknoll:
So, the way this is fixed with redshift/postgresql driver conflicts is that you add `OpenSourceSubProtocolOverride=true` (described [here](https://stackoverflow.com/questions/31951518/redshift-and-postgres-jdbc-driver-both-intercept-jdbc-postgresql-connection-st)).

Do you think that instead of shuffling around the JDBC drivers, we instead change the JDBC url in the test cases?

Either way, I'll test this PR to see if I get the error after launching the CacheTests and approve if it works, but I wanted to ask about fixing this with the JDBC url param because that was the fix applied to #699 and #592.

chrisknoll:
I looked into this a bit, and on line 44 of your `AbstractInMemoryTest.java`:
```
    static class JdbcTemplateTestWrapper extends ExternalResource {
        @Override
        protected void before() throws Throwable {
            jdbcTemplate = new JdbcTemplate(getDataSource());
            try {
                System.setProperty("datasource.url", getDataSource().getConnection().getMetaData().getURL());
                System.setProperty("flyway.datasource.url", System.getProperty("datasource.url"));
            } catch (Exception ex) {
                throw new RuntimeException(ex);
            }
        }
    }
```
Inside where the `datasource.url` environment property is set, we could add the additional paramater to the URL there?

ssuvorov-fls:
@chrisknoll 
The test rule where connection is established for first time
`pg = EmbeddedPostgresRules.singleInstance()`
is executed before JdbcTemplateTestWrapper and I can't find possibility of changing url for this rule

chrisknoll:
Ok, I'll mark this approved for this approach, but I've opened a request under OpenTable: https://github.com/opentable/otj-pg-embedded/issues/132 requesting an option to override the JDBC url.  I'm a bit nervous about digging into the driver manager and re-ordering drivers, it seems like a hack, but if it's our only option at this point, we'll have to go with it.  Once OpenTable has the option to specify JDBC url params, we can change it.

PR #1391, opened 2019-12-18
Cohort Pathway fixes.

### LINKED ISSUE #1387: Cohort Pathways fails with >31 event cohorts defined
Because of a INT type for calculating the combo groups [here](https://github.com/OHDSI/WebAPI/blob/5357ce66e2a2a96424644dc109f9a5b363314274/src/main/resources/resources/pathway/runPathwayAnalysis.sql#L123), when you have > 31 event cohorts defined, you exceed the size of an int (POWER(2,32) > MAX_INT).  

The solution is to cast the 2 to a big int, and the result of power will also be a big int.  Then SUM() will also be a bigint, and you will avoid the overflow.

### LINKED ISSUE #1390: Pathway date replacements leads to duplicate event cohort records
### Expected Results
When an event cohort with multiple micro-episodes (ie: 1 day intervals) get combined due to a larger collapse window, only distinct event cohort episodes should be produced.

### Actual Results
If you have an event cohort with multiple micro-episodes that get collapsed into a start-end from `#date_replacements`, it will create duplicate records. This breaks the `SUM()` of the event_cohort_index into the combo_id logic of the pathway analysis.

### Steps to reproduce:
Create an event cohort which has multiple small intervals within a 30 day period.
create a pathway analysis to collapse events within 30 days
Observe: the event cohort episodes do not get combined via a `distinct` operation, but rather each individual record which had the cohort start/end dates replaced remains in the result, leading to duplicate records.

chrisknoll:
Cast result of power(2,cohort_index) to bigint.
Remove duplicate event cohort records after date replacements.

Fixes #1387
Fixes #1390

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -118,7 +118,7 @@ events AS (
 	LEFT JOIN #collapsed_dates_events e ON e.subject_id = tp.subject_id
 	WHERE (e.cohort_start_date <= tp.cohort_date AND e.cohort_end_date >= tp.next_cohort_date)
 ) 
-SELECT SUM(POWER(cast(2 as bigint), e.event_cohort_index)) as combo_id,  subject_id , cohort_start_date, cohort_end_date
+SELECT SUM(POWER(2, e.event_cohort_index)) as combo_id,  subject_id , cohort_start_date, cohort_end_date
 into #combo_events
 FROM events e
 GROUP BY subject_id, cohort_start_date, cohort_end_date;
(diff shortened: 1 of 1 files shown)


src/main/resources/resources/pathway/runPathwayAnalysis.sql
```diff
@@ -120,7 +118,7 @@ events AS (
 	LEFT JOIN #collapsed_dates_events e ON e.subject_id = tp.subject_id
 	WHERE (e.cohort_start_date <= tp.cohort_date AND e.cohort_end_date >= tp.next_cohort_date)
 ) 
-SELECT SUM(POWER(2, e.event_cohort_index)) as combo_id,  subject_id , cohort_start_date, cohort_end_date
+SELECT SUM(POWER(cast(2 as bigint), e.event_cohort_index)) as combo_id,  subject_id , cohort_start_date, cohort_end_date
```
anthonysena: In #1388, you have an outer CAST on the SUM in this SELECT statement: 
https://github.com/OHDSI/WebAPI/pull/1388/files#diff-02c181aadb1a447251b457feccfa48eaL123. I think you'll want to carry that over here.

chrisknoll: Hey, i had forgotten I already had a separate PR to fix that one.

I think we can just go forward with this one, close the other, and if there's a problem with impala they can open a new issue on it.

Issue #1400, opened 2020-01-03
Miredot plugin fails during Maven package step


pavgra:
Miredot fails to generate documentation for:
1. `org.ohdsi.webapi.cohortcharacterization.dto.GenerationResults` -> `org.ohdsi.webapi.cohortcharacterization.report.Report` -> `org.ohdsi.webapi.cohortcharacterization.report.ExportItem` due to generic parameter referencing the same class (miredot goes into infinite recursion and throws stackoverflow)
1. EstimationAnalysisImpl -> most possibly `org.ohdsi.webapi.estimation.specification.EstimationAnalysisSettingsImpl`

PR #1401, opened 2020-01-03
Workaround for methods that cause Miredot failure

pavgra:
fixes https://github.com/OHDSI/WebAPI/issues/1400

That's not a good solution but 1) proper solution would require a bunch of changes in interfaces/classes structure, 2) I do not think that the inability of the plugin to handle existing cases is a good sign, the cases are really not that complicated and fully valid - I would think of replacing the plugin in 3.0

### DIFF
--- a/src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
+++ b/src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
@@ -1,6 +1,8 @@
 package org.ohdsi.webapi.estimation;
 
 import com.odysseusinc.arachne.commons.utils.ConverterUtils;
+import com.qmino.miredot.annotations.MireDotIgnore;
+import com.qmino.miredot.annotations.ReturnType;
 import org.ohdsi.webapi.Constants;
 import org.ohdsi.webapi.common.SourceMapKey;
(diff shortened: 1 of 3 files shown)


pavgra:
tagging @chrisknoll for awareness since he raised the issue in OHDSI forums  (https://forums.ohdsi.org/t/miredot-error-setting-property-antijarlocking-to-true-did-not-find-a-matching-property-webapi/7222/4)

src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
```diff
@@ -181,7 +185,7 @@ public Response download(@PathParam("id") int id, @QueryParam("packageName") Str
         packageName = "estimation" + String.valueOf(id);
     }
 
-    EstimationAnalysisImpl analysis = this.exportAnalysis(id);
+    EstimationAnalysisImpl analysis = null;//this.exportAnalysis(id);
```
anthonysena: I think this is an error introduced during testing? Please review and correct. Thanks!

pavgra: Ouch, fixed

PR #1415, opened 2020-01-13
Fix syntax for INTO statement for #collapsed_dates_events

chrisknoll:


chrisknoll:
Interestingly, we didn't see this on PDW because the translation rules 'fixed it' into a CTAS.

On postgresql, the statement is transformed to a CTAS, but the table is not made as a TEMP table. 

I don't think we need to look into any SQLRender fixes, the syntax from prior to this change was just wrong.

anthonysena:
Tagging @olga-ganina - we should put this into the 2.7.6 branch as well as master .

PR #1420, opened 2020-01-17
Exposes additional thread pool options

wivern:
* Exposes Thread Pool options such as queueCapacity, threadGroupName and threadNamePrefix
* Publish Thread Pool Executor bean to JMX

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,9 @@
 
     <spring.batch.taskExecutor.corePoolSize>10</spring.batch.taskExecutor.corePoolSize>
     <spring.batch.taskExecutor.maxPoolSize>20</spring.batch.taskExecutor.maxPoolSize>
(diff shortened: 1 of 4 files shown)


olga-ganina:
Fixes #1424

pavgra:
@olga-ganina / @wivern, how does this *fix* the issue (I'm referring to @olga-ganina 's comment that would close the issue as soon as the PR accepted)? I do not see any behavioral changes in the PR. The only changed setting is queueCapacity set to `2147483647`. However, from reading the issue I assume that it's default value has already been equal to MAXINT. Otherwise, PR provides only additional configuration / debugging capabilities from what I see

wivern:
@pavgra You're right in this is not about resolving the issue but providing additional capabilities to collect information and tune the thread pool in a proper way for a concrete environment. Therefore I've changed the label to avoid autoclosing the issue.

PR #1427, opened 2020-01-22
Fixes showing CC design as actual when cohort changed

### LINKED ISSUE #1389: CC design is shown as actual when cohort design changed
1. Create CC
2. Save
3. Run
4. Observe `Same as now` caption in the Generations' Design column for the execution
5. Change cohort utilized in the CC
6. The `Same as now` caption is still in place

wivern:
Fixes #1389


wivern:
Yes, you are completely right. Event handler just tracking the fact of cohort definition was changed. It is possible redundant characterization design to be reloaded when a cohort expression wasn't changed. But to avoid such case we have to store somewhere all cohort expression hashes included in the characterization and compare them with actual values. From my point of view, there is too much effort with little benefits.

PR #1428, opened 2020-01-22
query parameter for hiding completed jobs was added 

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/539

### DIFF
--- a/src/main/java/org/ohdsi/webapi/job/NotificationController.java
+++ b/src/main/java/org/ohdsi/webapi/job/NotificationController.java
@@ -5,10 +5,12 @@ import org.springframework.core.convert.support.GenericConversionService;
 import org.springframework.stereotype.Controller;
 import org.springframework.transaction.annotation.Transactional;
 
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/job/NotificationController.java
```diff
@@ -31,8 +33,9 @@
     @Path("/")
     @Produces(MediaType.APPLICATION_JSON)
     @Transactional(readOnly = true)
-    public List<JobExecutionResource> list() {
-        return service.findLast10().stream().map(this::toDTO).collect(Collectors.toList());
+    public List<JobExecutionResource> list(
+            @QueryParam("hideCompleted") @DefaultValue("false") boolean hideCompleted) {
```
wivern: According to general REST recommendations, it's better to use a `snake_case` or a `spinal-case`. To be consistent pls change query parameter to `snake_case` as one is widely used in the WebAPI. 
Second approach is the: why the API method is limited to just show/hide jobs with completed status. Why not to resolve list of statuses that should be filtered out?

src/main/java/org/ohdsi/webapi/job/NotificationController.java
```diff
@@ -31,8 +35,15 @@
...
     @Produces(MediaType.APPLICATION_JSON)
     @Transactional(readOnly = true)
-    public List<JobExecutionResource> list() {
-        return service.findLast10().stream().map(this::toDTO).collect(Collectors.toList());
+    public List<JobExecutionResource> list(
+            @QueryParam("hide_statuses") String hideStatuses) {
+        List<BatchStatus> statuses = new ArrayList<>();
+        if (StringUtils.isNotEmpty(hideStatuses)) {
+            for (String status : hideStatuses.split(",")) {
+                statuses.add(BatchStatus.valueOf(status));
```
wivern: Could you add some check in a case when status value is odd and cannot be converted to a `BatchStatus`?

src/main/java/org/ohdsi/webapi/job/NotificationController.java
ssuvorov-fls: done

src/main/java/org/ohdsi/webapi/job/NotificationController.java
ssuvorov-fls: try/catch was added

PR #1434, opened 2020-01-24
index for vocabulary schema was removed

### LINKED ISSUE #1327: Results schema script - adding vocabulary table index
I'm facing an issue while applying the results schema script via the `ddl/results` endpoint. Specifically this statement fails on PDW:

````
CREATE INDEX idx_concept_id_varchar ON dbo.concept (CAST(concept_id AS VARCHAR));
````

Tracing through the git history lead me to this issue & corresponding changes: _Heracles performance issue when joining strata to concept_ #1068. In thinking about this, my opinion is that this type of index creation would be more appropriate to store in the CommonDataModel repo as that is where indicies are applied for the vocabulary tables. For example: https://github.com/OHDSI/CommonDataModel/blob/master/PostgreSQL/OMOP%20CDM%20postgresql%20pk%20indexes.txt#L189. I do not thing that the WebAPI should direct changes to the vocabulary tables as was done here. 

Tagging @pavgra @chrisknoll for thoughts on this.

ssuvorov-fls:
fixes #1327


anthonysena:
Index moved to CDM repo: https://github.com/OHDSI/CommonDataModel/pull/324.

PR #1437, opened 2020-01-27
Restoring backwards compatibility for rest calls. Fixes #1436.

### LINKED ISSUE #1436: REST endpoint backwards compatibility
As a rule, we should maintain backwards compatibility for any REST end-points that are published on the WebAPI. While ATLAS is the main consumer of WebAPI, there are a number of other projects (https://github.com/OHDSI/ROhdsiWebApi to name one) that will break if we do not maintain backwards compatibility. The following endpoint have breaking changes between 2.7.x and the current master version of 2.8:

- `/WebAPI/source/priorityVocabulary`: This endpoint was removed entirely as part of changes in https://github.com/OHDSI/WebAPI/commit/83bba829ceaa85d94c94678d8d8fb8009dd36cd0#diff-1ff5a27a9736edb2375d9bb0eca8f6cfL170. This needs to be re-added to the code base.
- `/WebAPI/cohort-characterization/generation/{generationId}/result`: The HTTP Verb was changed from a GET to POST as part of these changes: https://github.com/OHDSI/WebAPI/commit/5357ce66e2a2a96424644dc109f9a5b363314274#diff-16e6e60cf5961441c04b52de614b6d10R260. Instead of a POST, we should consider revising to a GET with a query string to allow for filtering based on a threshold value.

anthonysena:
Restores the REST endpoints removed from 2.7.x.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
@@ -257,6 +257,15 @@ public class CcController {
         return service.getCCResultsTotalCount(generationId);
     }
 
+    @GET
+    @Path("/generation/{generationId}/result")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    public List<CcResult> getGenerationsResults(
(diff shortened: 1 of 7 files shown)


anthonysena:
Updated via https://github.com/OHDSI/WebAPI/pull/1437/commits/46a8b84473b7cad656282e93326c9d2c9556542c to include setting permissions for restored endpoints. I tested on PG but not other dialects.

PR #1446, opened 2020-01-30
Remove Maven Central Repo from pom.xml

YaroslavTir:
remove http://repo.maven.apache.org/maven2
http is not longer supported by maven central. Default repo point to the correct https version 
https://stackoverflow.com/a/59764670/1167673

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -371,11 +371,6 @@
 
   <repositories>
     <repository>
-      <id>central</id>
-      <name>Central Repository</name>
-      <url>https://repo.maven.apache.org/maven2/ </url>
-    </repository>
-    <repository>
       <id>ohdsi</id>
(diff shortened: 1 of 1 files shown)


chrisknoll:
I'm not sure the solution here should be remove the central repo from the pom.xml.  Doesn't this depend on a specific version of Maven (>3.6)?  The reference to the stack overflow states that on Ubuntu, it didn't fix it.

Wouldn't it be safer to just to define the maven central repo URL as https in pom.xml?

YaroslavTir:
@chrisknoll this change is not required but I want all our repository to be consistent.

chrisknoll:
Understood, I've updated the title of the PR to reflect the change (since the reference in the pom was already using HTTPS, it was confusing to say 'fix http'.

Issue #1409, opened 2020-01-09
Temp tables aren't cleaned up


aklochkova:
Example for BigQuery:
after IR cancellation there is a bunch of temp tables

For Impala: after Heracles generation temp tables are left

need to check other databases

KSadomtsev:
![image](https://user-images.githubusercontent.com/56290675/75690229-89431e00-5cb3-11ea-95cc-b05363fbb9b7.png)
can see the same problem for IMPALA (PLP, PLE)

olga-ganina:
Original request and issue was about Netezza.
Do not have any reported problems for other database for now, so closing this one.

Issue #1067, opened 2019-04-25
Growing number of TEMP_COHORT tables in results schema for Netezza dataset


spbima:
Results schema has a growing number of TEMP_COHORT tables for Netezza dataset because these tables are not being removed

![netezza](https://user-images.githubusercontent.com/804934/56727495-74276580-6759-11e9-89c0-ca5fe06a680d.jpg)

Detected in 2.7.1 version

AntonStepanof:
cannot reproduce on http://odysseusovh02.odysseusinc.com/ environment

anthonysena:
@spbima - can you confirm if this issue still persists after the v2.7.2 release? If not can you please close? Thanks!

olga-ganina:
Relates to #1409

anthonysena:
I can confirm that this is still an issue in our environment. We recently had an issue where we ran out of disk space for DB and the job crashed. Here is an example error message:

````
org.springframework.jdbc.UncategorizedSQLException: StatementCallback; uncategorized SQLException for SQL [CREATE TABLE  CDM.ohdsi_results.temp_cohort_szthj7nv
(COHORT_DEFINITION_ID int NOT NULL,
   subject_id bigint NOT NULL,
  cohort_start_date date NOT NULL,
  cohort_end_date date NOT NULL
)
WITH (DISTRIBUTION = HASH(subject_id))]; 

SQL state [S0002]; error code [1105]; Could not allocate space for object 'sys.syscolpars'.'clst' in database 'DB_1d662369e61c4891ac8717a6458e0ae9' because the 'PRIMARY' filegroup is full. Create disk space by deleting unneeded files, dropping objects in the filegroup, adding additional files to the filegroup, or setting autogrowth on for existing files in the filegroup.; nested exception is com.microsoft.sqlserver.jdbc.SQLServerException: Could not allocate space for object 'sys.syscolpars'.'clst' in database 'DB_1d662369e61c4891ac8717a6458e0ae9' because the 'PRIMARY' filegroup is full. Create disk space by deleting unneeded files, dropping objects in the filegroup, adding additional files to the filegroup, or setting autogrowth on for existing files in the filegroup:
````

As you might imagine, this is not an expected error but none-the-less, the job crashes as a result and it leaves behind temporary tables. We experienced this with cohort characterization but presumably this would happen for any generation process potentially.

PR #1448, opened 2020-02-04
Added cleanup for temp tables in PLP/PLE (only Netezza)

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1409


olga-ganina:
Fixes #1067

PR #1459, opened 2020-02-14
added parameter to execute cohort generations in parallel or sequentially

### LINKED ISSUE #1435: Running cohort generations in parallel
On 2.8, when running a characterization (CC, pathways, IR), I'd expect for the cohort generations to happen serially. I believe that was the intention when we introduced the changes in in this PR : https://github.com/OHDSI/WebAPI/pull/1298 to address this issue: #1219. 

Instead what we're observing in our environment is that cohort generation processes are kicked off in parallel which is not desirable since this behavior can overwhelm the DB server and cause other issues.

ssuvorov-fls:
fixes #1435


src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
```diff
@@ -76,46 +87,66 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
...
+        if (useAsyncCohortGeneration) {
+            List<CompletableFuture> executions = cohortDefinitions.stream()
+                    .map(cd ->
+                            CompletableFuture.supplyAsync(() -> generateCohort(cd, source, resultSchema, targetTable),
+                                    Executors.newSingleThreadExecutor()
+                            )
+                    ).collect(Collectors.toList());
+            CompletableFuture.allOf(executions.toArray(new CompletableFuture[]{})).join();
+        } else {
+            CompletableFuture.runAsync(() ->
```
wivern: Does `CompletableFuture` required here? Why we can't simply run each cohort one by one?

src/main/java/org/ohdsi/webapi/util/StatementCancel.java
```diff
@@ -9,20 +9,22 @@
   private Statement statement;
   private boolean canceled = false;
 
-  public void setStatement(Statement statement) {
-
+  public synchronized void setStatement(Statement statement) {
```
wivern: Why we need synchronized block here, keeping in mind each thread has its own instance of `StatementCancel`?

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
```diff
@@ -179,26 +179,6 @@ public void checkCachingWithEmptyResultSet() {
         Assert.assertNotNull("Empty result set is cached", generationCache);
     }
 
-    @Test
```
wivern: Question: Why we want to remove this test?

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
ssuvorov-fls: CompletableFuture allows to run cohort generating in another thread. Running cohort one by one without CompletableFuture will use current worker thread

src/main/java/org/ohdsi/webapi/util/StatementCancel.java
ssuvorov-fls: fixed

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
ssuvorov-fls: the logic of generating new cache was changed and now previous caches do not affect new one

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
wivern: That's true and why different thread is required?

src/test/java/org/ohdsi/webapi/generationcache/GenerationCacheTest.java
wivern: Still, not clear to me. As far as I understood the test verifies that cache isn't computed twice for the same data.
What changes breaks this rule?
If test slightly different from the behavior of the current implementation why not to fix the test rather than completely remove one?

ssuvorov-fls: Testing that cache isn't computed twice for the same data performs in test "generateCohort"

src/main/java/org/ohdsi/webapi/cohortcharacterization/GenerateLocalCohortTasklet.java
ssuvorov-fls: To prevent blocking of the job thread during cohort generation

wivern:
@anthonysena Could you inspect how this conforms to your expectations?

PR #1464, opened 2020-02-20
Fixes Non-optimal evaluation of priority daimons

### LINKED ISSUE #1429: Non-optimal evaluation of priority daimons
1. `SourceController.getPriorityDaimons` iterates over all source diamons and for each of them executes `getPrioritySourceForDaimon`
2. The `getPrioritySourceForDaimon` relies on `sourceAccessor.hasAccess(source)`
3. Therefore, for a single source, we can do the same security check 6 times (number of diamons) which is redundant. 
**However, the biggest problem** takes place when some Shiro Realm points e.g. to Active Directory that is offline. Then, the realm would wait until the connection attempt reaches timeout and the security check would take ~20sec for the offline Active Directory. But, due to the repetition of the check, the total execution time would be 20sec * 6=120 sec = 2 min instead of just 20 sec.

wivern:
Fixes #1429

### DIFF
--- a/src/main/java/org/ohdsi/webapi/source/SourceService.java
+++ b/src/main/java/org/ohdsi/webapi/source/SourceService.java
@@ -124,16 +124,23 @@ public class SourceService extends AbstractDaoService {
 
     public Map<SourceDaimon.DaimonType, Source> getPriorityDaimons() {
 
-        Map<Integer, Boolean> checkedSources = new HashMap<>();
+        class SourceValidator {
+            private Map<Integer, Boolean> checkedSources = new HashMap<>();
+
+            private boolean isSourceAvaialble(Source source) {
+                return checkedSources.computeIfAbsent(source.getSourceId(),
+                        v -> sourceAccessor.hasAccess(source) && connectionAvailability.computeIfAbsent(source, SourceService.this::checkConnectionSafe));
+            }
+        }
+
+        SourceValidator sourceValidator = new SourceValidator();
         Map<SourceDaimon.DaimonType, Source> priorityDaimons = new HashMap<>();
         Arrays.asList(SourceDaimon.DaimonType.values()).forEach(d -> {
 
             List<Source> sources = sourceRepository.findAllSortedByDiamonPrioirty(d);
-            Optional<Source> source = sources.stream().filter(s -> checkedSources.computeIfAbsent(s.getSourceId(),
-                    v -> sourceAccessor.hasAccess(s)
-                            && connectionAvailability.computeIfAbsent(s, this::checkConnectionSafe)))
+            Optional<Source> source = sources.stream().filter(sourceValidator::isSourceAvaialble)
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/source/SourceService.java
```diff
@@ -126,6 +122,22 @@ public Source getPrioritySourceForDaimon(SourceDaimon.DaimonType daimonType) {
...
 
+    public Map<SourceDaimon.DaimonType, Source> getPriorityDaimons() {
+
+        Map<Integer, Boolean> checkedSources = new HashMap<>();
+        Map<SourceDaimon.DaimonType, Source> priorityDaimons = new HashMap<>();
+        Arrays.asList(SourceDaimon.DaimonType.values()).forEach(d -> {
+
+            List<Source> sources = sourceRepository.findAllSortedByDiamonPrioirty(d);
+            Optional<Source> source = sources.stream().filter(s -> checkedSources.computeIfAbsent(s.getSourceId(),
+                    v -> sourceAccessor.hasAccess(s)
```
YaroslavTir: this part is not clear. I personally don't like it when in filter happens something besides checking. I understand that is it is webapi approach, but can you extract the check to separate method, with an understandable name.
exp: 
```
    public Map<SourceDaimon.DaimonType, Source> getPriorityDaimons() {

        Map<Integer, Boolean> checkedSources = new HashMap<>();
        Map<SourceDaimon.DaimonType, Source> priorityDaimons = new HashMap<>();

        Arrays.asList(SourceDaimon.DaimonType.values()).forEach(d -> {

            Optional<Source> source = sourceRepository.findAllSortedByDiamonPrioirty(d).stream()
                    .filter(s -> isSourceAvailable(s, checkedSources, connectionAvailability))
                    .findFirst();

            source.ifPresent(s -> priorityDaimons.put(d, s));
        });
        return priorityDaimons;
    }

    private Boolean isSourceAvailable(Source s, Map<Integer, Boolean> sourceCheckedCache, Map<Source, Boolean> connectionAvailabilityCache) {
        return sourceCheckedCache.computeIfAbsent(
                s.getSourceId(),
                v -> sourceAccessor.hasAccess(s) && connectionAvailabilityCache.computeIfAbsent(s, this::checkConnectionSafe));
    }
```

src/main/java/org/ohdsi/webapi/source/SourceService.java
```diff
@@ -126,6 +122,22 @@ public Source getPrioritySourceForDaimon(SourceDaimon.DaimonType daimonType) {
...
+                            && connectionAvailability.computeIfAbsent(s, this::checkConnectionSafe)))
+                    .findFirst();
+            source.map(s -> priorityDaimons.put(d, s));
```
YaroslavTir: I guess ifPresent fits better here.

PR #1466, opened 2020-02-25
added checks for null values (#1209)

### LINKED ISSUE #1209: NPE: GenerationJobExecutionListener unhandled exception within open transaction
Under some circumstances unhandled exception might occur at line 81 which breaks BatchJob logic and leave uncommitted transaction....
actual code line:
info.setExecutionDuration((int) (je.getEndTime().getTime() - je.getStartTime().getTime()));

At least must be  handled and logged properly 

Exception encountered in afterStep callback
java.lang.NullPointerException
    at org.ohdsi.webapi.cohortdefinition.GenerationJobExecutionListener.afterJob(GenerationJobExecutionListener.java:81)
    at org.springframework.batch.core.listener.CompositeJobExecutionListener.afterJob(CompositeJobExecutionListener.java:60)
    at org.springframework.batch.core.job.AbstractJob.execute(AbstractJob.java:352)
    at org.springframework.batch.core.launch.support.SimpleJobLauncher$1.run(SimpleJobLauncher.java:135)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
2019-06-21 10:16:39.586 INFO taskExecutor-5 org.springframewo

ssuvorov-fls:
fixes #1209

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
+++ b/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
@@ -15,8 +15,6 @@
  */
 package org.ohdsi.webapi.cohortdefinition;
 
-import java.util.*;
-
 import org.apache.commons.lang3.StringUtils;
 import org.ohdsi.sql.SqlRender;
 import org.ohdsi.sql.SqlTranslate;
@@ -33,6 +31,12 @@ import org.springframework.transaction.TransactionDefinition;
 import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.support.DefaultTransactionDefinition;
 import org.springframework.transaction.support.TransactionTemplate;
+
+import java.util.Calendar;
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+import java.util.Objects;
 
 import static org.ohdsi.webapi.Constants.Params.TARGET_DATABASE_SCHEMA;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
```diff
@@ -81,29 +85,33 @@ public void afterJob(JobExecution je) {
 		TransactionStatus completeStatus = this.transactionTemplate.getTransactionManager().getTransaction(completeTx);
```
YaroslavTir, line 85: formating is still wrong.

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
```diff
@@ -81,29 +85,33 @@ public void afterJob(JobExecution je) {
 		TransactionStatus completeStatus = this.transactionTemplate.getTransactionManager().getTransaction(completeTx);      
 		CohortDefinition df = this.cohortDefinitionRepository.findOne(defId);
 		CohortGenerationInfo info = findBySourceId(df.getGenerationInfoList(), sourceId);
-		info.setExecutionDuration((int)(je.getEndTime().getTime() - je.getStartTime().getTime()));
-		info.setStatus(GenerationStatus.COMPLETE);
+		if (Objects.nonNull(info)) {
+			if (Objects.nonNull(je.getEndTime()) && Objects.nonNull(je.getStartTime())) {
```
YaroslavTir, line 89: this does not help to find the real issue, it just hides it even deep :)  I would add logger and extract setExecutionDuration method in separate method like this:
```
	private void setExecutionDurationIfPossible(JobExecution je, CohortGenerationInfo info) {

		if (Objects.isNull(je.getEndTime()) || Objects.isNull(je.getStartTime())) {
			log.error("Cannot set duration time for cohortGenerationInfo[{}]. startData[{}] and endData[{}] cannot be empty.", info.getId(), je.getStartTime(), je.getEndTime());
			return;
		}
		info.setExecutionDuration((int) (je.getEndTime().getTime() - je.getStartTime().getTime()));

	}
```

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
```diff
@@ -81,29 +85,33 @@ public void afterJob(JobExecution je) {
 		TransactionStatus completeStatus = this.transactionTemplate.getTransactionManager().getTransaction(completeTx);      
 		CohortDefinition df = this.cohortDefinitionRepository.findOne(defId);
 		CohortGenerationInfo info = findBySourceId(df.getGenerationInfoList(), sourceId);
-		info.setExecutionDuration((int)(je.getEndTime().getTime() - je.getStartTime().getTime()));
-		info.setStatus(GenerationStatus.COMPLETE);
+		if (Objects.nonNull(info)) {
```
YaroslavTir, line 88: I don't think it is good idea just skip the part in case of empty info, this can be a reason for the frozen job I would do something like this:
```
CohortGenerationInfo info = findBySourceId(df, sourceId);
....
	private CohortGenerationInfo findBySourceId(CohortDefinition df, Integer sourceId) {

		df.getGenerationInfoList().stream()
				.filter(info -> info.getId().getSourceId().equals(sourceId))
				.findFirst()
				.orElseThrow(() -> new IllegalStateException(String.format("Cannot find cohortGenerationInfo for cohortDefinition[{}] sourceId[%s]", df.getId(), sourceId)));
	}
```

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
ssuvorov-fls: fixed

PR #1468, opened 2020-02-27
added invalidation of jobs in status "STARTING"

### LINKED ISSUE #1467: WebAPI does not invalidate jobs after restart
Sometimes generation jobs hang in 'STARTING' state in JOBS list. Restarting of WebAPI does not help too.

Workaround is to run the sql script manually:
```
update ${ohdsiSchema}.batch_job_execution set status='FAILED', exit_code='FAILED', exit_message='Invalidated by system' where status='STARTING';
```

ssuvorov-fls:
fixes #1467

### DIFF
--- a/src/main/java/org/ohdsi/webapi/JobInvalidator.java
+++ b/src/main/java/org/ohdsi/webapi/JobInvalidator.java
@@ -33,10 +33,7 @@ public class JobInvalidator {
     @PostConstruct
     private void invalidateGenerations() {
         transactionTemplateRequiresNew.execute(s -> {
-            jobExecutionDao.getRunningJobExecutions().stream()
-                    // do not invalidate currently running jobs such as warming cache
-                    .filter(job -> BatchStatus.STARTING.equals(job.getStatus()))
-                    .forEach(this::invalidationJobExecution);
+            jobExecutionDao.getRunningJobExecutions().forEach(this::invalidationJobExecution);
             return null;
         });
     }
--- a/src/main/java/org/ohdsi/webapi/JobInvalidator.java
+++ b/src/main/java/org/ohdsi/webapi/JobInvalidator.java
@@ -1,8 +1,7 @@
 package org.ohdsi.webapi;
 
-import java.util.Calendar;
-import javax.annotation.PostConstruct;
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/JobInvalidator.java
```diff
@@ -20,22 +22,25 @@
...
 
     @PostConstruct
     private void invalidateGenerations() {
         transactionTemplateRequiresNew.execute(s -> {
-            jobExplorer.getJobNames().forEach(name ->
-                    jobExplorer
-                            .findRunningJobExecutions(name)
-                            .forEach(this::invalidationJobExecution));
+            jobExecutionDao.getRunningJobExecutions().stream()
+                    // do not invalidate currently running jobs such as warming cache
```
wivern: By such filtering you leaving jobs running on the Execution Engine in the inconsistent state.
To preserve warm cache jobs from improper invalidation it's better to align bean creation to the right order. Probably using `dependsOn` annotation or smth else.

src/main/java/org/ohdsi/webapi/JobInvalidator.java
```diff
@@ -20,22 +22,25 @@
     private final JobExplorer jobExplorer;
```
wivern: `JobExplorer` could be completely replaced by the `SearchableJobExecutionDao`

src/main/java/org/ohdsi/webapi/JobInvalidator.java
ssuvorov-fls: Execution engine jobs are invalidated in class ScriptExecutionServiceImpl.
I don't think that bean creation ordering is a good choice because new beans with new jobs can be created in future and its developers can simply forget about possible conflicts

src/main/java/org/ohdsi/webapi/JobInvalidator.java
ssuvorov-fls: agree, fixed

ssuvorov-fls:
@anthonysena Hello, why did you close this pr?

PR #1477, opened 2020-03-06
added atlas saml authorization support

### LINKED ISSUE #396: SAML support for authentication

ssuvorov-fls:
fixes #396


src/main/java/org/ohdsi/webapi/security/SSOController.java
```diff
@@ -0,0 +1,70 @@
...
+        response.setHeader("Expires", "0");
+        org.apache.commons.io.IOUtils.copy(is, response.getOutputStream());
+        response.flushBuffer();
+    }
+
+    @GET
+    @Path("/slo")
+    public Response logout() throws URISyntaxException {
+        return Response.status(302)
+                .header("Access-Control-Allow-Origin", "http://localhost")
```
YaroslavTir, line 66: this is right to set localhost as origin?

ssuvorov-fls: You are right. I changed it to parameter, which is already used by cors filter

src/main/java/org/ohdsi/webapi/security/SSOController.java
```diff
@@ -0,0 +1,72 @@
...
+    public void samlMetadata(@Context HttpServletResponse response) throws IOException {
+
+        ClassPathResource resource = new ClassPathResource(metadataLocation);
+        final InputStream is = resource.getInputStream();
+        response.setContentType("application/xml");
+        response.setHeader("Content-type", "application/xml");
+        response.setHeader("Cache-Control", "no-cache, no-store, must-revalidate");
+        response.setHeader("Pragma", "no-cache");
+        response.setHeader("Expires", "0");
+        org.apache.commons.io.IOUtils.copy(is, response.getOutputStream());
```
YaroslavTir, line 60: you can be replaced with an import

src/main/java/org/ohdsi/webapi/shiro/filters/SendTokenInUrlFilter.java
```diff
@@ -21,12 +25,13 @@ public SendTokenInUrlFilter(String url) {
   @Override
   protected boolean preHandle(ServletRequest request, ServletResponse response) throws Exception {
     String jwt = (String)request.getAttribute("TOKEN");
-    if (Guard.isNullOrEmpty(jwt)) {
-      WebUtils.toHttp(response).sendRedirect(url);
-    }
-    else {
-      WebUtils.toHttp(response).sendRedirect(url.replaceAll("/+$", "") + "/" + jwt);
+    String client = (String)request.getAttribute(AUTH_CLIENT_ATTRIBUTE);
+    String clientValue = Guard.isNullOrEmpty(client) ? AUTH_CLIENT_ALL : client;
```
YaroslavTir, line 29: don't really like this custom Guard class, there is a standard apache StringUtils.isNotEmpty()

src/main/java/org/ohdsi/webapi/shiro/filters/auth/SamlHandleFilter.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @see org.apache.shiro.web.filter.authc.AuthenticatingFilter#createToken(ServletRequest,
+     *      ServletResponse)
+     */
+    @Override
+    protected AuthenticationToken createToken(ServletRequest servletRequest,
+                                              ServletResponse servletResponse) throws Exception {
+        final ShiroHttpServletRequest request = (ShiroHttpServletRequest) servletRequest;
+        HttpSession session = request.getSession();
```
YaroslavTir, line 42: do we need this variable here? we don't really use it in this method, only in if condition. even in this case, you can inline it  `if (request.getSession() != null)`

src/main/java/org/ohdsi/webapi/security/SSOController.java
```diff
@@ -0,0 +1,72 @@
...
+        response.flushBuffer();
+    }
+
+    @GET
+    @Path("/slo")
+    public Response logout() throws URISyntaxException {
+        return Response.status(302)
+                .header("Access-Control-Allow-Origin", this.origin)
+                .location(new URI(sloUri))
+                .build();
```
YaroslavTir, line 70: you can use constants for status and header
```        return Response.status(HttpConstants.TEMP_REDIRECT)
                .header(HttpConstants.ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, this.origin)
                .location(new URI(sloUri))
                .build();
```

pom.xml
```diff
@@ -682,6 +691,11 @@
       <artifactId>pac4j-oidc</artifactId>
       <version>1.9.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.pac4j</groupId>
+      <artifactId>pac4j-saml</artifactId>
+      <version>1.9.2</version>
```
YaroslavTir, line 697: can you extract pac4j.version as param

YaroslavTir:
it would be nice to add a description to the Read.me which parameters need to be configured. Maybe this is a separate story because we do not have a description for any type of connection

src/main/java/org/ohdsi/webapi/security/SSOController.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/shiro/filters/SendTokenInUrlFilter.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/security/SSOController.java
ssuvorov-fls: fixed

pom.xml
ssuvorov-fls: fixed

PR #1492, opened 2020-03-25
I18n characterization

YaroslavTir:


anthonysena:
@YaroslavTir @wivern - where are these localization files being used? It is not clear and I do not see an issue associated with these changes.

anthonysena:
Nevermind - spoke with @gklebanov and I'm set. Let's keep these changes out of the 2.8 release until we can discuss localization further. Thanks!

PR #1508, opened 2020-04-14
Split zip archives to chunks to reduce single file size

YaroslavTir:
Fixes an issue related to Huge results for PLE/PLP cause Postgres failure during Download request from ATLAS


pom.xml
```diff
@@ -182,7 +182,7 @@
     <!-- Use "-" for files without extension, "*" for all files, extension must not include a leading dot. Use comma to separate values.
      In case of "*" other values will be ignored -->
     <sensitiveinfo.analysis.extensions>txt</sensitiveinfo.analysis.extensions>
-
+    <sensitiveinfo.analysis.zip-chunk-size-mb>700</sensitiveinfo.analysis.zip-chunk-size-mb>
```
wivern, line 185: Keeping in mind parameters could be externalized with environment variables it's not safe to use underscore or hyphen or similar symbols. It's better to use snake-case if necessary.

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileContentSensitiveInfoServiceImpl.java
```diff
@@ -56,33 +62,34 @@ public void init() {
...
-                    // Path contains information about archive volume, must be deleted
-                    // because we create new archive without volumes
-                    iter.remove();
+            Map<Path, AnalysisResultFileContent> savedFiles = saveFilesToProcessMultivolumeArchives(temporaryDir, source.getFiles());
+
+            savedFiles.keySet().stream()
+                    .filter(path -> !isArchiveVolume(path))
+                    .forEach(path -> processFile(path, variables));
+
+            source.getFiles().clear();
```
wivern, line 82: Not sure the logic wasn't changed with PR, we can have archive inside archive including multi-module

src/main/java/org/ohdsi/webapi/executionengine/service/AnalysisResultFileContentSensitiveInfoServiceImpl.java
```diff
@@ -94,16 +101,41 @@ public AnalysisResultFileContentList filterSensitiveInfo(AnalysisResultFileConte
...
+    private AnalysisResultFileContent getAnalysisResultFileOrCreatNewForZipVolume(Path path, Map<Path, AnalysisResultFileContent> filePathWithResultFileContent) {
+
+        if (isArchiveVolume(path)) {
+
+            ExecutionEngineAnalysisStatus execution = filePathWithResultFileContent.values().stream().map(f -> f.getAnalysisResultFile().getExecution()).findAny().orElse(null);
+
+            AnalysisResultFileContent analysisResultFileContent = new AnalysisResultFileContent();
+            AnalysisResultFile analysisResultFile = new AnalysisResultFile();
+            analysisResultFile.setFileName(path.getFileName().toString());
+            analysisResultFile.setMediaType("application");
```
wivern, line 122: why `application`?

pom.xml
```diff
@@ -150,7 +150,7 @@
         <executionengine.resultExclusions></executionengine.resultExclusions>
 		<executionengine.updateStatusCallback>http://localhost:8080/WebAPI/executionservice/callbacks/submission/{id}/status/update/{password}</executionengine.updateStatusCallback>
 		<executionengine.resultCallback>http://localhost:8080/WebAPI/executionservice/callbacks/submission/{id}/result/{password}</executionengine.resultCallback>
-
+    <analysis.result.zipChunkSizeMb>700</analysis.result.zipChunkSizeMb>
```
wivern, line 153: Let's decrease volume size to 100Mb, since this would provide to use small size containers or AWS instances.
And I guess 700Mb is a highly rare case.

YaroslavTir: done

PR #1519, opened 2020-04-22
Issue 1308 record count cache optimization

### LINKED ISSUE #1308: Record count caching optimization
In speaking with @leeevans, we've noted that it takes a while for WebAPI to completely start up since there are jobs that run to cache record counts through this query: https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cdmresults/sql/loadConceptRecordCountCache.sql. We'd like to see if this is something that we could potentially do as part of running Achilles and have it store these results in a stand-alone table so that the ConceptRecordCountCache.sql could be a simple SELECT from that table which would help improve startup time.

ssuvorov-fls:
fixes #1308


src/main/resources/resources/cdmresults/sql/loadConceptRecordCountCache.sql
```diff
@@ -1,79 +1,5 @@
...
-FROM concepts
-  LEFT JOIN counts c1 ON concepts.ancestor_id = c1.concept_id
-  LEFT JOIN counts c2 ON concepts.descendant_id = c2.concept_id
-GROUP BY concepts.ancestor_id
-;
\ No newline at end of file
+  concept_id,
+  record_count,
+  descendant_record_count
+FROM @resultTableQualifier.achilles_result_concept_count;
```
chrisknoll, line 5: Sorry, not familiar with this table:  does this depend on a new version of Achilles or is this something that needs to be constructed as an out-of-band step?

YaroslavTir:
> @ssuvorov-fls - I'm thinking that it would be useful to keep the original query in the code base and have the record count cache use the following logic: issue the query to `achilles_result_concept_count` and if that returns, we're done. If not, issue the original record count query. Otherwise this becomes a breaking change. Let me know if you have any concerns with this approach.

@anthonysena  This PR looks fine from my point of view, I approved it. Can you also take a look (since I was not involved from the begin, and could miss something)?

src/main/java/org/ohdsi/webapi/cdmresults/CDMResultsCacheTasklet.java
```diff
@@ -57,25 +57,33 @@ public RepeatStatus execute(final StepContribution contribution, final ChunkCont
...
+            }
+        }
+    }
 
+    private void warmCdmCache(CDMResultsCache cdmResultsCache, String sql) {
         String resultTableQualifier = source.getTableQualifier(SourceDaimon.DaimonType.Results);
         String vocabularyTableQualifier = source.getTableQualifier(SourceDaimon.DaimonType.Vocabulary);
 
-        String sql_statement = ResourceHelper.GetResourceAsString("/resources/cdmresults/sql/loadConceptRecordCountCache.sql");
+        String sql_statement = ResourceHelper.GetResourceAsString(sql);
```
chrisknoll, line 77: So, this looks a little odd. Usually GetResourceAsString is a path to a resource file, but the sqlParam is that path?

ssuvorov-fls: @chrisknoll GetResourceAsString returns the content of the resource, not the path to it

chrisknoll: Right, so the param 'sql' makes it sound like it is sql, not a path:
```
String sql_statement = ResourceHelper.GetResourceAsString(sql);
```

That was the confusion.

ssuvorov-fls: You are right. Fixed

Issue #1471, opened 2020-03-02
Cohort Generation Cache:  ERROR: duplicate key value violates unique constraint "uq_gc_hash"


chrisknoll:
The following error is reported when failing to generate a cohort.  I'm unable to share the specific cohort details that results in this error; there seems to be a case where a duplicate generation hash is calculated and stored.  

Logs:

```
2020-02-29 04:50:31.892 ERROR taskExecutor-2 org.hibernate.engine.jdbc.spi.SqlExceptionHelper -  - ERROR: duplicate key value violates unique constraint "uq_gc_hash"
  Detail: Key (type, design_hash, source_id)=(COHORT, 1431287467, 235) already exists.
2020-02-29 04:50:31.892 ERROR taskExecutor-2 org.ohdsi.webapi.cohortdefinition.GenerateCohortTasklet -  - org.hibernate.exception.ConstraintViolationException: could not execute statement
...
...
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:213)
	at org.ohdsi.webapi.generationcache.$Proxy238.saveAndFlush(Unknown Source)
	at org.ohdsi.webapi.generationcache.GenerationCacheServiceImpl.cacheResults(GenerationCacheServiceImpl.java:79)
	at org.ohdsi.webapi.generationcache.GenerationCacheHelper.computeCacheIfAbsent(GenerationCacheHelper.java:60)
	at org.ohdsi.webapi.cohortdefinition.GenerateCohortTasklet.prepareQueries(GenerateCohortTasklet.java:82)
	at org.ohdsi.webapi.common.generation.CancelableTasklet.doTask(CancelableTasklet.java:33)

...
...
```

It appears that when the cache is persisted in GenerationCacheServiceImpl.cacheResults():

```
        GenerationCache generationCache = new GenerationCache();
        generationCache.setType(type);
        generationCache.setDesignHash(designHash);
        generationCache.setSource(source);
        generationCache.setResultChecksum(checksum);
        generationCache.setCreatedDate(new Date());

        generationCache = generationCacheRepository.saveAndFlush(generationCache);

        log.info(CACHE_CREATED, type, designHash);
```

[...]

PR #1522, opened 2020-04-23
Placed 'synchronized' earlier

aklochkova:
Fixes https://github.com/OHDSI/WebAPI/issues/1471

The problem was caused by too late synchronization. It was made inside the 'transactional' block and one thread didn't see changes in another after `saveAndFlush` method because the transaction wasn't completed yet.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
+++ b/src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
@@ -47,30 +47,32 @@ public class GenerationCacheHelper {
         log.info("Computes cache if absent for type = {}, design = {}, source id = {}", type, designHash.toString(), source.getSourceId());
 
         synchronized (monitors.computeIfAbsent(new CacheableResource(type, designHash, source.getSourceId()), cr -> new Object())) {
-            log.info("Retrieves or invalidates cache for cohort id = {}", cohortDefinition.getId());
-            GenerationCache cache = generationCacheService.getCacheOrEraseInvalid(type, designHash, source.getSourceId());
-            if (cache == null) {
-                log.info("Cache is absent for cohort id = {}. Calculating with design hash = {}", cohortDefinition.getId(), designHash);
-                // Ensure that there are no records in results schema with which we could mess up
-                generationCacheService.removeCache(type, source, designHash);
-                CohortGenerationRequest cohortGenerationRequest = requestBuilder
-                    .withExpression(cohortDefinition.getDetails().getExpressionObject())
-                    .withSource(source)
-                    .withTargetId(designHash)
-                    .build();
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
```diff
@@ -36,39 +36,37 @@ public GenerationCacheHelper(GenerationCacheService generationCacheService) {
     public Integer computeHash(String expression) {
         return generationCacheService.getDesignHash(CacheableGenerationType.COHORT, expression);
     }
+    
+    //should be called from synchronized block
     public CacheResult computeCacheIfAbsent(CohortDefinition cohortDefinition, Source source, CohortGenerationRequestBuilder requestBuilder, BiConsumer<Integer, String[]> sqlExecutor) {
```
YaroslavTir, line 41: I would keep the monitor in this method,  but run synchronized part in a separate transaction.  Cohort generation sqls anyway contains DDLs, and any ddl commits a transaction, so I don't see any harm in such approach.
@wivern what do you think?

src/main/java/org/ohdsi/webapi/generationcache/GenerationCacheHelper.java
```diff
@@ -57,7 +60,7 @@ public CacheResult computeCacheIfAbsent(CohortDefinition cohortDefinition, Sourc
                     .build();
                 String[] sqls = CohortGenerationUtils.buildGenerationSql(cohortGenerationRequest);
                 sqlExecutor.accept(designHash, sqls);
-                cache = generationCacheService.cacheResults(CacheableGenerationType.COHORT, designHash, source.getSourceId());
+                cache = transactionTemplateRequiresNew.execute(s -> generationCacheService.cacheResults(CacheableGenerationType.COHORT, designHash, source.getSourceId()));
             } else {
```
YaroslavTir, line 64: I think it is better to execute everything in the sync block under new transaction:
```

        synchronized (monitors.computeIfAbsent(new CacheableResource(type, designHash, source.getSourceId()), cr -> new Object())) {
            // we execute the synchronized block in a separate transaction, so the changes become immediately visible to other threads
            return transactionTemplate.execute(s -> {
                ...
            });
        }
    }
```
`transactionTemplate` is better fit here.

chrisknoll: This will only block on the thread that grabbed the object from `monitors`  to synchronize on, correct?  I want to ensure that we don't block all threads from generating, just the threads that happen to want to generate the same design....

YaroslavTir: @chandryou yes this is  correct, only same design thread is going to be blocked

PR #1526, opened 2020-04-27
Authz google

wivern:
Closes OHDSI/ArachneNodeAPI#400

### DIFF
--- a/src/main/java/org/ohdsi/webapi/shiro/filters/GoogleAccessTokenFilter.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/filters/GoogleAccessTokenFilter.java
@@ -1,0 +1,70 @@
+package org.ohdsi.webapi.shiro.filters;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.shiro.authc.AuthenticationException;
+import org.apache.shiro.authc.AuthenticationToken;
+import org.ohdsi.webapi.shiro.TokenManager;
+import org.ohdsi.webapi.shiro.tokens.JwtAuthToken;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.RestTemplate;
+
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import java.io.IOException;
+import java.util.Objects;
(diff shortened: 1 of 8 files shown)


src/main/java/org/ohdsi/webapi/shiro/management/AtlasRegularSecurity.java
```diff
@@ -302,7 +273,7 @@ protected FilterChainBuilder getFilterChainBuilder() {
                 .setBeforeOAuthFilters(SSL, CORS, FORCE_SESSION_CREATION)
                 .setAfterOAuthFilters(UPDATE_TOKEN, SEND_TOKEN_IN_URL)
                 .setRestFilters(SSL, NO_SESSION_CREATION, CORS, NO_CACHE)
-                .setAuthcFilter(JWT_AUTHC)
+                .setAuthcFilter(ACCESS_AUTHC, JWT_AUTHC)
```
YaroslavTir, line 276: Maybe I am wrong, but it looks like that we always sent a request to the google in GoogleAccessTokenFilter even this is JWT. It this is a case, I would swap filters(and another reason is  that JWT are used more frequently)

src/main/java/org/ohdsi/webapi/shiro/filters/GoogleAccessTokenFilter.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Override
+    protected boolean onAccessDenied(ServletRequest servletRequest, ServletResponse servletResponse) throws Exception {
+
+        try {
+            if (TokenManager.extractToken(servletRequest) != null) {
+                boolean loggedIn = executeLogin(servletRequest, servletResponse);
+                if (loggedIn) {
+                    final PrincipalCollection principals = SecurityUtils.getSubject().getPrincipals();
```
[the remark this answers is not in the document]
wivern: How you image this when code written by different people across a long time period?
Ok, let's reduce the `final` out

src/main/java/org/ohdsi/webapi/shiro/management/AtlasRegularSecurity.java
wivern: I don't feel the swap of filters is good idea, since this requires modification of AtlasJwtFilter behavior.
I'd like to suggest add an option to turn on or off the Access Token filter due to possible security restrictions.

PR #1528, opened 2020-05-01
Cohort Pathway Duplication Issue Fix

chrisknoll:
Fixes for Pathways analysis.
Added embedded PostgreSQL instance tests for Pathways.
Some refactoring on existing test cases.

### DIFF
--- a/src/test/java/org/ohdsi/webapi/pathway/PathwayAnalysisTest.java
+++ b/src/test/java/org/ohdsi/webapi/pathway/PathwayAnalysisTest.java
@@ -172,6 +172,34 @@ public class PathwayAnalysisTest extends AbstractDatabaseTest {
 
     return source;
   }
+  
+  private void loadPrepData(String[] datasetPaths) throws Exception {
+    final IDatabaseConnection dbUnitCon = getConnection();
+    final IDataSet ds = DataSetFactory.createDataSet(datasetPaths);
+
+    assertNotNull("No dataset found", ds);
+
+    try {
+      DatabaseOperation.CLEAN_INSERT.execute(dbUnitCon, ds); // clean load of the DB. Careful, clean means "delete the old stuff"
+    } catch (final DatabaseUnitException e) {
+      fail(e.getMessage());
+    } finally {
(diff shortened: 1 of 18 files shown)


src/test/java/org/ohdsi/webapi/pathway/PathwayAnalysisTest.java
```diff
@@ -0,0 +1,333 @@
...
+
+    assertNotNull("No dataset found", ds);
+
+    try {
+      DatabaseOperation.CLEAN_INSERT.execute(dbUnitCon, ds); // clean load of the DB. Careful, clean means "delete the old stuff"
+    } catch (final DatabaseUnitException e) {
+      fail(e.getMessage());
+    } finally {
+      dbUnitCon.close();
+    }
```
azimov, line 294: This block is identical to the block in the unit above. It would be good to see a method that takes `testDataSetsPaths` and `dbUnitCon` and then handles the rest of this.

src/test/java/org/ohdsi/webapi/pathway/PathwayAnalysisTest.java
```diff
@@ -0,0 +1,333 @@
...
+    JobExecutionResource executionResource = pathwayService.generatePathways(entity.getId(), SOURCE_ID);
+    PathwayAnalysisGenerationEntity generationEntity;
+    while (true) {
+        generationEntity = pathwayService.getGeneration(executionResource.getExecutionId());
+        if (generationEntity.getStatus().equals("FAILED") || generationEntity.getStatus().equals("COMPLETED")) {
+            break;
+        }
+        Thread.sleep(2000L);
+    }
+    assertEquals("COMPLETED", generationEntity.getStatus());
```
azimov, line 311: As with the comment above, this code block is identical to the unit above. Is it necessary to be explicit here? Maybe a method that takes the `PathwayAnalysisEntity` instance and the path to the verification json file would be cleaner.

chrisknoll:
Thank you for the review. I've addressed your comments. Let me know if you find any other issues.

PR #1533, opened 2020-05-12
Fix for CEM sibilings

anthonysena:
Tested for 2.8 per #1485 and would like to patch this for 2.7.7.

Thanks @ericaVoss for this fix!


src/main/resources/resources/evidence/sql/negativecontrols/optimizeEvidence.sql
```diff
@@ -56,30 +62,88 @@ conceptSetAnalysis AS (
...
   FROM CTE_NOT_PREVELANT_CONCEPTS
-) z
-;
+) z;
+
+/* ===STEP 2=== */
+/*ELIMINATE SIBLINGS*/
+WITH CTE_LIST AS (
+    /*Using the rows that were okay from above, let's further filter siblings*/
+  	SELECT *, OUTCOME_OF_INTEREST_CONCEPT_ID AS CONCEPT_ID, OUTCOME_OF_INTEREST_CONCEPT_NAME AS CONCEPT_NAME
```
YaroslavTir, line 94: I personally don't like asterisks in the quotes, they may be causing bugs in the future. So if it possible replace it with columns names

anthonysena: Agreed that explicit column names would be better here. For now though, I'm trying to keep this code consistent with the way it is done in the CEM code base where Erica maintains this. So, could we live with this as-is for now? Thanks.

YaroslavTir: I don't mind

Issue #1511, opened 2020-04-16
Google OAuth couldn't be used for new projects due to legacy API still in use


wivern:
Google OAuth authentication in the WebAPI uses Legacy People API to retreive info about person who has logged in.
But this API is [deprecated](https://developers.google.com/people/legacy) since March 7, 2019 and cannot be used in new projects. This means none is able to deploy new WebAPI istance with Google OAuth authentication.
Google recommends migrate to the new People API.

anthonysena:
Looks like this has been fixed for 2.7.7 - can you make a PR for 2.8.0? Also, it appears that this closed out #1502 as well?

chrisknoll:
Does this update require JDK 11?

anthonysena:
@chrisknoll - just checked this on my side and for me the 2.7.7 build used opensaml-saml-api 3.2.0 which is JDK8 compatible.

chrisknoll:
odd...I pulled the commits from the 2.7.7 branch which led to that error I reported, so...I don't know how that happened.

PR #1534, opened 2020-05-12
Rollback logout option #1457

anthonysena:
Rolls back changes from #1457. 

Tagging @phozzy - apologies but I need to roll this back since we're using an older version of pac4j that does not support the logout option (http://www.pac4j.org/apidocs/pac4j/1.9.2/org/pac4j/oidc/config/OidcConfiguration.html).

### DIFF
--- a/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
+++ b/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
@@ -36,9 +36,6 @@ public class OidcConfCreator {
     @Value("${security.oid.url}")
     private String url;
 
-    @Value("${security.oid.logoutUrl}")
-    private String logoutUrl;
-
     @Value("${security.oauth.callback.api}")
     private String oauthApiCallback;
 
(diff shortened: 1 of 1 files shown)


anthonysena:
Linking this to #1511 - when we back merge that work into master, we should apply the work that was done in #1457 since that will update the pac4j components required.

phozzy:
oh sorry @anthonysena . that is sad... the reason for my PR was that it works for TheHyve ¯\_(ツ)_/¯ but may be we use more recent pac4j.

anthonysena:
Same here @phozzy - talking with @chrisknoll on this topic, we're going to take the work done in this PR: https://github.com/OHDSI/WebAPI/pull/1526/files and bring that into master which will upgrade the pac4j dependency. Then we can re-apply your contribution with that set of changes :-)

phozzy:
> Same here @phozzy - talking with @chrisknoll on this topic, we're going to take the work done in this PR: https://github.com/OHDSI/WebAPI/pull/1526/files and bring that into master which will upgrade the pac4j dependency. Then we can re-apply your contribution with that set of changes :-)

Great! Thanks!

PR #1537, opened 2020-05-13
pac4j and Authz google id from 2.7

chrisknoll:
This is an attempt to bring in pac4j changes from 2.7 into master.

There were several API changes and dependencies on JDK 11 that were needed to be resolved for use in 2.8.

I am unable to confirm oAuth or SAML authentications, so I will depend on the origional authors to verify these commits look correct.

Important note:  when adding the changes for the OidcConfiguration object, i noticed that there is no longer a `conf.setCallbackUrl(oauthApiCallback);` available on the configuration object.  I'm not sure if that needs to be specified elsewhere, but it leaves the application.properties value 'security.oauth.callback.api' unused.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -699,7 +699,7 @@
     </dependency>
     <dependency>
       <groupId>org.pac4j</groupId>
-      <artifactId>pac4j-saml</artifactId>
+      <artifactId>pac4j-saml-opensamlv3</artifactId>
       <version>${pac4j.version}</version>
     </dependency>
     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,8 @@
     <security.ssl.enabled>false</security.ssl.enabled>
     <security.oauth.callback.ui>http://localhost/Atlas/#/welcome</security.oauth.callback.ui>
     <security.oauth.callback.api>http://localhost:8080/WebAPI/user/oauth/callback</security.oauth.callback.api>
+    <!-- Available options for callback urlResolver are: query and path -->
+    <security.oauth.callback.urlResolver>query</security.oauth.callback.urlResolver>
     <security.oauth.google.apiKey></security.oauth.google.apiKey>
     <security.oauth.google.apiSecret></security.oauth.google.apiSecret>
     <security.oauth.facebook.apiKey></security.oauth.facebook.apiKey>
--- a/pom.xml
+++ b/pom.xml
@@ -128,6 +128,7 @@
 
     <security.googleIap.cloudProjectId></security.googleIap.cloudProjectId>
     <security.googleIap.backendServiceId></security.googleIap.backendServiceId>
+    <security.google.accessToken.enabled>false</security.google.accessToken.enabled>
 
     <security.cors.enabled>true</security.cors.enabled>
     <security.maxLoginAttempts>3</security.maxLoginAttempts>
--- a/pom.xml
+++ b/pom.xml
@@ -15,12 +15,12 @@
     <spring.boot.version>1.5.20.RELEASE</spring.boot.version>
     <flyway.version>4.2.0</flyway.version>
     <waffle.version>1.7.3</waffle.version>
+
     <circe.version>1.9.0-SNAPSHOT</circe.version>
     <jersey.version>2.14</jersey.version>
     <SqlRender.version>1.6.3-SNAPSHOT</SqlRender.version>
     <hive-jdbc.version>3.1.2</hive-jdbc.version>
-    <pac4j.version>1.9.2</pac4j.version>
-
+    <pac4j.version>4.0.0</pac4j.version>
     <start-class>org.ohdsi.webapi.WebApi</start-class>
     <skipTests>true</skipTests>
 
@@ -684,7 +684,7 @@
     <dependency>
       <groupId>io.buji</groupId>
       <artifactId>buji-pac4j</artifactId>
-      <version>2.0.2</version>
+      <version>5.0.0</version>
     </dependency>
     <dependency>
       <groupId>org.pac4j</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
     <security.oid.clientId></security.oid.clientId>
     <security.oid.apiSecret></security.oid.apiSecret>
     <security.oid.url></security.oid.url>
+    <security.oid.logoutUrl></security.oid.logoutUrl>
     <security.oid.redirectUrl>http://localhost/index.html#/welcome/</security.oid.redirectUrl>
     <security.kerberos.spn></security.kerberos.spn>
     <security.kerberos.keytabPath></security.kerberos.keytabPath>
--- a/src/main/java/org/ohdsi/webapi/Constants.java
+++ b/src/main/java/org/ohdsi/webapi/Constants.java
@@ -88,4 +88,9 @@ public interface Constants {
     String COHORT_SUMMARY_STATS_CACHE = "cohort_summary_stats_cache";
     String COHORT_CENSOR_STATS_CACHE = "cohort_censor_stats_cache";
   }
+
+  interface CallbackUrlResolvers {
+    String QUERY = "query";
+    String PATH = "path";
+  }
 }
--- a/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
+++ b/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
@@ -44,7 +44,6 @@ public class OidcConfCreator {
         conf.setClientId(clientId);
         conf.setSecret(apiSecret);
         conf.setDiscoveryURI(url);
-        conf.setCallbackUrl(oauthApiCallback);
         conf.setPreferredJwsAlgorithm(JWSAlgorithm.RS256);
         return conf;
     }
--- a/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
+++ b/src/main/java/org/ohdsi/webapi/OidcConfCreator.java
@@ -35,7 +35,10 @@ public class OidcConfCreator {
 
     @Value("${security.oid.url}")
     private String url;
-
+    
+    @Value("${security.oid.logoutUrl}")
+    private String logoutUrl;
+    
     @Value("${security.oauth.callback.api}")
     private String oauthApiCallback;
 
@@ -44,6 +47,7 @@ public class OidcConfCreator {
         conf.setClientId(clientId);
         conf.setSecret(apiSecret);
         conf.setDiscoveryURI(url);
+        conf.setLogoutUrl(logoutUrl);
         conf.setPreferredJwsAlgorithm(JWSAlgorithm.RS256);
         return conf;
     }
--- a/src/main/java/org/ohdsi/webapi/security/SSOController.java
+++ b/src/main/java/org/ohdsi/webapi/security/SSOController.java
@@ -68,7 +68,7 @@ public class SSOController {
     @GET
     @Path("/slo")
     public Response logout() throws URISyntaxException {
-        return Response.status(HttpConstants.TEMP_REDIRECT)
+        return Response.status(HttpConstants.TEMPORARY_REDIRECT)
                 .header(HttpConstants.ACCESS_CONTROL_ALLOW_ORIGIN_HEADER, this.origin)
                 .location(new URI(sloUri))
                 .build();
--- a/src/main/java/org/ohdsi/webapi/shiro/filters/CasHandleFilter.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/filters/CasHandleFilter.java
@@ -18,6 +18,7 @@ import org.slf4j.LoggerFactory;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpSession;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 
 /**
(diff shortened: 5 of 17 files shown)


chrisknoll:
@phozzy : is this soemthing you can test on your end?

anthonysena:
@chrisknoll thanks for making these updates! The point you raised around `conf.setCallbackUrl(oauthApiCallback)` earlier is a good one and something we can test in our environment since I've configured an authorization client that uses OpenID in one of our non-production environments. I'll try that out and let you know the results.

anthonysena:
I think there is a problem - when attempting to test with a configuration for OpenID that was working previously, I'm now getting the following error from the OpenID provider:

````
Invalid redirect: https://{server}/WebAPI/user/oauth/callback?client_name=OidcClient/OidcClient does not match one of the registered values: [https://{server}/WebAPI/user/oauth/callback?client_name=OidcClient]
````

I need to dig into the details a bit more but wanted to share this in case it is helpful. Tagging @wivern for thoughts.

**Comment by chrisknoll** [comment]
I'm no expert on OID, but

```
Invalid redirect: https://{server}/WebAPI/user/oauth/callback?client_name=OidcClient/OidcClient does not match one of the registered values: [https://{server}/WebAPI/user/oauth/callback?client_name=OidcClient]
```
I am wondering where the mismatch between client_name = OidcClient and OidcClient/OidcClient comes from.  is it related to the removal of `conf.setCallbackUrl(oauthApiCallback)`.

**Comment by anthonysena** [comment]
Potentially - I do see that we still use the `oauthApiCallback` when we initialize the client: https://github.com/OHDSI/WebAPI/pull/1537/files#diff-c81452c048d6a1d9ca058c6a94776ec4R277. I just updated my comment above to mention that another part of the error mentioned "Error invalid_grant". 

FWIW, the configuration I'm using in the settings.xml when building the WebAPI.war is: 

<security.oauth.callback.api>https://{server}/WebAPI/user/oauth/callback?client_name=OidcClient</security.oauth.callback.api>

Maybe this needs an adjustment to work with the callback endpoint now that we're using a newer version of pac4j?

**Comment by chrisknoll** [comment]
Ah, you're correct, it's passed into the constructor.  I found this article on the callback endpoint, and they speak specifically on the callback endpiont, and they discuss changes between V2 and V3.
https://www.pac4j.org/blog/understanding-the-callback-endpoint.html

I'm pretty sure the callback URL specified should not specify the client name.  Taht seems to be dynamic based on the specific oauth client (facebook, google, etc).

**Comment by phozzy** [comment]
> @phozzy : is this soemthing you can test on your end?

We use only OpenID Connect. So I can not test wit SAML.

**Comment by anthonysena** [comment]
@chrisknoll thanks for the information above related to the callback URL. I made some adjustments and think I'm getting closer but could use some input from @wivern since he mentioned that he was able to make this work for oAuth.

I'm using the OpenID Connect (OidcClient) to test this out. In my settings.xml, I'm now using:

`<security.oauth.callback.api>https://<server>/WebAPI/user/oauth/callback</security.oauth.callback.api>`

Then, on the OpenID side, I've set the redirect URL to be:

`https://<server>/WebAPI/user/oauth/callback/OidcClient`

This was necessary since, in theory, once the OpenID provider authenticates the user, they are pushed back to redirect URL. In my case, the resulting URL to WebAPI looks like this:

`https://awsagunva3088.jnj.com:8443/WebAPI/user/oauth/callback/OidcClient?code=xHzhYCpH5efWhRxbktgVlC&state=oq-sACpfBvQI1sSVpcemufnLut7scM74XqvPH5zew4Q`

This is resulting in a `401 - NOT FOUND` error. 

I fully expect this is just me being ignorant of how to configure the client vs. a problem in the code. That said, I'd like to make sure this is in working order before we merge so that I can also document any changes that may be required here.

**Comment by chrisknoll** [comment]
So, I feel like the 401 is because the path to /user/oauth/callback is defined, but when the `/OidcClient` is appended, it results in a path that isn't mapped to webapi, and you get a 401.  my question about this is: I see that these paths are defined:

```
                .addOAuthPath("/user/oauth/google", GOOGLE_AUTHC)
                .addOAuthPath("/user/oauth/facebook", FACEBOOK_AUTHC)
                .addOAuthPath("/user/oauth/github", GITHUB_AUTHC)
                .addPath("/user/login/cas", SSL, CORS, FORCE_SESSION_CREATION, CAS_AUTHC, UPDATE_TOKEN, SEND_TOKEN_IN_URL)
                .addPath("/user/oauth/callback", SSL, HANDLE_UNSUCCESSFUL_OAUTH, OAUTH_CALLBACK)
```
I wonder why the /user/oauth/callback is added via `.addPath` and the others (google, facebook, github) are added via .addOAuthPath().   I'm also very ignorant as to how this is implemented (we didn't write it), so what we really need right now is either @wivern , @ssuvorov-fls or others from @OHDSI/odysseus  to chime in on what's happening.  And I hope they can do this soon since 1) this has been open for over 2 weeks and 2) this is just pulling stuff that was working in 2.7x branch so this should be a very straightforward activity.

**Comment by wivern** [comment]
@chrisknoll You're completely right, I changed URL resolver to the `QueryParameterCallbackUrlResolver` to avoid mentioned issue.
Regarding `addOauthPath` and `addPath`, to use OAuth providers such as Google or GitHub we should provide them some callback url. It should be used during authentication process, but OAuth providers knows nothing about our service, therefore callbacks should be unsecured, and we added them with `addPath`. But OAuth entrypoints should provide redirection to the related provider, such functionality supported by the `addOuathPath`.

**Comment by chrisknoll** [comment]
Thanks for the information, @wivern.  I really do need to get more familiar with the guts of the 'oAuth dance', and I'll try to set up time with @anthonysena to take me through a deep dive of the custom OAuth provider that I think he's working from.

**Comment by anthonysena** [comment]
@wivern @chrisknoll a quick update here. I deployed this new branch in our environment which is connected to Elixir OpenID. I'm still facing an issue and have contacted Elixir for support since I'm unsure where the problem lies. Here is the relevant configuration I'm using for WebAPI:

````
<security.oid.clientId>omitted</security.oid.clientId>
<security.oid.apiSecret>omitted</security.oid.apiSecret>
<security.oid.url>https://login.elixir-czech.org/oidc/.well-known/openid-configuration</security.oid.url>
<security.oid.redirectUrl>https://<server>/atlas/#/welcome/</security.oid.redirectUrl>
<security.oauth.callback.ui>https://<server/atlas/#/welcome/</security.oauth.callback.ui>
<security.oauth.callback.api>https://<server>:8443/WebAPI/user/oauth/callback</security.oauth.callback.api>
````

Please note above that for `<security.oauth.callback.api>` I am using:  `https://<server>:8443/WebAPI/user/oauth/callback` vs `https://<server>:8443/WebAPI/user/oauth/callback/OidcClient`. On the Elixir side, I've configured this to use `https://<server>:8443/WebAPI/user/oauth/callback/OidcClient`
Please let me know if this all looks correct. On the Elixir side, I'm getting the following error:

````
Error invalid_grant
There was an error processing your request.

Invalid redirect: https://<server>:8443/WebAPI/user/oauth/callback?client_name=OidcClient does not match one of the registered values: [https://<server>:8443/WebAPI/user/oauth/callback/OidcClient]
````

I'm hoping Elixir support will have some insights into why the old URI is still being used for the redirect. Please let me know if you see any problems with the configuration above. Thanks!

**Comment by anthonysena** [comment]
Heard back from support - they think that it could be the login button from Atlas that is causing the issue. Looking at the config on Atlas:

````
	configLocal.authProviders = [
    {
      "name": "Windows",
      "url": "user/login/windows",
      "ajax": true,
      "icon": "fa fa-windows"
    },
    {
      "name": "Elixir AAI",
      "url": "user/login/openid",
      "ajax": false,
      "icon": "fa fa-openid"
    }];
````

The endpoint `user/login/openid` has the following configuration: in [AtlasRegularSecurity.java](https://github.com/OHDSI/WebAPI/blob/d3a2fd1a4aa80512042c9e177f46e19a006a433a/src/main/java/org/ohdsi/webapi/shiro/management/AtlasRegularSecurity.java#L332): 

`.addRestPath("/user/login/openid", FORCE_SESSION_CREATION, OIDC_AUTH, UPDATE_TOKEN, SEND_TOKEN_IN_REDIRECT)` 

Unfortunately I'm hitting my limit with tracing through this code since it appears that each of the constants above reference filters that I don't completely understand. The question I have: could one of these filters still be using the old querystring approach for OpenID?


user/login/openi

**Comment by wivern** [comment]
@anthonysena  Only the `OIDC_AUTH` is involved to OpenID authentication process.
One initializing by the following:
```
       SecurityFilter oidcFilter = new SecurityFilter();
        oidcFilter.setConfig(cfg);
        oidcFilter.setClients("OidcClient");
        filters.put(OIDC_AUTH, oidcFilter);
```
and the `SecurityFilter` is from pac4j, therefore this should not use old-style querystring unless older version of the pac4j still remains in the classpath.

anthonysena:
Thanks @wivern. I'm still facing an issue here but let me describe what I'm seeing:

- I've confirmed that this build is using Pac4J v4 by inspecting the files unpacked by the WebAPI.war:

image

- I traced the web request via ATLAS and found the request header is still using the old querystring style. Here is the output (with some values censored) from Chrome developer tools:

````
General:

Request URL: https://<server>:8443/WebAPI/user/login/openid
Request Method: GET
Status Code: 302 
Remote Address: <server ip>:8443
Referrer Policy: no-referrer-when-downgrade

Response Headers:

Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Date: Tue, 16 Jun 2020 18:33:06 GMT
Expires: 0
Location: https://login.elixir-czech.org/oidc/authorize?response_type=code&redirect_uri=https%3A%2F%2F<server>%3A8443%2FWebAPI%2Fuser%2Foauth%2Fcallback%3Fclient_name%3DOidcClient&state=aFCt4_6ef9yxnucB-kwi3BH7EftTyPoUkxpkerLaBv8&client_id=c1d2aa10-f023-405c-8880-5b44e2fdba17&scope=openid+profile+email
Pragma: no-cache
Set-Cookie: JSESSIONID=25D61A064BB814AFE2B721E16581C839; Path=/WebAPI; Secure; HttpOnly
Set-Cookie: pac4jCsrfToken=e192fb76-4d0f-4a2a-8d9d-41198fdd902c; Domain=<server>; Path=/
Strict-Transport-Security: max-age=15768000 ; includeSubDomains
Transfer-Encoding: chunked
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block

Request Headers

Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Connection: keep-alive
Cookie: _ga=GA1.2.1787178878.1586795835; jenkins-timestamper-offset=14400000; __utmz=28368675.1588258145.1.1.utmcsr=(direct)|utmccn=(direct)|utmcmd=(none); _fbp=fb.1.1589902316600.1496563961; __utma=28368675.1787178878.1586795835.1588258145.1591011227.2; bearerToken=eyJhbGciOiJIUzUxMiJ9.eyJzdWIiOiJuYVxcYXNlbmE1IiwiZXhwIjoxNTkyMjkxODUxfQ.LEW0waQxmv15mOU6g3sYvK89OK595GQQazSoxx1rmTL9hkCFaMdCAKzNMYUfg4Qz8jwiCc7r_CFYTuNeZz3MGg
Host: <server>:8443
Referer: https://<server>/atlas/
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: same-site
Sec-Fetch-User: ?1
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.106 Safari/537.36
````

Note in the "redirect_uri" parameter in the request header's the `Location`: 

https://login.elixir-czech.org/oidc/authorize?response_type=code&redirect_uri=https%3A%2F%2F<server>%3A8443%2FWebAPI%2Fuser%2Foauth%2Fcallback%3Fclient_name%3DOidcClient&state=aFCt4_6ef9yxnucB-kwi3BH7EftTyPoUkxpkerLaBv8&client_id=c1d2aa10-f023-405c-8880-5b44e2fdba17&scope=openid+profile+email


Here is the "redirect_uri" parameter decoded:

https:// <server> :8443/WebAPI/user/oauth/callback?client_name=OidcClient&state=aFCt4_6ef9yxnucB-kwi3BH7EftTyPoUkxpkerLaBv8&client_id=c1d2aa10-f023-405c-8880-5b44e2fdba17&scope=openid profile email


Question: can you reproduce this behavior on your side? We should ensure that both of our teams can use Elixir for authentication for the EHDEN project.

wivern:
Unfortunately we don't have access to the Elixir. We tested this using Google OAuth from our side, but one's not exactly the same.

gklebanov:
Elixir is a federated identity provider. Used in EU in public projects often but offer similar functionality to other similar providers such as WSO2 (open source) or PingFederate (commercial). Since Odysseus is part of EHDEN, I will request @wivern to have access to Elixir. But longer term, we should consider standing up the WSO2 somewhere in OHDSI

wivern:
@anthonysena Deeper learning of pac4j internals causing the following changes:
1. I added configurable option `security.oauth.callback.urlResolver` which has two options:
 - `query` as a default value and with this option webapi would expect old-style callback with client identity as a query parameter
 - `path` would provide new-style callback URL with client identity encoded as a path parameter like `http://localhost/WebAPI/user/oauth/callback/Google2Client`
2. Added additional logger for pac4j that provides turn on the security chain trace separately from the whole app

I've tested this from my side over Google Authentication and WSO2 as an openid provider.

anthonysena:
Thanks @wivern - thank worked! For completeness, here are the settings I used, including the newly added `urlResolver` property:

````
<security.oid.clientId>omitted</security.oid.clientId>
<security.oid.apiSecret>omitted</security.oid.apiSecret>
<security.oid.url>https://login.elixir-czech.org/oidc/.well-known/openid-configuration</security.oid.url>
<security.oid.redirectUrl>https://<server>/atlas/#/welcome/</security.oid.redirectUrl>
<security.oauth.callback.ui>https://<server/atlas/#/welcome/</security.oauth.callback.ui>
<security.oauth.callback.api>https://<server>:8443/WebAPI/user/oauth/callback</security.oauth.callback.api>
<security.oauth.callback.urlResolver>path</security.oauth.callback.urlResolver>
````

PR #1552, opened 2020-06-08
add diamon priority for 2.7.7 for fix vocabulary switching(https://gi…

YaroslavTir:
Fixes https://github.com/OHDSI/Atlas/issues/2223
That fix was extracted from the master branch.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/SourcePriorityService.java
+++ b/src/main/java/org/ohdsi/webapi/service/SourcePriorityService.java
@@ -1,0 +1,83 @@
+package org.ohdsi.webapi.service;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.commons.collections4.map.PassiveExpiringMap;
+import org.ohdsi.sql.SqlTranslate;
+import org.ohdsi.webapi.shiro.management.datasource.SourceAccessor;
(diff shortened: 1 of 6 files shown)


chrisknoll:
Pardons, but, i'm looking in master, and I do not see a 'SourcePriorityService' in the `org.ohdsi/webapi/service` package.  So, I'm not sure what is meant by `extracted from 2.8'?

This is good tho, because I think I would have objected to a dedicated service for 'source priority' when it seems like this should just be a function of the source service.  So I'm not sure why it was split out this way.   So, two mysteries I hope you all can help me with.

YaroslavTir:
> Pardons, but, i'm looking in master, and I do not see a 'SourcePriorityService' in the `org.ohdsi/webapi/service` package. So, I'm not sure what is meant by `extracted from 2.8'?
> 
> This is good tho, because I think I would have objected to a dedicated service for 'source priority' when it seems like this should just be a function of the source service. So I'm not sure why it was split out this way. So, two mysteries I hope you all can help me with.

@chrisknoll sure, I can help you with this. In master, we split service to the service and controller, so technically code is different for two branches, and there is no way just cherrypicked, or copy-paste it. I put the necessary code from master `SourceService` to the `SourcePriorityService` to make back merge easier, basically, `SourcePriorityService` can be through away after merge.

chrisknoll:
Ok, got it.  I found the code in SourceService in master related to the `priorityDaimon` so I think all is well from a master branch perspective. Thanks for clearing that up.

PR #1556, opened 2020-06-10
Add "Shared with me" flag to common entities

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/2167


src/main/java/org/ohdsi/webapi/security/PermissionService.java
```diff
@@ -139,10 +154,65 @@ public String getPermission(String template, Object entityId) {
...
+        if (permissionCache.get().isEmpty()) {
+            preparePermissionCache(permissionTemplates);
+        }
+
+        List<String> permissions = permissionTemplates.stream()
+                .map(pt -> getPermission(pt, id))
+                .collect(Collectors.toList());
+        int fitCount = permissions.size();
+        Map<RoleDTO, Long> roleMap = permissions.stream()
+                .flatMap(p -> permissionCache.get().get(p).stream())
```
anton-abushkevich, line 206: NPE here. 
Try opening any characterization with cohorts filled. Apart from characterization, there will be multiple calls of this method checking permissions for all cohorts in characterization. Trying to check permissions for cohorts while not updating the permissionCache (`if (permissionCache.get().isEmpty())` is false) results in NPE. 
Maybe can do without ThreadLocal cache altogether?

anton-abushkevich (approved):

PR #1567, opened 2020-06-26
Issue 1561 web api failed to start

### LINKED ISSUE #1561: WebAPI failed to start in some cases with enabled AD synchronization
Steps to reproduce:

1. Create new data source
2. Setup AD Group synchronization with role of created DS.
3. Delete DS
4. Restart WebAPI and observe the error.

Problem: AD configuration is not cleaned up with deleted role of datasource and it causes failure during WebAPI start.

Please check the same during Sync process. What will happen with removed roles?

anton-abushkevich:
Fixes #1561


anton-abushkevich:
Creation of DeleteRoleEvent moved from UserService to PermissionManager, because UserService doesn't participate in our call chain: SourceService -> AtlasSecurity -> PermissionManager. Yet we need this Event to remove groups.

PR #1573, opened 2020-07-08
Impoves Achilles reports caching

### LINKED ISSUE #1551: ACHILLES reports and caching
Every time when opening Achilles reports it takes minutes to load it for the first time.
It should already be cached and cache shared across users and sessions.

wivern:
Closes #1551


src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -318,10 +364,28 @@ private String getWarmCacheJobName(String sourceKey) {
...
+        Step resultsStep = stepBuilderFactory.get(jobName + " results").tasklet(resultsTasklet).build();
+
+        SimpleJobBuilder builder = jobBuilders.get(jobName)
+                .start(dashboardStep)
+                .next(personStep)
+                .next(dataDensityStep)
+                .next(resultsStep);
+        JobParametersBuilder jobParametersBuilder = new JobParametersBuilder();
+        jobParametersBuilder.addString(Constants.Params.JOB_NAME, jobName);
+        return jobService.runJob(builder.build(), jobParametersBuilder.toJobParameters());
```
YaroslavTir, line 389: That is a matter of taste, but I think that builder like formatting is more clear:
```
        Step dashboardStep = stepBuilderFactory
                .get(jobName + " dashboard")
                .tasklet(new DashboardCacheTasklet(source, instance, cacheManager))
                .build();
        Step personStep = stepBuilderFactory
                .get(jobName + " person")
                .tasklet(new PersonCacheTasklet(source, instance, cacheManager))
                .build();
        Step dataDensityStep = stepBuilderFactory
                .get(jobName + " data density")
                .tasklet(new DataDensityCacheTasklet(source, instance, cacheManager))
                .build();
        Step resultsStep = stepBuilderFactory
                .get(jobName + " results")
                .tasklet(new CDMResultsCacheTasklet(this.getSourceJdbcTemplate(source), getTransactionTemplateRequiresNew(), source))
                .build();
        SimpleJobBuilder builder = jobBuilders
                .get(jobName)
                .start(dashboardStep)
                .next(personStep)
                .next(dataDensityStep)
                .next(resultsStep);
        JobParameters jobParameters = new JobParametersBuilder()
                .addString(Constants.Params.JOB_NAME, jobName)
                .toJobParameters();
```

src/main/java/org/ohdsi/webapi/cdmresults/eviction/CDMEvictionAdvisor.java
```diff
@@ -0,0 +1,7 @@
+package org.ohdsi.webapi.cdmresults.eviction;
+
+import org.ehcache.config.EvictionAdvisor;
+
+public interface CDMEvictionAdvisor<K, V> extends EvictionAdvisor<K, V> {
```
YaroslavTir, line 5: I know that is not a trivial thing to explain why we need that interface, but without this other developers will spend a significant amount of time figuring it out. I personally did not get at first why do we need the Advisors in the first place. So I would recommend adding jdoc with an explanation like this 
1) We need to refresh the cache seamless for users,  the evict + cache data is not a solution for us, so we do it in the CacheService 
2) The easiest way to attach get data for cache to the concrete type of  cache is to use advisors

src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -180,24 +200,29 @@ public CDMDashboard getDashboard(@PathParam("sourceKey")
...
     public CDMPersonSummary getPerson(@PathParam("sourceKey")
             final String sourceKey, @DefaultValue("false")
             @QueryParam("refresh") boolean refresh) {
-        Source source = getSourceRepository().findBySourceKey(sourceKey);
-        CDMPersonSummary person = this.queryRunner.getPersonResults(this.getSourceJdbcTemplate(source), source);
+        CDMPersonSummary person = getRawPerson(sourceKey, refresh);
         return person;
     }
 
+    public CDMPersonSummary getRawPerson(String sourceKey, boolean refresh) {
```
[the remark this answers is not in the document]
wivern: I know, but couldn't break the existing external contract

src/main/java/org/ohdsi/webapi/cdmresults/eviction/CDMEvictionAdvisor.java
wivern: I added some comment to the appropriate method of the CacheService.
As to me pointed interface isn't suitable place for a suggested explanation.

PR #1576, opened 2020-07-08
Cast combo_id to bigint.

### LINKED ISSUE #1575: Error generating pathway anlaysis with > 32 event cohorts
There is an issue in casting the combo_id to INT when calculating the `#non_repetitive_events`

```
FROM (
  SELECT
    combo_id, subject_id, cohort_start_date, cohort_end_date,
    CASE WHEN (combo_id = LAG(combo_id) OVER (PARTITION BY subject_id ORDER BY subject_id, cohort_start_date ASC))
      THEN 1
      ELSE 0
    END repetitive_event, 
		case when ROW_NUMBER() OVER (PARTITION BY subject_id, CAST(combo_id AS INT) ORDER BY cohort_start_date) > 1 then 1 else 0 end is_repeat
  FROM #combo_events
) AS marked_repetitive_events
```

The issue is casting the combo_id to INT, when combo_id is actually a BIGINT.

chrisknoll:
Fixes #1575

### DIFF
--- a/src/main/resources/resources/pathway/runPathwayAnalysis.sql
+++ b/src/main/resources/resources/pathway/runPathwayAnalysis.sql
@@ -203,7 +203,7 @@ FROM (
       THEN 1
       ELSE 0
     END repetitive_event, 
-		case when ROW_NUMBER() OVER (PARTITION BY subject_id, CAST(combo_id AS INT) ORDER BY cohort_start_date) > 1 then 1 else 0 end is_repeat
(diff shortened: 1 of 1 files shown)


chrisknoll:
Btw: while this cast does address the column type defined in the DB, there's still INT types used in the java layer, so there will be an additional future task to support BIGINT across the board. For now (even for 2.8) we can live with the INT limitation in pathway analysis (which limits to 32 event cohorts).

PR #1580, opened 2020-07-08
Add event cohorts combo_ids to the result.

### LINKED ISSUE #1577: Pathway  buildPathwayCodes() misses event Cohorts
In `buildPathwayCodes`, we use the pathway result data to generate the lookukp of code-to-cohortName mapping, which is then used in the UI to resolve the code to the name.

The issue is that the individual event cohorts are not always appearing in the pathway codes because it's possible that the individual event cohorts are only found in the context of combos.  However, this leads to a problem on the front end when we try to find the individual event cohort components of a combo.

We need to put the individual event cohorts (and their associated combo_id) into the database when we `savePathwayCodes()`.

chrisknoll:
Fixes #1577.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
+++ b/src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
@@ -33,139 +33,141 @@ import org.springframework.jdbc.core.PreparedStatementCreator;
 
 import static org.ohdsi.webapi.Constants.Params.GENERATION_ID;
 
-
 public class PathwayStatisticsTasklet extends CancelableTasklet {
-    
-    private final CancelableJdbcTemplate jdbcTemplate;
-    private final Source source;
-    private Long generationId;
-    private final PathwayService pathwayService;
-    private final GenericConversionService genericConversionService;
-    
-    public PathwayStatisticsTasklet(CancelableJdbcTemplate jdbcTemplate, TransactionTemplate transactionTemplate, Source source,
-						PathwayService pathwayService, GenericConversionService genericConversionService) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
```diff
@@ -103,20 +105,38 @@ private void initTx() {
 							new Object[] { generationId }
 			);
 
-			return jdbcTemplate.query(pathwayStatsPsr.getSql(), pathwayStatsPsr.getSetter(), (rs, rowNum) -> {
+      Map<Integer, Integer> eventCodes = pathwayService.getEventCohortCodes(design);
```
azimov, line 108: Code indentation in this method is a bit odd

src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
```diff
@@ -103,20 +105,38 @@ private void initTx() {
...
+       
+			List<PathwayCode> codesFromDb = jdbcTemplate.query(pathwayStatsPsr.getSql(), pathwayStatsPsr.getSetter(), (rs, rowNum) -> {
```
azimov, line 110: Again, code could be tidier here.

chrisknoll:
I've reformatted this code to single tab indent. The Dev group has had conversations in this in the past (at least in the context of Atlas) and I believe that the decison made there is for tabs.  Even in some cases where people still use spaces, there is inconsistency of if space-indent is 2 or 4.  So, to allow people to have the freedom of their own indent sizes, we standardize on 1 tab indent (note not 2 tab indent) and people can configure their editor to give the space between indents that they prefer.

PR #1582, opened 2020-07-15
Use DATEFROMPARTS for date formatted strings.

### LINKED ISSUE #1581: Incidence Rates > Study Window doesn't work in Oracle
Atlas/WebAPI v2.7.6 + Oracle for both CDM and API schemas.

If the incidence rate design includes a study window, the job fails with `ORA-01790: expression must have same datatype as corresponding expression`. The reason for that is the expansion of the [`@EndDateUnions`](https://github.com/OHDSI/WebAPI/blob/b2baf33bceb8a72de3cfe44652ab877fbe2a1a23/src/main/resources/resources/incidencerate/sql/performAnalysis.sql#L61) template:

```
SELECT
...
t.op_end_date   AS followup_end,
...
FROM
...

UNION
SELECT
...
'2019-03-07' followup_end,
...
FROM
...
```

Here the value `'2019-03-07'` is a character datatype, not a datetime datatype. It requires an explicit conversion. I fixed this with the following patch of [IRAnalysisQueryBuilder.java](https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/ircalc/IRAnalysisQueryBuilder.java):

```
@@ -117,7 +117,7 @@
     // add end dates if study window end is defined
     if (analysisExpression.studyWindow != null && analysisExpression.studyWindow.endDate != null && analysisExpression.studyWindow.endDate.length() > 0)
     {
-      StringBuilder endDatesQuery = new StringBuilder(String.format("UNION\nselect combos.target_id, combos.outcome_id, t.subject_id, t.cohort_start_date, '%s' as followup_end, 0 as is_case", analysisExpression.studyWindow.endDate));
+      StringBuilder endDatesQuery = new StringBuilder(String.format("UNION\nselect combos.target_id, combos.outcome_id, t.subject_id, t.cohort_start_date, to_date('%s', 'YYYY-MM-DD') as followup_end, 0 as is_case", analysisExpression.studyWindow.endDate));
       endDatesQuery.append("\nFROM cteCohortCombos combos");
       endDatesQuery.append("\nJOIN  cteCohortData t on combos.target_id = t.target_id and combos.outcome_id = t.outcome_id");
```

It solved the initial problem, but the job failed with another error: `ORA-01861: literal does not match format string`. The reason for that is the expansion of the [@cohortDataFilter](https://github.com/OHDSI/WebAPI/blob/b2baf33bceb8a72de3cfe44652ab877fbe2a1a23/src/main/resources/resources/incidencerate/sql/performAnalysis.sql#L33) template. This can be fixed with the following patch in the same file:

```
@@ -105,9 +105,9 @@
     if (analysisExpression.studyWindow != null)
     {
       if (analysisExpression.studyWindow.startDate != null && analysisExpression.studyWindow.startDate.length() > 0)
-        studyWindowClauses.add(String.format("t.cohort_start_date >= '%s'", analysisExpression.studyWindow.startDate));
+        studyWindowClauses.add(String.format("t.cohort_start_date >= to_date('%s', 'YYYY-MM-DD')", analysisExpression.studyWindow.startDate));
       if (analysisExpression.studyWindow.endDate != null && analysisExpression.studyWindow.endDate.length() > 0)
-        studyWindowClauses.add(String.format("t.cohort_start_date <= '%s'", analysisExpression.studyWindow.endDate));
+        studyWindowClauses.add(String.format("t.cohort_start_date <= to_date('%s', 'YYYY-MM-DD')", analysisExpression.studyWindow.endDate));
     }
     if (studyWindowClauses.size() > 0)
       resultSql = StringUtils.replace(resultSql, "@cohortDataFilter", "AND " + StringUtils.join(studyWindowClauses," AND "));
```

I'm not submitting these changes in a pull request since 1) there might be a better solution for Oracle and 2) I'm not sure if this will work with other SQL flavors. For example, instead of explicit format specifications, it might be simpler to set the `NLS_DATE_FORMAT`.

chrisknoll:
Fixes #1581.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/util/SqlUtils.java
+++ b/src/main/java/org/ohdsi/webapi/util/SqlUtils.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2020 cknoll1.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ohdsi.webapi.util;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ *
+ * @author cknoll1
+ */
+public final class SqlUtils {
+  public static String dateStringToSql(String date) {
+    String[] dateParts = StringUtils.split(date, '-');
+    return String.format("DATEFROMPARTS(%s, %s, %s)", Integer.valueOf(dateParts[0]), Integer.valueOf(dateParts[1]), Integer.valueOf(dateParts[2]));
+  }  
+}
--- a/src/main/java/org/ohdsi/webapi/util/SqlUtils.java
+++ b/src/main/java/org/ohdsi/webapi/util/SqlUtils.java
@@ -18,11 +18,20 @@ package org.ohdsi.webapi.util;
 import org.apache.commons.lang3.StringUtils;
 
 /**
- *
  * @author cknoll1
  */
 public final class SqlUtils {
+
+	/**
+		* Creates a DATEFROMPARTS() expression based on an input in form YYYY-MM-DD.
+		* @param date a string in YYYY-MM-DD format
+		* @return a String containing the DATEFROMPARTS expression
+		* @throws IllegalArgumentException
+		* @author cknoll1
+	*/
   public static String dateStringToSql(String date) {
+		if (!date.matches("^\\d{4}-\\d{2}-\\d{2}"))
+			throw new IllegalArgumentException(String.format("%s is not in valid YYYY-MM-DD format.", date));
     String[] dateParts = StringUtils.split(date, '-');
     return String.format("DATEFROMPARTS(%s, %s, %s)", Integer.valueOf(dateParts[0]), Integer.valueOf(dateParts[1]), Integer.valueOf(dateParts[2]));
   }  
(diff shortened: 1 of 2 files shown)


src/main/java/org/ohdsi/webapi/util/SqlUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ *
+ * @author cknoll1
+ */
+public final class SqlUtils {
+  public static String dateStringToSql(String date) {
+    String[] dateParts = StringUtils.split(date, '-');
+    return String.format("DATEFROMPARTS(%s, %s, %s)", Integer.valueOf(dateParts[0]), Integer.valueOf(dateParts[1]), Integer.valueOf(dateParts[2]));
```
azimov, line 27: I think this will throw an IndexOutOfBoundsException if the string passed isn't in the date format "x-y-z" so some exception handling is perhaps required. Perhaps using a java LocalDate parser would provide a more useful runtime error?

e.g. something along the lines of:
` LocalDate dateObj = LocalDate.parse(date); // will throw a parse error if date is bad`
`return String.format("DATEFROMPARTS(%s, %s, %s)", dateObj.getDayOfMonth(), dateObj.getMonthValue(), dateObj.getDayofYear());`

chrisknoll: Ok I can add some exception handling (but I was thinking that just throwing the exception that didn't have the 3 parts would be good enough).  I'm not for the approach of using a local date parsing becuase we're not letting people use any local date format to specify the YYYY-MM-DD format, this is just the expected format expected from the UI.  So, I can validate that the string splits into 3 parts, and each part is a number (Or use some sort of regex to checkt he format), and I'll put that in with a meaningful exception message (like formatexception or somesuch).

azimov: Just a check with an error message seems reasonable a comment/docstring stating the expected input string format will probably help anyone else using the conversion utility SqlUtils class in the future.

PR #1588, opened 2020-07-28
Added CS endpoint with Vocabulary

aklochkova:
Fixes https://github.com/OHDSI/Atlas/issues/2223

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
+++ b/src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
@@ -101,6 +101,14 @@ public class ConceptSetService extends AbstractDaoService {
     @Path("{id}/expression")
     @Produces(MediaType.APPLICATION_JSON)
     public ConceptSetExpression getConceptSetExpression(@PathParam("id") final int id) {
+        
+        return getConceptSetExpression(id, null);
+    }
+    
+    @GET
+    @Path("{id}/expression/{sourceKey}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ConceptSetExpression getConceptSetExpression(@PathParam("id") final int id, @PathParam("sourceKey") final String sourceKey) {
         HashMap<Long, ConceptSetItem> map = new HashMap<>();
 
         // collect the concept set items so we can lookup their properties later
@@ -122,7 +130,7 @@ public class ConceptSetService extends AbstractDaoService {
         }
 
         // assume we want to resolve using the priority vocabulary provider
-        SourceInfo vocabSourceInfo = sourceService.getPriorityVocabularySourceInfo();
+        SourceInfo vocabSourceInfo = sourceKey == null ? sourceService.getPriorityVocabularySourceInfo() : sourceService.getSource(sourceKey);
         Collection<Concept> concepts = vocabService.executeIdentifierLookup(vocabSourceInfo.sourceKey, identifiers);
 
         // put the concept information into the expression along with the concept set item information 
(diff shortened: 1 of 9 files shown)


src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -101,6 +101,14 @@ public ConceptSetDTO getConceptSet(@PathParam("id") final int id) {
     @Path("{id}/expression")
     @Produces(MediaType.APPLICATION_JSON)
     public ConceptSetExpression getConceptSetExpression(@PathParam("id") final int id) {
+        
+        return getConceptSetExpression(id, null);
```
wivern, line 105: Why not to send `getPriorityVocabularySourceInfo()` as param value instead of `null`?

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -101,6 +101,14 @@ public ConceptSetDTO getConceptSet(@PathParam("id") final int id) {
...
+    }
+    
+    @GET
+    @Path("{id}/expression/{sourceKey}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public ConceptSetExpression getConceptSetExpression(@PathParam("id") final int id, @PathParam("sourceKey") final String sourceKey) {
```
wivern, line 111: There's missed access check for a source

src/main/resources/db/migration/postgresql/V2.7.8.20200728164800__add_conceptset_permission.sql
```diff
@@ -0,0 +1,10 @@
+INSERT INTO ${ohdsiSchema}.sec_permission(id, value, description) VALUES
+  (nextval('${ohdsiSchema}.sec_permission_id_seq'), 'conceptset:*:expression:*:get', 
+  'Resolve concept set expression with data source');
+
+INSERT INTO ${ohdsiSchema}.sec_role_permission(id, role_id, permission_id)
+  SELECT nextval('${ohdsiSchema}.sec_role_permission_sequence'), sr.id, sp.id
+FROM ${ohdsiSchema}.sec_permission SP, ${ohdsiSchema}.sec_role sr
+WHERE sp.value IN (
+  'conceptset:*:expression:*:get'
+) AND sr.name IN ('Atlas users');
```
wivern, line 10: We should not grant access to sourced endpoint to all users, it should depend on source access rights
and then migration should be a little more complicated

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
chrisknoll: I think we should take a closer look at this:  this is a service method that returns a ConceptSetExpression, but with the added information about a target source...the confusing thing to me about it is that why would you get a different concept set expression given a source?

aklochkova: @chrisknoll some vocabularies content unique concept sets while others don't, that's why we need to specify a data source here, which is taken from user's 'Current session' (https://github.com/OHDSI/Atlas/pull/2256/files)
also, if we run PLP/PLE containing this unique CS we need to do it on the same data source from 'Current session', if we run it on a data source from 'Whole application' there might be a situation when this last DS won't contain our necessary CS

chrisknoll: Got it.  It's a strange usecase...in the future (with CIRCE) I think we're going to remove all the embedded concept details from the concept set expression and let resolving the concepts-to-names is handled separately....I'm not sure exactly what the result is when a concept in a concept set is not found int he vocabulary, but if it drops the concept set item from the concept set, that feels like a 'bad' action, but I do understand that it has to try to resolve the concept_ids to names...I'm not suggesting you do anythign differently here, just making a mental note for myself that cocnept sets may end up being modified if a concept isn't found in a vocabulary query...

PR #1623, opened 2020-09-11
Each auth provider can be disabled. For details see README.md

### LINKED ISSUE #1227: Add possibility to disable redundant authProviders
Currently WebAPI supports different auth providers. When application starts it tries to connect to all auth providers. 
In case of AD auth provider configuration we get a problem that WebAPI hangs for about a 1 minute trying to connect to some PostgreSQL host using authDatabase provider and then fails. 
It's good to have options to enable/disable needed authProviders

ssuvorov-fls:
fixes #1227

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -150,6 +150,17 @@
     <security.saml.callbackUrl></security.saml.callbackUrl>
     <security.saml.sloUrl></security.saml.sloUrl>
 
+    <security.auth.windows.enabled>true</security.auth.windows.enabled>
+    <security.auth.kerberos.enabled>true</security.auth.kerberos.enabled>
+    <security.auth.openid.enabled>true</security.auth.openid.enabled>
+    <security.auth.facebook.enabled>true</security.auth.facebook.enabled>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -150,6 +150,17 @@
...
+    <security.auth.kerberos.enabled>true</security.auth.kerberos.enabled>
+    <security.auth.openid.enabled>true</security.auth.openid.enabled>
+    <security.auth.facebook.enabled>true</security.auth.facebook.enabled>
+    <security.auth.github.enabled>true</security.auth.github.enabled>
+    <security.auth.google.enabled>true</security.auth.google.enabled>
+    <security.auth.jdbc.enabled>true</security.auth.jdbc.enabled>
+    <security.auth.ldap.enabled>true</security.auth.ldap.enabled>
+    <security.auth.ad.enabled>true</security.auth.ad.enabled>
+    <security.auth.cas.enabled>true</security.auth.cas.enabled>
+
```
anthonysena, line 163: Shouldn't these be set to "false" by default?

anton-abushkevich (approved):

PR #1628, opened 2020-09-16
Excludes cdm-only sources from caching

### LINKED ISSUE #1627: Dashboard Caching when only results daimon configured
The dashboard caching mechanism fails when a source exists with only a results daimon configured. After tracing through the code with @chrisknoll, we thought it might be best to add a check to the code here: 

https://github.com/OHDSI/WebAPI/blob/fc5b439b0607926b7bbac3e8ff702c511aa0f243/src/main/java/org/ohdsi/webapi/service/CDMResultsService.java#L198

The check is: if the source contains a CDM & Results daimon configured, then cache the dashboard otherwise skip and potentially log an information message.

wivern:
Fixes #1627

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
+++ b/src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
@@ -27,6 +27,7 @@ import org.ohdsi.webapi.source.SourceDaimon;
 import org.ohdsi.webapi.source.SourceService;
 import org.ohdsi.webapi.util.PreparedSqlRender;
 import org.ohdsi.webapi.util.PreparedStatementRenderer;
+import org.ohdsi.webapi.util.SourceUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.JobParametersBuilder;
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/service/CDMResultsService.java
```diff
@@ -108,10 +109,25 @@ public void afterPropertiesSet() throws Exception {
...
             CDMResultsService instance = applicationContext.getBean(CDMResultsService.class);
-			sourceService.getSources()
-				.stream()
-				.filter(s -> s.getDaimons().stream().anyMatch(sd -> Objects.equals(sd.getDaimonType(), SourceDaimon.DaimonType.Results)) && s.getDaimons().stream().anyMatch(sd -> sd.getPriority() > 0))
-				.forEach(s -> warmCache(s.getSourceKey(), instance));
+            Collection<Source> sources =  sourceService.getSources();
+            sources
+                .stream()
+                .filter(s -> SourceUtils.hasSourceDaimon(s, SourceDaimon.DaimonType.Results)
+                        && SourceUtils.hasSourceDaimon(s, SourceDaimon.DaimonType.Vocabulary)
```
chrisknoll, line 116: this may need to be (SourceUtils.hasSourceDaimons(s,Vocabulary) || SourceUtils.hasSourceDaimons(s.CDM)) because CDM's cover having a vocabulary, so we should allow either, but must have one or the other.

src/main/java/org/ohdsi/webapi/util/SourceUtils.java
```diff
@@ -42,4 +42,9 @@ public static String getVocabQualifierOrNull(Source source) {
         return source.getTableQualifierOrNull(Vocabulary);
     }
 
+    public static boolean hasSourceDaimon(Source source, SourceDaimon.DaimonType daimonType) {
```
chrisknoll, line 45: If we want to make it so that 'A CDM implies you have a Vocabulary' we could add custom logic to the hasSourceDaimon and getTableQualifier() calls to check if Vocabulary is requested, but the source doesn't have a Vocabulary, but DOEShave a CDM defined, return the CDM information.  That way, it's a bit seamless to handle cases where you want to check for vocabulary on a source, but only the CDM is defined, the CDM satisfies the presence of a Vocabulary.

chrisknoll (approved):

PR #1632, opened 2020-09-21
Prevent trailing white spaces in asset names #1510

### LINKED ISSUE #1510: Prevent trailing white spaces in asset names?
Asset names (like cohort definition names, concept set names) currently allow trailing (and leading) white spaces. This is maybe not very pretty, and also can [lead to trouble downstream](https://github.com/OHDSI/CohortDiagnostics/issues/39#issuecomment-613944951).

Maybe trim asset names before storing them?

anton-abushkevich:
resolves #1510

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/converter/BaseCcDTOToCcEntityConverter.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/converter/BaseCcDTOToCcEntityConverter.java
@@ -1,6 +1,7 @@
 package org.ohdsi.webapi.cohortcharacterization.converter;
 
 import com.odysseusinc.arachne.commons.utils.ConverterUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.ohdsi.analysis.CohortMetadata;
 import org.ohdsi.analysis.Utils;
 import org.ohdsi.analysis.cohortcharacterization.design.CcResultType;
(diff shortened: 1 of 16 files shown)


src/main/java/org/ohdsi/webapi/cohortcharacterization/converter/BaseCcDTOToCcEntityConverter.java
```diff
@@ -32,7 +32,7 @@ public CohortCharacterizationEntity convert(T source) {
 
     final CohortCharacterizationEntity cohortCharacterization = new CohortCharacterizationEntity();
 
-    cohortCharacterization.setName(source.getName());
+    cohortCharacterization.setName(source.getName().trim());
```
wivern, line 35: Why not to use smth like `StringUtils.trim` to prevent unexpected NPE?

wivern (changes requested):

src/main/java/org/ohdsi/webapi/cohortcharacterization/converter/BaseCcDTOToCcEntityConverter.java
anton-abushkevich: Fixed.

chrisknoll (approved):

wivern (approved):

PR #1667, opened 2020-10-29
getCurrentUser must be called before any update/delete operations()

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/2359

### DIFF
--- a/src/main/resources/db/migration/postgresql/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql
+++ b/src/main/resources/db/migration/postgresql/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql
@@ -1,0 +1,1 @@
+ALTER TABLE ${ohdsiSchema}.source_daimon ADD CONSTRAINT un_source_daimon UNIQUE (source_id,daimon_type);
--- a/src/main/resources/db/migration/postgresql/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql
+++ b/src/main/resources/db/migration/postgresql/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql
@@ -1,1 +1,9 @@
+delete from	${ohdsiSchema}.source_daimon sd1
+where sd1.priority = -1
+	and sd1.source_daimon_id < (
+    select max(source_daimon_id)
+    from ${ohdsiSchema}.source_daimon sd2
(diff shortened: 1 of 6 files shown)


anton-abushkevich:
It works, but:
1. Open a source, uncheck daimon (for example, "temp"), save
2. Open the source, create same daimon again, save
3. Repeat 1 and 2 several times

Now check the database table "source_daimon" - there will be all created and deleted daimons with priority set to "-1". Question: may be it will be cleaner to reuse existing deleted daimon, when creating a new one? I.e. check if we already have a daimon in the source with daimon_type and if we do - then update its name and priority, not creating a new one.

anton-abushkevich:
Also:

if multiple daimons with the same type is not allowed for one source, then it would be appropriate to add a unique constraint on two column: `UNIQUE (source_id, daimon_type)`.

chrisknoll:
I think maybe a reason to not reuse is that it will also reuse the source_id value, which may then accidently associate results (ie: generation results) to a given source_id.

Maybe a solution is to set the source_key to NULL if you do a soft delete?

src/main/resources/db/migration/postgresql/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE ${ohdsiSchema}.source_daimon ADD CONSTRAINT un_source_daimon UNIQUE (source_id,daimon_type);
```
anton-abushkevich, line 1: Migration fails, if the table already has dublicates. First, have to delete those with priority = -1.

anton-abushkevich (changes requested):

anton-abushkevich (approved):

chrisknoll:
This migration failed on SqlServer:

SQL State  : S0001
Error Code : 102
Message    : Incorrect syntax near 'sd1'.
Location   : db/migration/sqlserver/V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql (C:\Git\WebAPI\target\classes\db\migration\sqlserver\V2.8.0.20201103171300__add_source_daimon_unique_constraint.sql)

ssuvorov-fls:
@chrisknoll Thanks a lot and sorry for mistake

PR #1673, opened 2020-11-03
Issue 1670 timestamps for date fields

### LINKED ISSUE #1670: Apply timestamp for date fields
In contrast to the previous 2.7 version date format, inside analyses DTO objects, was changed from default timestamp to formatted string. This could be causing the issue with the timezone when a server and client are in different timezones. 
The suggestion is to revert timestamp for createdDate and modifiedDate fields for all types of analyses.

ssuvorov-fls:
date fields are now read only for jackson
date fields are set to null for export
fixes #1670


src/main/java/org/ohdsi/webapi/util/ConversionUtils.java
```diff
@@ -6,11 +6,17 @@
 import org.ohdsi.webapi.user.dto.UserDTO;
 import org.springframework.core.convert.support.GenericConversionService;
 
+import java.util.Objects;
+
 public class ConversionUtils {
     public static void convertMetadata(GenericConversionService conversionService, CommonEntity source, CommonEntityDTO target) {
-        target.setCreatedBy(conversionService.convert(source.getCreatedBy(), UserDTO.class));
+        if (Objects.nonNull(source.getCreatedBy())) {
```
anton-abushkevich, line 13: We do not need to do null-check here: org.ohdsi.webapi.user.converter.UserEntityToUserDTOConverter already have it.

src/main/java/org/ohdsi/webapi/service/dto/CommonEntityDTO.java
```diff
@@ -1,18 +1,20 @@
 package org.ohdsi.webapi.service.dto;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.ohdsi.webapi.user.dto.UserDTO;
 
 import java.util.Date;
 import org.ohdsi.webapi.CommonDTO;
```
anton-abushkevich, line 9: We do not need null fields in exported JSON. I suggest to add here @JsonInclude(Include.NON_NULL)

anton-abushkevich (dismissed):

wivern (approved):

PR #1717, opened 2020-12-10
prepare for next development iteration 2.9.0-SNAPSHOT

anton-abushkevich:


chrisknoll:
Hold this PR until 2.8 is done.

chrisknoll:
@anton-abushkevich : I'd like to have changes made to this PR for the next iteration.

1) Let's set the version in the POM to 2.9.0-SNAPSHOT.  that's the next development iteration.

2) All the non-snapshot dependencies look good (I think) but let's confirm that those dependencies are up to date.  

3) For hot-fix releases, we'll branch off the v2.8.X tag to create the hotfix. Within that hotfix branch, we'll have the v2.8.(1..x)-SNAPSHOT in the POM.  The process of creating a hotfix should be to cherry-pick commits from master, enforcing that fixes should be put into master first, then pulled out for a patch release.  The exception: when the fix for a patch does not match what is in master due to other changes, making a cherry pick not possible. In the case of this exception, we commit directly into the hotfix branch.  But I want to avoid the situation where a fix in a patch isn't in the master branch. 

Let me know if you can make these updates, or if you prefer, I can do it.

anton-abushkevich:
@chrisknoll OK!
1. Done.
3. I created branch https://github.com/OHDSI/WebAPI/tree/master-2.8.1 from tag v2.8.0.

chrisknoll:
I resolved conflicts, but I'm encountering runtime errors related to something about 'duplicate method in class', but i'm not sure it was from the 'resolve conflicts' that I had to do in this branch, so I'm testing it over on master. When I have a complete picture, I will post it here.

chrisknoll:
Ok false alarm, must have been some local cache issue.

chrisknoll (approved):

PR #1719, opened 2020-12-14
Result schema creation error #1718

### LINKED ISSUE #1718: Result schema creation error
Error in ddl\results\create_index.sql. Column "rank" doesn't exists in table "cohort_sample_element".

anton-abushkevich:
fixes #1718


konstjar (approved):
Lgfm

PR #1740, opened 2021-01-14
raised version of dom4j

### LINKED ISSUE #1739: Vulnerabilities in dom4j
## Issues
XML External Entity (XXE) Injection
> https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing

## Remediation
- Upgrade to org.dom4j:dom4j@2.1.3

ssuvorov-fls:
fixes #1739

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,6 +17,7 @@
     <waffle.version>2.2.1</waffle.version>
     <jna.version>5.5.0</jna.version>
     <postgresql.version>42.2.18</postgresql.version>
+    <dom4j.version>2.1.3</dom4j.version>
     <bouncycastle.version>1.68</bouncycastle.version>
 
     <circe.version>1.9.0-SNAPSHOT</circe.version>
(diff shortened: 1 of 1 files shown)


chrisknoll:
Hi, @ssuvorov-fls . quick question about a pattern i'm seeing in these recent PRs:

I see that for every dependency, the version of the dependency is being extracted to a property, and I was wondering why were were doing this?  We don't do this in all cases, and my understanding of why we would want to set up a property for version is when there are multiple dependencies (like spring) that we want to re-use the version across multiple dependencies so that we assign the version once.  

But, circe doesn't work that way, and neither does dom4j, postgresql drivers, and others.  Can you tell me why we started doing this?

ssuvorov-fls:
@chrisknoll, hi.
I think, that version parameters should be grouped in one place, so we can change them easily, even if version parameter corresponds to one dependency
If you think that such approach is not good, I can put them to original place

chrisknoll:
No, it's fine, i was just curious why it was happening. I think makes total sense for  those libraries that may change frequently (like, our OHDSI dependencies).  I was curious about the general approach but grouping them together so that it's easier to manage is a good reason.  If it's limited to those frequently changing ones, I think that's good, but if every dependency gets put into a property, that might be cluttered.  Thanks for the explanation.

chrisknoll (approved):

chrisknoll (approved):

Issue #1761, opened 2021-01-26
PLE/PLP JSON lacking packageName attribute


anthonysena:
Users should be able to export the JSON representation of their PLE/PLP study from ATLAS to hydrate their specification as described here: https://ohdsi.github.io/Hydra/articles/HydratingPackages.html. 

When attempting this using the JSON supplied in https://github.com/OHDSI/WebAPI/issues/1755#issuecomment-767856055, I encountered the following error:

```r
Error in .jcall("RJavaTools", "Ljava/lang/Object;", "invokeMethod", cl,  : 
  org.json.JSONException: JSONObject["packageName"] not found.
```

If memory serves me right, I think this boils down to the JSON serialization approach used. In the case of PLE/PLP, I needed to override the default serialization to ensure all JSON elements are written even if the property's value is null.

ssuvorov-fls:
@anthonysena 
This json is not for hydra. There is a special tab in Atlas (utilities->download) where you can specify and download the prepared and hydrated study.

anthonysena:
Hi @ssuvorov-fls - the JSON expression that you obtain from Atlas (utilities -> download) is the same JSON that is used when hydrating the study package. The problem here pertains to the fact that the JSON obtained from ATLAS is suppressing elements of the JSON expression where the value is `null` such as the `packageName`. At one point, I had overridden the default serialization method in the StdApi library to allow for all properties to get serialized, even those with a NULL. Can we restore this behavior for PLE/PLP to avoid this issue when users would like to hydrate the design directly in R?

chrisknoll:
Or can we mark the packageName attribute `@include`?

I'm nervous that a missing attribute in JSON will cause failures, since JSON can be very fluid with what attributes exist or not. Shouldn't the hydration from JSON account for this?

anthonysena:
@chrisknoll - yes, I think `packageName` is a special case that should be handled by ensuring that it is always included, even when NULL. The other attributes of the design are optional and should be inferred by default values when not present in the JSON.

Here is [the code](https://github.com/OHDSI/StandardizedAnalysisAPI/blob/master/src/main/java/org/ohdsi/analysis/estimation/design/EstimationAnalysis.java#L159) for the `packageName` on the estimation side. If we added the `@JsonInclude` attribute that should work. I'm also thinking it might be useful to create a "HydraAnalysis" base class that includes this attribute to ensure that any/all Hydra specifications extend this to ensure they work properly.

[...]

PR #1777, opened 2021-02-09
Update StdApi for #1761

anthonysena:
This PR has a dependency on https://github.com/OHDSI/StandardizedAnalysisAPI/pull/43 to obtain the latest updates reflected in the Standardized Analysis API 1.3.1-SNAPSHOT.


chrisknoll:
If we can, can we just get a release of StandardAPI 1.3.1 and move our dependency directly to that version?

anthonysena:
Makes sense - put in a PR to update StdApi and once that is set, I'll revise this PR.

anthonysena:
Just noting that v1.3.1 of StdApi is now released and I've updated this PR to use the released version.

chrisknoll (approved):

PR #1799, opened 2021-02-24
[CI] Fix caching keys

blootsvoets:
- Fixes the GitHub actions cache `restore-keys` property syntax.
- Removed CI tests, since they are currently always failing.


blootsvoets:
@chrisknoll @leeevans this implements the fix we discussed earlier.

chrisknoll (approved):

PR #1813, opened 2021-03-09
Fixed calculation for standard mean difference.

### LINKED ISSUE #1812: Incorrect prevalence Standard Mean Difference calculation
### Expected behavior
Standard Mean Difference (SMD) should be based on Austin paper, which for prevalence stats, you get the same magnitude but opposite direction if you calculate the prevalnce and 'anti-prevalence' of the same population (ie: std diff between males or std diff between females).

### Actual behavior
WebAPI uses an alternate form which does not exhibit the same performance as the Austin formula.

Check the [implementation here](https://github.com/OHDSI/WebAPI/blob/26d2c8d97805f0b22207a913e156e5a7ff143230/src/main/java/org/ohdsi/webapi/cohortcharacterization/report/PrevalenceItem.java#L42).  Note the Austin calculation is defined as:

image

This formula is for dichotomous (binary) variables and p-hats are the prevalence of the variable within each population.

chrisknoll:
Fixes #1812.


src/main/java/org/ohdsi/webapi/cohortcharacterization/report/PrevalenceItem.java
```diff
@@ -38,23 +39,31 @@ public PrevalenceItem(CcPrevalenceStat prevalenceStat, String cohortName) {
...
+     *
+    **/
     @Override
     protected double calcDiff(PrevalenceItem another) {
-        if (count == null || another.count == null || avg == null || another.avg == null) {
-            return 0d;
+      double pTreatment = avg == null ? 0 : avg;
+      double pCompare = another.avg == null ? 0 : another.avg;
+
+      double pooledPrevalence = (pTreatment * (1.0 - pTreatment)) + (pCompare * (1.0 - pCompare));
```
anthonysena, line 53: I think the idea with `pooledPrevalence` is to compute the overall SD? That should take the form:

`SD_overall = SQRT((SD_t ^2 + SD_c^2) / 2)`

src/main/java/org/ohdsi/webapi/cohortcharacterization/report/PrevalenceItem.java
```diff
@@ -38,23 +39,31 @@ public PrevalenceItem(CcPrevalenceStat prevalenceStat, String cohortName) {
...
-
-        double sd1 = Math.sqrt((n1 * count + count) / (n1 * n1));
-        double sd2 = Math.sqrt((n2 * another.count + another.count) / (n2 * n2));
-
-        double sd = Math.sqrt(sd1 * sd1 + sd2 * sd2);
-
-        return (mean2 - mean1) / sd;
+      } else {
+        // calculate the standard mean differnce
+        return (pTreatment - pCompare) / Math.sqrt((pooledPrevalence / 2));
```
anthonysena, line 65: Assuming you revise `pooledPrevalence` as noted above, then this return should be revised to:

`return = (pCompare - pTreatment) / pooledPrevalence;`

Note the change in the numerator to `pCompare - pTreatment`

anthonysena (changes requested):
Noted some differences in the calc when comparing to:

https://github.com/OHDSI/FeatureExtraction/blob/develop/R/CompareCohorts.R#L75-L82

src/main/java/org/ohdsi/webapi/cohortcharacterization/report/PrevalenceItem.java
chrisknoll: I'll work on these, but note:  the proper form (according to the formula):
image

The formula is pTreatment - pControl. I notice that in feature extraction they do m2 - m1, i'm not sure which is the treatment/control in that formula (variable names FTW).  ALso, it's a bit confusing in this bit of code from FE:
```
    m$sd1 <- sqrt(m$mean1 * (1 - m$mean1))
    m$sd2 <- sqrt(m$mean2 * (1 - m$mean2))
    m$sd <- sqrt((m$sd1^2 + m$sd2^2) / 2)
    m$stdDiff <- (m$mean2 - m$mean1)/m$sd
```

Note that they are sqrt() for sd1 and sd2, but then squaring those (to revert the square root) before averaging the two values (by adding them together, then divide by 2).  IT's a bit of a round-about way of getting to the right answer.

To your note:

> I think the idea with pooledPrevalence is to compute the overall SD? That should take the form:
> SD_overall = SQRT((SD_t ^2 + SD_c^2) / 2)

That would be true if SD_t was actually the sqrt(pTreatment * (1.0 - pTreatment)) and SD_c was sqrt((pCompare * (1.0 - pCompare)).

So, I did invent a new term 'pooledPrevalence' because what I was calculating wasn't exactly a standard deviation, but I needed a value to check for 0. However, I can do the entire denominator calcuation (which I can call sdOverall) up front, and if it is zero, handle the zero case.  I'll put that change in.

We still need a ruling on if it should be pTreatment - pCompare or the other way, but from the Java code structure perspective, the StdDiff is determine by a PrevlanceItem.calcDiff(PrevalenceItem other), and we're assuming the 'other' in this call is the 'thing to compare'.

chrisknoll:
@anthonysena I made updates, let's get feedback from @schuemie and @pbr6cornell on if it should be T-C or C-T.  Also, I'm considering updating the variable pTreatment to pTarget.

The core of the question is: if the prevalence is greater in T, should the value be positive or negative?  If it should be positive, then it should be pT - pC. If it should be negative, it should be pC - pT.

chrisknoll:
Ok, based on comments, I've called the 'pooledPrevlance' 'pooledError' since that reflects what that value is.  I'm leaving it as pT - pC for now, and we'll revisit if we decide that we do want it pC - pT, but I feel that feature extraction will want to change to mean1 - mean2.

I think I covered all the comments so far.

anthonysena:
Thanks @chrisknoll for digging into this. A few comments:

> @anthonysena I made updates, let's get feedback from @schuemie and @pbr6cornell on if it should be T-C or C-T. Also, I'm considering updating the variable pTreatment to pTarget.

T-C is the way to go. I opened up https://github.com/OHDSI/FeatureExtraction/issues/123 to address this on the FE side.

anthonysena (approved):

PR #1822, opened 2021-03-19
Checking of age value was added

ssuvorov-fls:
fixes https://github.com/OHDSI/Atlas/issues/2495


chrisknoll:
Hi, I'm not sure if this PR is simply a matter of 'adding a check age validator'?  If I look at the issue, and look at what this PR says it's doing, I would have expected to see a new checker added, and a call made to it to validate check age.  But, there's more happening here.

Could you go into some detail about the new classes you added, why you refactored, why you need a 'max recursion', and what 'fakeTrueValidator' means?

ssuvorov-fls:
I need more than one checker because i need to check a value inside array and nested arrays. 
Max recursion value is used to determine who many times array of CriteriaGroup can be nested inside another CriteriaGroup.
FakeTrueValidator is the last validator in chain of nested validators for CriteriaGroup (i can't use null there).
Age is checked in DemographicHelper.prepareAgeBuilder() function.
Refactoring was made to support iterable validator for arrays and group valudators inside both iterable validators

chrisknoll:
Ok, Let's rename 'FakeTrue' to 'DefaultTrue'.
For the DemographicHelper, looks like that validates the age criteria in a DemographicCriteria, but you can specify age in other criteria, so question is if that criteria is also being checked in ConditionOccurrence, DrugExposure, etc. (not sure if this logic uss the validation rules from Circe or if this is just specific to IR Analysis).
For Max_recursion:  why do we need to limit this?   Any technical limit to recursion will be reached when trying to create the nested criteria so I'm not sure we need to add the complication of a notion of 'max recursion'..

ssuvorov-fls:
Max recursion must be used because the chain of validators is created statically before it is used and I get stack overflow without this parameter.
I'll check age parameter in other classes and rename classes with "Fake" in their names

ssuvorov-fls:
I've deleted recursion in builders and moved it to validators, so nested validators will be created during checking
Also I've created validators for date range, numeric range and period values inside criteria

chrisknoll:
Question:  I noticed a few criteria checks being applied to constructs defined in Circe (such as CriteriaGroups, ConditionEra, etc), vs. other checks that are being applied to constructs that are part of WebAPI (ie: the analysis design objects like CohortCharacterizaton or others).  Should those checks being applied to Circe be part of the Circe package, and WebAPI just calls out to Circe to check the values, or is the mechanism from Circe a different thing than what's being applied here?

ssuvorov-fls:
The idea was to create new mechanism that will allow checking of all entities not only cohorts. So the checking of CriteriaGroups, different criterias, cohorts and so on will gradually move to new checking mechanism from circe

chrisknoll:
> So the checking of CriteriaGroups, different criterias, cohorts and so on will gradually move to new checking mechanism from circe

If you're saying that we'll move the checking code out of circe and into web api (move to new checking mechansim from circe), then I need to object to this:  we have been trying to modularize the code around WebAPI by pulling out functionality from a monolithic codebase.  We did this with Circe and we have done this with Charactierization (and externalized standard interfaces to StandardAnalysisAPI, etc).  So, to pack validation logic from these external sources into WebAPI would be moving in the wrong direction.

ssuvorov-fls:
No, not into webapi, but into new checking mechanism. It can be moved to another module, may be circe or new one.

chrisknoll:
Thanks for the clarification. I think a base capability around 'checker definitions' allowing context-specific validation would be an awesome capability to develop, and then each of our individual components can implement 'standardized' validation rules based on that general capability.  Certainly that can start in WebAPI, but then we can refactor it out to something generalized and have our individual repositories depend on that for their own validation.

ssuvorov-fls:
You are absolutely right. That is the main idea of new checking mechanism

src/main/java/org/ohdsi/webapi/check/validator/common/ArrayForEachValidator.java
```diff
@@ -0,0 +1,51 @@
...
+import org.ohdsi.webapi.check.validator.Validator;
+import org.ohdsi.webapi.check.validator.ValidatorGroup;
+import org.ohdsi.webapi.check.warning.WarningSeverity;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+public class ArrayForEachValidator<T> extends Validator<T[]> {
+    final private List<Validator<T>> validators;
```
anton-abushkevich, line 14: make it ``private final`` please, in that order. 
other field (``validatorGroups``) can also be final

src/main/java/org/ohdsi/webapi/check/validator/ValidatorGroup.java
```diff
@@ -8,14 +8,17 @@
     private List<Validator<V>> validators;
     private List<ValidatorGroup<V, ?>> validatorGroups;
     private Function<T, V> valueForValidationGetter;
+    private Function<T, Boolean> conditionGetter = t -> true;
```
anton-abushkevich, line 11: ``= t -> true`` - is it necessary? you rewrite it in constructor and it is already ``t -> true`` in ``ValidatorGroupBuilder``.
also make fields ``private final``.

anton-abushkevich (changes requested):

src/main/java/org/ohdsi/webapi/check/validator/common/ArrayForEachValidator.java
ssuvorov-fls: fixed

src/main/java/org/ohdsi/webapi/check/validator/ValidatorGroup.java
anton-abushkevich: also make fields ``private final`` please

ssuvorov-fls: sorry, fixed

anton-abushkevich (approved):

Issue #1750, opened 2021-01-19
Use the latest version of Hydra


anthonysena:
Pom.xml is referencing 0.0.11 and the latest release of Hydra is 0.1.1 (https://github.com/OHDSI/Hydra/tags).

anthonysena:
Tagging @olga-ganina @chrisknoll based on our conversation, to determine if we can include in 2.8.1 or a subsequent release based on results of testing this changes once I've completed them.

PR #1835, opened 2021-04-02
hydra version was updated to 0.2.0

ssuvorov-fls:
local skeletons were removed, skeletons from hydra library will be used for analyses (#1750)
must be merged after upgrading plp skeleton in hydra (https://github.com/OHDSI/SkeletonPredictionStudy/issues/10)


src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
```diff
@@ -86,7 +86,9 @@
     private static final String CONCEPT_SET_XREF_KEY_INCLUDED_COVARIATE_CONCEPT_IDS = "includedCovariateConceptIds";
     private static final String CONCEPT_SET_XREF_KEY_EXCLUDED_COVARIATE_CONCEPT_IDS = "excludedCovariateConceptIds";
 
-    private static final String ESTIMATION_SKELETON = "/resources/estimation/skeleton/ComparativeEffectStudy_v0.0.1.zip";
+    // This path can used when skeleton is packed inside WebAPI.war
+    // For example ESTIMATION_SKELETON = "/resources/estimation/skeleton/ComparativeEffectStudy_v0.0.1.zip";
+    private static final String ESTIMATION_SKELETON = "/resources/estimation/skeleton/";
```
anthonysena, line 91: Why are we embedding the Skeleton packages into WebAPI? These should be accessed via Hydra.

anthonysena (dismissed):

src/main/java/org/ohdsi/webapi/estimation/EstimationServiceImpl.java
ssuvorov-fls: Skeletons were embedded because there were no time to wait until hydra release. I didn't remove this code completely for such cases.
Now it uses skeletons from hydra

anthonysena: Fair enough - can we then also remove this constant and code references from the service implementations for Estimation/Prediction?

Also, just saw that Hydra v0.1.2 was released in Feb so perhaps we should target that version? Looking at the [differences in versions](https://github.com/OHDSI/Hydra/compare/v0.1.1...v0.1.2), it should not pose any issues as the only change is the addition of a new skeleton for Prediction.

anthonysena:
In speaking with @alex-odysseus, we'll need to coordinate with the Hydra team to get the new Prediction skeleton released before we can test/merge this work.

chrisknoll (approved):

chrisknoll (approved):

PR #1847, opened 2021-04-08
FEAggregate serialization fixes.

### LINKED ISSUE #1843: Incomplete JSON serialization of feanalysisaggregate
### Expected behavior
Properties such as 'Missing Means Zero' should be included in json serialization (for exporting Cohort Characterization)

### Actual behavior
MissingMeansZero is missing.

1. Create a cohort characterization that includes a custom distribution feature.
2. Go to export CC: Observe missing fields from fe analysis for the distribution.

chrisknoll:
Added missing fields for FEAggregate deserialization and DTO converstion.
Added missing `@Produces` annotation

Fixes #1843

### DIFF
--- a/src/main/java/org/ohdsi/webapi/feanalysis/dto/FeAnalysisAggregateDTO.java
+++ b/src/main/java/org/ohdsi/webapi/feanalysis/dto/FeAnalysisAggregateDTO.java
@@ -1,11 +1,15 @@
 package org.ohdsi.webapi.feanalysis.dto;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
 import org.ohdsi.analysis.TableJoin;
 import org.ohdsi.analysis.cohortcharacterization.design.AggregateFunction;
+import org.ohdsi.analysis.cohortcharacterization.design.FeatureAnalysisAggregate;
 import org.ohdsi.analysis.cohortcharacterization.design.StandardFeatureAnalysisDomain;
+import org.ohdsi.circe.cohortdefinition.builders.CriteriaColumn;
 
-public class FeAnalysisAggregateDTO {
+public class FeAnalysisAggregateDTO implements FeatureAnalysisAggregate {
 
   @JsonProperty("id")
   private Integer id;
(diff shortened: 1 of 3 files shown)


src/main/java/org/ohdsi/webapi/feanalysis/dto/FeAnalysisAggregateDTO.java
```diff
@@ -23,8 +27,22 @@
   private TableJoin joinType;
   @JsonProperty("joinCondition")
   private String joinCondition;
-  @JsonProperty("isDefault")
+  @JsonProperty("default")
```
ssuvorov-fls, line 30: atlas uses "isDefault" field
is it really necessarily to rename it?

chrisknoll: This was a bit confusing to me, but the JSON property was just being called 'default' in the Java object (ie: you access the property via `isDefault()`. Is  that what you mean?  Because while the getter was called 'isDefault' and the setter is 'setDefault', the property itself (under those rules) isn't 'isDefault' it is actually just 'default'.

HOwever, the local variable is called 'isDefault':
```
  @JsonProperty("default")
  private boolean isDefault;
```

So, the setters/getters should have been isIsDefault() and setIsDefault() so, I was simply making the JSON property reflect what the property (via the getter/setter) was labeled.

This is just about how the JSON is serialized, and this would only impact anything where someone tried to load up a previously saved JSON and tried to load it into the app after this change...which considering that the feature is so new, it might not impact anyone but I get your point that it is a breaking change. I just wish that we followed the property naming conventions correctly to beginwith so that we don't have these flaws as part of the API.  

I think I see 3 options:
- Change the field isDefault, make the setters/getters isIsDefault/getIsDefault. This would break the API but not an 'interface' because the interface actually doesn't have the 'isDefault' bean properties declared on it.
- Change the field to 'default', leave the setters/getters isDefault/getDefault. This would break the JSON serialization (because the `default` field is being serialized as `isDefault`.  Is ther actually a javascript restricton on making a field 'default' which is why this was serialized as 'isDefault'?)
- Leave it as it was, but knowing that the 'isDefault' field is accessed via getter/setter as 'isDefault/setDefault'

I can go with the last option, it's fine, i wanted to just make it consistent, but I think the reason it was done was that 'default' is a reserved word in javascript...correct me if i'm wrong?

chrisknoll:
@ssuvorov-fls : I've reconsidered the options, and I think you're right that we shouldn't change these fields on the json serialization, so I reverted that change. Please let me know if you have any other issues.

ssuvorov-fls (approved):

PR #1902, opened 2021-06-22
javax.ws.rs.ForbiddenException in alert if user does not have vocabulary data source role #2568

anton-abushkevich:
Fixes OHDSI/Atlas#2568

### DIFF
--- a/src/main/java/org/ohdsi/webapi/util/GenericExceptionMapper.java
+++ b/src/main/java/org/ohdsi/webapi/util/GenericExceptionMapper.java
@@ -24,6 +24,7 @@ import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.messaging.support.ErrorMessage;
 
 import javax.ws.rs.BadRequestException;
+import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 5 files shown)


ssuvorov-fls (approved):

blootsvoets:
This PR broke some unit tests. @anton-abushkevich could you please fix the tests
`CohortCharacterizationServiceTest.testExportGeneration` and `CohortCharacterizationServiceTest.testExportGenerationWithStrata`? See
https://github.com/OHDSI/WebAPI/runs/2882977025?check_suite_focus=true
for the CI result.

anton-abushkevich:
These tests passed in my local env and in Jenkins. The problem only happens in github check suite. It has something to do with temp directory on the server where build runs, assertion fails in this line: https://github.com/OHDSI/WebAPI/blob/master/src/test/java/org/ohdsi/webapi/cohortcharacterization/CohortCharacterizationServiceTest.java#L147. 

I do not sure why, but checks for other PRs (prior to this PR) were running with maven.test.skip=true, for example https://github.com/OHDSI/WebAPI/pull/1898/checks?check_run_id=2802600666. That is why checks in other PRs are green.

blootsvoets:
Thanks for the followup! I've made a PR to disable the unit tests altogether.

PR #1913, opened 2021-08-04
Audit trail

anton-abushkevich:
**Audit trail** logging in syslog format (RFC 5424).
Env variables to enable and configure:
```
audit.trail.enabled=true
audit.trail.log.file=/tmp/atlas/audit/audit.log
audit.trail.log.extraFile=/tmp/atlas/audit/audit-extra.log 
```

Extra file stores lines exceeding 2048 characters in main log file, referenced by entry id in main log file. 

Log files rotates daily (see src/main/resources/log4j.xml).

Log entries include:
- user's successful and failed login, logout
- job start/stop/failure
- rest calls with users's current location in Atlas frontend (see PR OHDSI/Atlas#2588)


chrisknoll:
Quick question: is it necessary to define a new header `action-location` for auditing requests?  The reason I ask is that this adds some kind of burdeon to any client calling up to WebAPI that they need to set a header for proper audit (If I am understanding the function correctly).  Doesn't that mean that we have to 'trust the client' that they are setting this header? Or could something be set by the header that is potentially compromising?  So my queston boils down to: can auditing in WebAPI be performed independently of any client-submitted information (other than the requested resource, of course).

anton-abushkevich:
@chrisknoll 
`action-location` header is not necessary. Without it, log just will not contain the page URL. So yes - auditing in WebAPI can be performed independently of any client-side application (Atlas or any other).

chrisknoll (approved):

chrisknoll:
Please do not take the default commit message from the squashed commit, this is what was put into the commit message:

```
* [Geospatial] Map or geospatial tab should not be shown for "non-geo" data sources #2412

* [Geospatial] Map or geospatial tab should not be shown for "non-geo" data sources #2412

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation

* Audit trail implementation - sessionId

* Audit trail implementation - sessionId

* Audit trail implementation

* Audit trail - exclude another versions of log4j

* Audit trail implementation

* Audit Trail - expired token fix

* Audit Trail - disabled by default
```

This is not useful since many of the commit messages are just duplicates and don't describe the actual change in the commit.

This is very important: we want commit messages to contain information that is useful to a deveoper to understand the purpose of the commit.  What feature was added?  Did new dependencies get added? What issue does this close?  Any given PR should be able to provide a short summary of changes.  Please make an effort to have clean commit messages.

PR #1920, opened 2021-08-13
Versioning capability

ssuvorov-fls:
Please refer to the feature description: https://github.com/OHDSI/WebAPI/issues/1923


chrisknoll:
Please note: when this PR gets merged, clean up the commit message to represent a clean, concise message about what is changed, with a Fixed message indicating what issue this is referencing.

ssuvorov-fls:
Alex will create issue with detailed description as soon as possible

src/main/java/org/ohdsi/webapi/cdmresults/mapper/DescendantRecordCountMapper.java
```diff
@@ -20,14 +20,8 @@
...
-public class DescendantRecordCountMapper {
-
-	public DescendantRecordCount mapRow(ResultSet rs)
-			throws SQLException {
-		DescendantRecordCount descendantRecordCount = new DescendantRecordCount();
-		descendantRecordCount.setId(rs.getInt("concept_id"));
-		descendantRecordCount.setRecordCount(rs.getLong("record_count"));
-		descendantRecordCount.setDescendantRecordCount(rs.getLong("descendant_record_count"));
-		return descendantRecordCount;
+public class DescendantRecordCountMapper extends BaseRecordCountMapper<DescendantRecordCount> {
```
chrisknoll, line 23: I don't think I like this level of abstraction here (ie: using a `BaseRecordCountMapper` abstraction):  The reason is, the previous DescendantRecordCountWrapper provided everything you need to know by looking at the map row implementation.  You can see what object is created, and how it is mapped.   The new form splits it into separate functions `getResultObject` and the mapping is now found in the base class.   Is there a use case to where there is some other implemenation of `BaseRecordCountMapper`?  I can't find it, so this change feels to me like it only makes the code changes more complicated by adding more classes, splitting up a coheasive class, and adds more lines of code to the PR than necessary.

ssuvorov-fls: Sorry, my mistake. Removed.

chrisknoll: Is `BaseRecordCountMapper` class also needed? Or can this class be removed?

ssuvorov-fls: No, it is not need and it was removed in my last commit

anthonysena:
This PR looks good from my review. Tagging @alex-odysseus so that he can create the issue that links to this pull request describing the functionality.

anthonysena (approved):

PR #1937, opened 2021-09-09
fix ple on bq

### LINKED ISSUE #1933: A PLE analysis fails on an execution on a Google BigQuery data source
### Expected behavior

There are no errors and the analysis results are generated

### Actual behavior

The analysis execution is interrupted with an error in the errorReportR.txt

Message:  Error executing SQL:
java.sql.SQLException: [Simba][BigQueryJDBCDriver](100032) Error executing query job. Message: Table name "xooa2u87cohort_person" missing dataset while no default dataset is set in the request.

- create a PLE analysis
- run it on a Google BigQuery data source clicking on the "Generate" button on the "Executions" tab
- download the results and check an error log

ssuvorov-fls:
parameter `sqlRenderTempEmulationSchema `was added to `runAnalysis.R` script because `CohortMethod/DataLoadingSaving.R` does not use `oracleTempSchema` parameter any more
fixes #1933

### DIFF
--- a/src/main/resources/resources/estimation/r/runAnalysis.R
+++ b/src/main/resources/resources/estimation/r/runAnalysis.R
@@ -40,6 +40,8 @@ tryCatch({
         connectionDetails$password <- function() Sys.getenv("DBMS_PASSWORD")
         connectionDetails$connectionString <- function() Sys.getenv("CONNECTION_STRING")
 
+        options(sqlRenderTempEmulationSchema = resultsDatabaseSchema)
+
         outputFolder <- file.path(getwd(), 'results')
         dir.create(outputFolder)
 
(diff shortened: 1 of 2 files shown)


alex-odysseus (approved):
The solution looks OK

If during an analysis execution in the respective R package there is any special logic when the IF (sqlRenderTempEmulationSchema = NULL) statement is 'true' and there will be no code similar to this
  if (!is.null(oracleTempSchema) && oracleTempSchema != "") {
    warning("The 'oracleTempSchema' argument is deprecated. Use 'tempEmulationSchema' instead.")
    tempEmulationSchema <- oracleTempSchema
  }

it is necessary to fix such places one by one. Hopefully there will be no such

PR #1967, opened 2021-12-12
Security patches

anthonysena:
Aims to address the current dependabot alerts for [log4j](https://github.com/OHDSI/WebAPI/security/dependabot/pom.xml/log4j:log4j/open) and [Google OAuth](https://github.com/OHDSI/WebAPI/security/dependabot/pom.xml/com.google.oauth-client:google-oauth-client/open).

Reference: https://logging.apache.org/log4j/2.x/manual/migration.html.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1126,9 +1126,9 @@
         <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>log4j</groupId>
-      <artifactId>log4j</artifactId>
-      <version>1.2.17</version>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1385,7 +1385,7 @@
         <dependency>
           <groupId>com.google.oauth-client</groupId>
           <artifactId>google-oauth-client</artifactId>
-          <version>1.28.0</version>
```
ssuvorov-fls, line 1388: Pervious version of this library is installed during the init phase. Should it be removed from build section?

ssuvorov-fls (approved):

ssuvorov-fls:
We made some investigation and found out that we use version of log4j below 2, so we are not affected by this vulnerability

Issue #1969, opened 2021-12-13
log4j and Google OAuth security updates


anthonysena:
Creating an issue based on #1967 and the need to update log4j based on [log4j](https://github.com/OHDSI/WebAPI/security/dependabot/pom.xml/log4j:log4j/open) and [Google OAuth](https://github.com/OHDSI/WebAPI/security/dependabot/pom.xml/com.google.oauth-client:google-oauth-client/open).

Reference: https://logging.apache.org/log4j/2.x/manual/migration.html. 

Per @ssuvorov-fls [here](https://github.com/OHDSI/WebAPI/pull/1967#issuecomment-992304592):

> We made some investigation and found out that we use version of log4j below 2, so we are not affected by this vulnerability

Agreed since we are using log4j v1.2.17. However, v1.2.17 is vulnerable to a different CVE as described here: https://www.cvedetails.com/cve/CVE-2019-17571/. 

I made an attempt to use log4j2-core and log4j-api v2.15.0 in #1967 with less success. I then finalized the PR by only updating to use log4j-api v2.15.0 which is not the right solution since the log4j-1.2.17.jar is still present in WEB-INF/lib after inspecting the .war contents.

PR #1970, opened 2021-12-13
log4j v2 update

anthonysena:
Aims to address #1969 by updating log4j to V2.15.0. This is a draft of a PR since it requires additional work to migrate [log4j.xml](https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/log4j.xml) to use the new format. I've included `src/main/resources/log4j2.xml` as part of this PR to give us a starting point.


anthonysena:
I've made my best attempt to migrate [log4j.xml](https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/log4j.xml) to the new log4j v2 format. As part of these changes, we can/should remove the `log4j.xml` when this is merged so I've deleted that file on this branch. If easier for review purposes, I can restore that file so please let me know your thoughts.

anthonysena:
Also worth noting: I've done my best to preserve backwards compatibility while also incorporating the changes that were recently introduced in #1913.

anthonysena:
In comparing the logs from an instance of WebAPI v2.10.1 vs an instance running this branch I have two observations:

- On v2.10.1 with auditing enabled, I see both the `audit.log` and `audit-extra.log` while on this branch I am missing the `audit-extra.log`. Here is a snippet of the settings.xml for both environments:

```xml
<audit.trail.enabled>true</audit.trail.enabled>
<audit.trail.log.file>D:/apache-tomcat-8.5.63/logs/audit/audit.log</audit.trail.log.file>
<audit.trail.log.extraFile>D:/apache-tomcat-8.5.63/logs/audit/audit-extra.log</audit.trail.log.extraFile>
```
- In this branch, some (but not all) logging messages are duplicated. For example:

```

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::       (v1.5.22.RELEASE)

2021-12-14 20:58:47.110 INFO localhost-startStop-1 org.springframework.boot.StartupInfoLogger - [] - Starting WebApi on <server> with PID 2884 (D:\apache-tomcat-8.5.63\webapps\WebAPI\WEB-INF\classes started by <server>$ in D:\apache-tomcat-8.5.63)
2021-12-14 20:58:47.110 INFO localhost-startStop-1 org.springframework.boot.StartupInfoLogger - [] - Starting WebApi on <server> with PID 2884 (D:\apache-tomcat-8.5.63\webapps\WebAPI\WEB-INF\classes started by <server>$ in D:\apache-tomcat-8.5.63)
2021-12-14 20:58:47.119 INFO localhost-startStop-1 org.springframework.boot.SpringApplication - [] - The following profiles are active: default
2021-12-14 20:58:47.119 INFO localhost-startStop-1 org.springframework.boot.SpringApplication - [] - The following profiles are active: default
```

My guess is that there may be duplicate loggers registered in the code base. This feels fairly minor but noting it for completeness.

pom.xml
```diff
@@ -13,7 +13,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- Spring Boot manages spring.version as well -->
     <spring.boot.version>1.5.22.RELEASE</spring.boot.version>
-    <log4j2.version>2.15.0</log4j2.version>
+    <log4j2.version>2.16.0</log4j2.version>
```
alex-odysseus, line 16: I think we should use 2.17.0

src/main/resources/log4j2.xml
```diff
@@ -1,6 +1,40 @@
...
+            <Pattern>%m%n</Pattern>
+          </PatternLayout>
+          <Policies>
+            <TimeBasedTriggeringPolicy interval="1" modulate="true" />
+          </Policies>
+        </RollingFile>
+        <Async name="audit-async">
+          <AppenderRef ref="audit"/>
+        </Async>        
+        <RollingFile name="audit-extra" fileName="${audit.trail.log.file}"
```
alex-odysseus, line 26: This is probably the reason why the second file is empty

It should be 'audit.trail.log.extraFile' instead

pom.xml
anthonysena: Agreed - as a matter of fact, we should probably keep this branch open until the release cycle slows a bit. There have been 3 releases in the past 3 weeks alone: https://logging.apache.org/log4j/2.x/changes-report.html.

src/main/resources/log4j2.xml
anthonysena: Yes indeed - fixed in the latest commit.

chrisknoll:
Ok, for the sake of expediting this fix, let's settle on a confirm fix for the reported vulnerability, and get it into master, and then cherry pick it for a hotfix into our maintenance branch.  Sound good?

alex-odysseus:
@chrisknoll I had to make it properly via a merge to the 'master' and then cherry-picking to the 'master-2.10'. You are right. In the same time Anthony mentioned this branch is going to be open for some time. I merged parts which belong to log4j (didn't include removing of no longer used profiels) right away into 'master-2.10'. There was an intention to release 2.10.2 in the nearest days which include log4j 1.x -> 2.x migration

Known issues:
- duplication of log entries

PR #1972, opened 2021-12-27
Fixes #1971.

### LINKED ISSUE #1971: Backwards compatibility for cohort definition checks
### Expected behavior
When adding new code in a new feature release (ie: 2.x -> 2.x+1) backwards compatibility should be maintained.

### Actual behavior
With PR #1914, the parameter to the cohort definition service `runDiagnostics` chnaged from a `CohrotExpression` to a `cohortDTO`.

Compare the files changed in #1914, and you will find the `runDiagnostics` signature has changed.

chrisknoll:


src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -730,7 +730,17 @@ public InclusionRuleReport getInclusionRuleReport(
...
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Transactional
-	public CheckResult runDiagnostics(CohortDTO cohortDTO) {
+	public CheckResultDTO runDiagnostics(CohortExpression expression) {
+		Checker checker = new Checker();
+		return new CheckResultDTO(checker.check(expression));
+	}
+	
+	@POST
+	@Path("/checkV2")
```
anthonysena, line 739: @alex-odysseus per our discussion, let's find a better name for this endpoint instead of `checkV2`

chrisknoll: I think we agreed that the endpoint name could stand as-is, even tho we understand that it doesn't follow other naming standards.

chrisknoll: the other option is to revert /check, remove /checkV2, and update the UI to handle the old signature, but I believe that the reason for this change was related to having cohortdefinition/check also deal with tags.

anthonysena: We decided to keep this endpoint named `checkV2` so resolving this discussion.

PR #1975, opened 2022-02-03
 Allow specification of vocabulary id in Import Source Codes and use it for import #2653

anton-abushkevich:
i18n for OHDSI/Atlas#2653


chrisknoll:
This PR only seems to include i18n changes.  Is there something missing?

chrisknoll:
Can you resolve these conflicts?

PR #1982, opened 2022-02-15
Issue 1981 add person count

### LINKED ISSUE #1981: PC and DPC and source picker in search results
Issue: RC and DRC are useful proxies for concept prevalence in the search data table, but they’re not accurate enough.

Solution:
1. Add person counts and descendant person counts, as these are available via Achilles results
2. Additionally, to facilitate easy usage of these columns, add a CDM source dropdown to allow users to change the source of the counts

anton-abushkevich:
Resolves #1981

### DIFF
--- a/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
@@ -102,6 +102,7 @@ public abstract class AtlasSecurity extends Security {
             // DDL service
             .addRestPath("/ddl/results")
             .addRestPath("/ddl/cemresults")
+            .addRestPath("/ddl/achilles")
 
             .addRestPath("/saml/saml-metadata")
             .addRestPath("/saml/slo")
--- a/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
@@ -102,7 +102,7 @@ public abstract class AtlasSecurity extends Security {
             // DDL service
             .addRestPath("/ddl/results")
             .addRestPath("/ddl/cemresults")
-            .addRestPath("/ddl/achilles")
+            .addRestPath("/ddl/extra")
 
             .addRestPath("/saml/saml-metadata")
             .addRestPath("/saml/slo")
--- a/src/main/resources/resources/cdmresults/sql/loadConceptRecordPersonCountCache.sql
+++ b/src/main/resources/resources/cdmresults/sql/loadConceptRecordPersonCountCache.sql
@@ -2,6 +2,6 @@ SELECT
     concept_id,
     record_count,
     descendant_record_count,
-    record_count as person_count,
-    descendant_record_count as descendant_person_count
+    person_count,
+    descendant_person_count
 FROM @resultTableQualifier.achilles_result_concept_count;
--- a/src/main/resources/resources/cdmresults/sql/loadConceptRecordPersonCountCache.sql
+++ b/src/main/resources/resources/cdmresults/sql/loadConceptRecordPersonCountCache.sql
@@ -1,0 +1,7 @@
+SELECT
+    concept_id,
+    record_count,
+    descendant_record_count,
+    record_count as person_count,
+    descendant_record_count as descendant_person_count
+FROM @resultTableQualifier.achilles_result_concept_count;
--- a/pom.xml
+++ b/pom.xml
@@ -32,8 +32,8 @@
     <jackson.version>2.10.5</jackson.version>
     <jackson.databind.version>2.10.5.1</jackson.databind.version>
     <start-class>org.ohdsi.webapi.WebApi</start-class>
-    <skipUnitTests>true</skipUnitTests>
-    <skipITtests>true</skipITtests>
+    <skipUnitTests>false</skipUnitTests>
+    <skipITtests>false</skipITtests>
     <miredot.phase>package</miredot.phase>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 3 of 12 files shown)


src/main/resources/resources/cdmresults/sql/loadConceptRecordPersonCountCache.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    concept_id,
+    record_count,
+    descendant_record_count,
+    person_count,
+    descendant_person_count
+FROM @resultTableQualifier.achilles_result_concept_count;
```
anthonysena, line 7: How is this table created? I don't see references to create this table via WebAPI.  @resultTableQualifier.achilles_result_concept_count. I believe this was discussed as part of https://github.com/OHDSI/Achilles/pull/592. Should this be added to the DDL endpoint to create this table similar to the way we're creating the concept_hierarchy?

anthonysena: Looking at [create_result_concept_table.sql](https://github.com/OHDSI/Achilles/blob/master/inst/sql/sql_server/analyses/create_result_concept_table.sql) on master, it does not have the `person_count` and `descendant_person_count` columns in the resulting `achilles_result_concept_count` table. Most likely you are referring to the version of [create_result_concept_table.sql from OHDSI/Achilles#592](https://github.com/OHDSI/Achilles/pull/592/files#diff-1ecc6781b665ec932a4dbc524857e55fbfb09723c661a4c251c71af49ab9a6fb) which was not merged into master. Reason being: the `achilles_result_concept_count` table is something that WebAPI can use vs. something that Achilles requires. 

In my opinion, we should have a mechanism on one of the `WebAPI/ddl` endpoints to produce this table. I'm not sure how much work this would require but my hope is that we could port over the Achilles SQL and make an endpoint for creating this table.

anton-abushkevich: If we add `achilles_result_concept_count` table creation to `WebAPI/ddl`, the table will drop and recreate without `person_count` and `descendant_person_count` columns on Achilles run: https://github.com/OHDSI/Achilles/blob/master/inst/sql/sql_server/analyses/create_result_concept_table.sql#L2. And vice versa: if the table was already created and filled by Achilles, ddl script will recreate it, loosing the cached data in the process.

Furthermore, if we create and try to fill the table through WebAPI ddl, the script will fail without `achilles_results` table, which is created in Achilles. And we will be forced to run Achilles prior to run ddl script, which are presently two independent runs. For now, Achilles doesn't need any table in result schema from ddl, and Webapi ddl doesn't need Achilles' tables. 

In this PR, if WebAPI cannot read `person_count` and `descendant_person_count` from `achilles_result_concept_count` for whatever reason (for example, select in `loadConceptRecordPersonCountCache.sql` fails because there is no table or columns), it will run sql select to aquire these data directly from `achilles_results` table (loadConceptRecordPersonCountCacheFull.sql). It is the same mechanism, that was implemented for RC/DRC. Chaching the data in `achilles_result_concept_count` is optional, but very desirable. That is the OHDSI/Achilles#592 was for.

anthonysena: Thanks @anton-abushkevich - this is helpful context for this PR. I need to consult @fdefalco on this item again since the Achilles package still exposes a function `optimizeAtlasCache` which should either be deprecated or we should include the work in OHDSI/Achilles#592.

anthonysena: I should clarify - we should include the DDL/SQL for building the cache. More specifically it should provide the SQL used in this [function call from Achilles](https://github.com/OHDSI/Achilles/blob/master/R/Achilles.R#L843-L891) which is in [/inst/sql/sql_server/analyses/create_result_concept_table.sql](https://github.com/OHDSI/Achilles/blob/master/inst/sql/sql_server/analyses/create_result_concept_table.sql).

This feels similar to the approach we have for the `concept_hierarchy` table build (described [here](https://github.com/OHDSI/WebAPI/wiki/CDM-Configuration#results-schema-setup)). The difference between `achilles_result_concept_count` and `concept_hierarchy` is that `concept_hierarchy` **must exist** for the Data Sources functionality to work. `achilles_result_concept_count` is optional as @anton-abushkevich described.

alondhe: I agree with @anthonysena on this. If we can't do it in Achilles, then an endpoint is the next best thing, for now. So would we then have the table creation and insert to fill it in the same /ddl endpoint?

@anton-abushkevich - what do you think?

anton-abushkevich: Sure, we can add separate DDL for `achilles_result_concept_count`. Just keep in mind that it must be run after Achilles, because it depends on `achilles_results` table.

anthonysena (changes requested):

src/main/java/org/ohdsi/webapi/shiro/management/AtlasSecurity.java
```diff
@@ -102,6 +102,7 @@ protected FilterChainBuilder setupProtectedPaths(FilterChainBuilder filterChainB
             // DDL service
             .addRestPath("/ddl/results")
             .addRestPath("/ddl/cemresults")
+            .addRestPath("/ddl/achilles")
```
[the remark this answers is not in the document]
anton-abushkevich: Done.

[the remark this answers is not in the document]
anthonysena: Thank you!

anthonysena (approved):

PR #1993, opened 2022-02-27
Complete caching of Achilles results in Data Sources #1978

### LINKED ISSUE #1978: Complete caching of Achilles results in Data Sources
Performance should be comparable to AchillesWeb (JSON file loading). 

Solution: cache Achilles results as JSON in WebApi DB.

anton-abushkevich:
Resolves #1978


chrisknoll:
I'm not sure I feel good about what is being done here: It removes the previous implementation using the Spring Framework to cache.  It uses the robust cache annotations (which you can find defined [here](https://github.com/spring-projects/spring-framework/tree/main/spring-context/src/main/java/org/springframework/cache),   My understanding was that you would be adding to the existing body of work, not ripping out the existing and replacing with a (roll-your-own?) implementation around caching which leverages AOP constructs (which I think is exactly what the Spring framework does).

I'm unclear why we would write our own set of annotations and AOP implementations when we have a standardized library (I think it even adheres to a standard spec in the Java API).

ssuvorov-fls:
@chrisknoll 
Hello 
We saw https://github.com/OHDSI/WebAPI/commit/b1173bc486d889b10a9758d5b13468a3a44ac016 and decided that it didn't suit our goal of reduction of used memory and time of preparing cache
After initial research we found that there were 2 posibilities for spring cache to store data - memory and disk. Solution with memory is not good as it will take a lot of memory and take a lot of time during each restart for create cache. Storing data on disk is not good too because this approach does not guarantee data integrity and it will require replication and folder configuration.
Because of feature delivery time constraints we decided to create another mechanism for storing cache in database and better control of creating cached values and putting them into database.
We can return to using spring cache and "write-behind" mechanism in future releases when any other cache related functionality to be implemented

Also we would like to put information about 2 mechanisms of cache into documentation and mention that EhCache caching is preferred

chrisknoll:
> Storing data on disk is not good too because this approach does not guarantee data integrity and it will require replication and folder configuration.

Can you elaborate on this further?  What do you mean caching to filesystem risks data integrity?

`Because of feature delivery time constraints we decided to create another mechanism for storing cache in database and better control of creating cached values and putting them into database.
`
Is it a good idea to rush to an implementation by creating our own ad-hoc solution to a problem that appears to be solved in spring?  Maybe the correct approach is to roll our own implementation, but what support and documentation exists around this custom implementation? There are plenty of documents, articles and examples out there that were created by numerous authors that demonstrate the spring cache implementation.  Isn't it wroth using a developed package to handle this instead of writing our own?

I don't think this caching function is on the critical path to 2.11. let's take our time and use existing frameworks to handle caching.

chrisknoll:
I also just want to clarify that I hope you don't take my comments as scornful or negative.  I both appreciate the hard work (with over a thousand lines of code added, and 111 files changed, it's a lot of work) and also trying to uphold the principles of leveraging existing work.  The initial implementation was intentionally naive and basic but demonstrated how to get a cache provider injected into the application via spring.  The hope was to expand it to either provide necessary configuration to cache in the most effective way possible.  I'm hoping we can still do this, and my concern with this PR is we're moving away from that principle.   I'm sorry I was not involved in conversations with this sooner, but it would have been helpful if the original issue declared what type of changes were going to be implemented before the work of creating the PR took place.

pom.xml
```diff
@@ -239,14 +242,25 @@
     <sensitiveinfo.analysis.extensions>txt</sensitiveinfo.analysis.extensions>
 
     <!-- Cache Config -->
-    <spring.cache.jcache.config>classpath:config/ehcache.xml</spring.cache.jcache.config>
     <cdm.result.cache.warming.enable>true</cdm.result.cache.warming.enable>
+    <!-- Enable or disable warming cdm cache by cron -->
+    <cdm.cache.cron.warming.enable>true</cdm.cache.cron.warming.enable>
```
anthonysena, line 247: What happens when this configuration setting is set to `false`? Is there a fall back when the cache is not warmed?

ssuvorov-fls: @anthonysena 
Yes, it has fall back in case of absent cached value in cache table.
When there is no cached value the server will call the original queries, get the data and put it into cache.
This parameter is used only for periodical warming of the cache. The logic is in AchillesCacheAspect class

anthonysena: Great thanks for confirming @ssuvorov-fls!

anthonysena:
From a discussion yesterday between @chrisknoll @alex-odysseus and myself, we've agreed to merge these changes not withstanding the concerns raised by @chrisknoll. We noted that in the future, we want to have more collaborative discussions on the design of features so there are no surprises when a pull request shows up for review. This is particularly important when existing functionality is being replaced by a different approach.

anthonysena (approved):

PR #2028, opened 2022-05-04
Add javadoc comments to  WebAPI endpoints

ablack3:


src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -439,7 +445,7 @@ public CohortDTO createCohortDefinition(CohortDTO dto) {
 	 * Returns the cohort definition for the given id
 	 *
 	 * @param id The cohort definition id
-	 * @return The CohortDefinition
```
ablack3, line 442: @anthonysena Is there a reason to use "CohortDefinition" instead of "cohort definition"? Does one imply something the other does not?

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -467,6 +472,13 @@ public CohortDTO getCohortDefinition(final int id) {
...
 
+        /**
+         * Check that a cohort exists
+         * 
+         * @param id The cohort definition id
+         * @param name The cohort definition name
+         * @return 1 if the a cohort with the given name and id exist in WebAPI and 0 otherwise
+         */
 	@GET
 	@Path("/{id}/exists")
```
ablack3, line 483: @anthonysena - I didn't understand this endpoint. It always returned 0 when I tried it even when I passed in the name as a query parameter. Is it used to check the if the id has a given name?

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -376,9 +382,9 @@ public GenerateSqlResult generateSql(GenerateSqlRequest request) {
 	}
 
 	/**
-	 * Returns all cohort definitions in the WebAPI database
```
ablack3, line 379: This endpoint returns metadata and not the actual definitions.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -682,7 +701,13 @@ private List<ConceptSetExport> getConceptSetExports(CohortDefinition def, Source
...
-
+        
+        /**
+         * Return concept sets used in a cohort definition
+         * 
+         * @param id a cohort definition id
+         * @return concept sets used in the cohort definition
+         */
 	@GET
 	@Path("/{id}/export/conceptset")
```
ablack3, line 712: I also got back raw data from this endpoint. I wasn't able to parse it into anything understandable.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
anthonysena: It is probably easiest for a consumer of the method to read "cohort definition" as opposed to "CohortDefinition" so I'd support changing that. This also ties to @alex-odysseus suggestion around a glossary of terms so we should think about having a "cohort definition" definition that then describes what is meant by that term.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
anthonysena: That's right - by way of example, you can try this URL:

https://api.ohdsi.org/WebAPI/cohortdefinition/1779286/exists?name=mike%5Ftesting

This will return 1 since the name "mike_testing" is used by another cohort definition that does not have the id 1779286. If the ID and name match, it means that the cohort name is used by that cohort id and that's acceptable. The # returned here indicates the # of cohort definitions that have that name. 

Tracing the code a bit the service [here](https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java#L472) uses this hibernate call: 

https://github.com/OHDSI/WebAPI/blob/c67260e51fab2177b7bcfa114e939537a0f9619b/src/main/java/org/ohdsi/webapi/cohortdefinition/CohortDefinitionRepository.java#L43-L44

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
anthonysena: Yes this comment is stale. Here is what I currently see:

```
	/**
	 * Returns all cohort definitions in the cohort schema
	 *
	 * @return List of cohort_definition
	 */
```

I'd propose this should read "Returns all cohort definition metadata in the WebAPI database" and it returns the cohort definition metadata.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
anthonysena: Yes, this particular endpoint is actually streaming the .ZIP file that contains the exported information about the concept sets in the cohort definition. The annotations here provide a bit of a hint:

https://github.com/OHDSI/WebAPI/blob/c67260e51fab2177b7bcfa114e939537a0f9619b/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java#L685-L686

When the browser encounters the `APPLICATION_OCTET_STREAM` it will know that it is receiving a file and start the download process. If you call this endpoint on a valid cohort definition with at least 1 concept set, it should automatically start a download of the .ZIP file.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
chrisknoll: @anthonysena is right, but i just want to clarify the rule:  a name is in use when it exists for a cohort other than the cohort_id that you specify in the id param.   When you are making a new cohort, then the name must be unique across all cohort definitions. When you are updating an existing one, then the name check should exclude the cohort you are checking from consideration.

It's a little confusing logic, but there was an insistence that cohort IDs aren't enough to uniquely identify a cohort, but the names should also be unique.  But you're raising the same questions that we raised ourselves when that rule was first implemented, so you are asking allt he right questions!

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
chrisknoll: I'm not sure I like the term 'metadata'....basically, cohort defintiions are huge objects (a single expression can be many megabytes). The purpose of this endpiont is just to return a summary list of cohort definitions (an id, name, maybe some tags).   Is that 'metadata'?  Is the name a person considered their 'metadata'?  Not sure.  I feel like when you say the 'metadata of a cohort definition' you mean returning the fields and field-types that define a cohort definition.  

This is more of a 'summary view of all cohort definitions in the WebAPI database'.   Specifically, this is used wherever we need to allow the user to select a cohort definition from the WebAPI service.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
ablack3: Thanks!

> When you are making a new cohort, then the name must be unique across all cohort definitions.
How is this checked? Is the {id}/exists endpoint used?

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
ablack3: ROhdsiWebApi uses the term metadata to describe what this and other WebAPI endpoints return. https://ohdsi.github.io/ROhdsiWebApi/reference/getCohortDefinitionsMetaData.html The term seems to fit to me. I suppose we should have a definition of what a cohort definition is. My understanding is that a cohort definition is exactly the data that is represented in the json expression. When it was last updated are not part of the cohort. Is the name or id part of the cohort definition? Seems like this is data (information) about the cohort definition in the context of a particular WebAPI instance and a particular user.

One question: The "hasWriteAccess" field tells the caller if they have write access to the cohort definition. Its value depends on who made the call. Is that right?

image

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
ablack3: Super helpful. Thanks.

anthonysena:
Hi - I'm going to rebase & merge this work into the `issue-1203-documentation` in preparation for the document-a-thon. We'll then make use of `issue-1203-documentation` as root branch for our collaboration on the documentation.

PR #2036, opened 2022-05-23
 Achilles Heel report is deprecated in Achilles #2489

anton-abushkevich:
Backend changes for OHDSI/Atlas#2489


chrisknoll:
Normally, if we remove an endpoint, that's an API breaking change, and this would go into a major version release update.

The question is: should we remove it from WebAPI or just modify the front end to hide the HEEL reports?

chrisknoll (approved):

chrisknoll:
The approval means that the change is OK but the question about releasing it as a minor or major release needs to be decided.

PR #2041, opened 2022-06-15
security fixes: add embed tomcat dependency 8.5.78

anton-abushkevich:


alex-odysseus (changes requested):
There should be a pull request to 'master' first and then a cherry-pick to 'master-2.11'

Any reason not to use the latest 8.5.81? Please additionally extract a new property and place it on top as we do for the others

Please place two last dependencies right under

<artifactId>spring-boot-starter-jdbc</artifactId>

as in the experimental branch

Additionally I would add a few comments to the pom.xml that we are intentionally overriding the Tomcat version from Spring Boot 1.5.22.RELEASE and assuming tomcat-jdbc.jar will be in the Tomcat installation folder (\apache-tomcat-8.5.xxx\lib)

Furthermore it is important to mention that the WebAPI.war.original file should not contain overridden libraries


Additionally please place

anton-abushkevich:
I will close this PR and create another one form master branch.

Issue #2049, opened 2022-06-29
Maven settings file is not imported correctly in Docker build


gutte:
### Expected behavior

Maven settings file `WebAPIConfig/settings.xml` should be used when building using the Dockerfile.

### Actual behavior

The settings are not included.

### Steps to reproduce behavior

`docker build -t webapi .`

chgl:
Most if not all of the parameters can also be set using environment variables. For example, if you're using docker-compose:

```yaml
services:
  ohdsi-webapi:
    image: docker.io/ohdsi/webapi:2.11.1@sha256:9bcb5003e6504669b53c94f3e0b4aa6ad9b5c78741f772ca402579632ce64fb3
    environment:
      DATASOURCE_DRIVERCLASSNAME: org.postgresql.Driver
      DATASOURCE_URL: jdbc:postgresql://ohdsi-omop-db:5432/ohdsi
      DATASOURCE_USERNAME: postgres
      DATASOURCE_PASSWORD: postgres
      DATASOURCE_OHDSI_SCHEMA: ohdsi
      SPRING_JPA_PROPERTIES_HIBERNATE_DIALECT: org.hibernate.dialect.PostgreSQLDialect
      SPRING_JPA_PROPERTIES_HIBERNATE_DEFAULT_SCHEMA: ohdsi
      SPRING_BATCH_REPOSITORY_TABLEPREFIX: ohdsi.BATCH_
      FLYWAY_DATASOURCE_DRIVERCLASSNAME: org.postgresql.Driver
      FLYWAY_DATASOURCE_URL: jdbc:postgresql://ohdsi-omop-db:5432/ohdsi
      FLYWAY_DATASOURCE_USERNAME: postgres
      FLYWAY_DATASOURCE_PASSWORD: postgres
      FLYWAY_LOCATIONS: classpath:db/migration/postgresql
      FLYWAY_PLACEHOLDERS_OHDSISCHEMA: ohdsi
      FLYWAY_SCHEMAS: ohdsi
      SECURITY_CORS_ENABLED: "true"
      SECURITY_ORIGIN: "http://ohdsi.127.0.0.1.nip.io"
    restart: unless-stopped
    ipc: none
    security_opt:
      - "no-new-privileges:true"
    cap_drop:
      - ALL
    deploy:
      resources:
        limits:
          memory: 4096m
    read_only: true
    tmpfs:
      - /tmp
    privileged: false
```

[...]

PR #2050, opened 2022-06-29
Issue 2049 add maven settings to Dockerfile

gutte:
Addresses issue #2049 

Copies the maven settings file from project directory `WebAPIConfig/settings.xml` to the default location in the builder `~/.m2/settings.xml`


leeevans:
using environment vars with docker-compose is the preferred approach

PR #2053, opened 2022-07-05
Provide implementation for calling calcDiff on ComparativeDistributionItem.

### LINKED ISSUE #2002: Error when comparing 2 CC cohorts with distribution stats
### Expected behavior
Should be able to compare 2 cohorts when one has a distribution stat and the other does not.

### Actual behavior

You get the following exception:
```
Caused by: java.lang.ClassCastException: org.ohdsi.webapi.cohortcharacterization.report.PrevalenceItem cannot be cast to org.ohdsi.webapi.cohortcharacterization.report.DistributionItem
	at org.ohdsi.webapi.cohortcharacterization.report.DistributionItem.calcDiff(DistributionItem.java:8)
	at org.ohdsi.webapi.cohortcharacterization.report.ComparativeItem.calcDiff(ComparativeItem.java:54)
	at org.ohdsi.webapi.cohortcharacterization.report.ComparativeItem.<init>(ComparativeItem.java:43)
	at org.ohdsi.webapi.cohortcharacterization.report.ComparativeDistributionItem.<init>(ComparativeDistributionItem.java:34)
	at org.ohdsi.webapi.cohortcharacterization.report.AnalysisItem.getComparativeItems(AnalysisItem.java:89)
	at org.ohdsi.webapi.cohortcharacterization.CcServiceImpl.prepareReportData(CcServiceImpl.java:833)
```

Create a cohort characterization where you have people in one cohort, and zero people in the other.  
Create a Cohort Characterization that includes a distribution type of analysis. (like age)
Open CC results, and go into comparative mode by selecting both cohorts.
You will find the UI hangs, and the above error will be presented in the logs.

### Notes:

This issue is related to the handling when one cohort has the statistic and the other does not.  In this case, we create a 'dummy' zero statistic for use in the comparison, however, we assume a 'Prevalence Item' [here](https://github.com/OHDSI/WebAPI/blob/35a987fffa6d9cd20ebb6c95efb0942ac456ff29/src/main/java/org/ohdsi/webapi/cohortcharacterization/report/ComparativeItem.java#L46).  If the second is null, we create a dummy PrevalenceItem() and then call first.calcDiff(second). `first` is a distributionItem, and being passed a `PrevalenceItem` which can't be down-cast into a distribution item, and therefore the call fails.

I believe if we move calcDiff() to the CompariativePrevlanceItem and ComparativeDistribtuionItem, we can bypass this problem by using the class-local instances of first and second.

I'd like to point out that the paramerized types and inheritance in the report objects is quite confusing.

I think the solution is to have a non-paramaterized calcDiff() function  impelemtned at the CompartiveItem and ComparativeDistrubtionItem.

chrisknoll:
Fixes #2002

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/report/ComparativeDistributionItem.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/report/ComparativeDistributionItem.java
@@ -3,6 +3,7 @@ package org.ohdsi.webapi.cohortcharacterization.report;
 import org.ohdsi.webapi.cohortdefinition.CohortDefinition;
 
 import java.util.Objects;
+import org.ohdsi.webapi.cohortcharacterization.dto.CcDistributionStat;
 
 public class ComparativeDistributionItem extends ComparativeItem {
     private final Double targetStdDev;
@@ -29,6 +30,8 @@ public class ComparativeDistributionItem extends ComparativeItem {
     private final String aggregateName;
     private final Boolean missingMeansZero;
 
+		private static final CcDistributionStat EMPTY_ITEM;
+		
     public ComparativeDistributionItem(DistributionItem firstItem, DistributionItem secondItem, CohortDefinition firstCohortDef,
                                        CohortDefinition secondCohortDef) {
         super(firstItem, secondItem, firstCohortDef, secondCohortDef);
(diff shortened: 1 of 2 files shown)


anthonysena:
Closing as duplicate of #2003

PR #2068, opened 2022-08-15
 Add valid_start_date and valid_end_date to concept endpoint #2057 

### LINKED ISSUE #2057: Add valid_start_date and valid_end_date to concept endpoint
In support of https://github.com/OHDSI/Atlas/issues/2723, we need to add the valid start and end dates of concepts to the concept endpoint.

anton-abushkevich:
Resolves #2057


anthonysena:
> Does the latest change to the SOLR query (that includes start/end date fields) require a rebuild of the core or will the core contain these fields already? I ask because we need to know if we have to tell people to rebuild their SOLR cores to support these fields.

Best I can tell, these changes will require a rebuild of the SOLR cores to incorporate the date fields.

chrisknoll:
Do you happen to know if the fields aren't in SOLR, will the UI break?

chrisknoll (approved):

PR #2070, opened 2022-08-15
Support storage of more description fields #2058

### LINKED ISSUE #2058: Support storage of more description fields
In support of https://github.com/OHDSI/Atlas/issues/2724, the WebAPI schema will need to store description text field values for all study asset types.

anton-abushkevich:
Resolves #2058


src/main/resources/db/migration/postgresql/V2.12.0.20220719110154__add_description_to_assets.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE ${ohdsiSchema}.pathway_analysis ADD description varchar(1000) NULL;
+ALTER TABLE ${ohdsiSchema}.concept_set ADD description varchar(1000) NULL;
+ALTER TABLE ${ohdsiSchema}.cohort_characterization ADD description varchar(1000) NULL;
```
chrisknoll, line 3: In all the documentation I've seen, the syntax of adding a column to a table in postgresql is ALTER TABLE ... ADD COLUMN...   Are you sure this works without using ADD COLUMN?

chrisknoll: Actually, it does look like the column syntax is optional: https://www.postgresql.org/docs/current/sql-altertable.html

ssuvorov-fls: Yes, "Column" word is not necessarily

PR #2083, opened 2022-08-20
configurable scopes for OpenID auth

m0nhawk:
I found out that some of the OpenID providers requires specifying special scopes.

This PR includes new configuration option `security.oid.scopes` which should have values space-separated values scopes, e.g. `openid user`.


chrisknoll:
Any chance that you could add a reasonable 'default' value in pom.xml or is a null value an acceptable value (null is what you'll get in the `@Value` annotation if there's no value copied in from the pom.xml)?

chrisknoll:
I found some documentation on OpenID 'scope' but not sure if the [scopes described there](https://auth0.com/docs/get-started/apis/scopes/openid-connect-scopes) are the same property as the scope you are referring in this PR.

It seems to indicate that `openid` is a required scope, but it's also possible that it assumes that scope will be used for any access, so maybe the property is for 'additional scopes'.  I don't know enough about OpenID to know.

src/main/java/org/ohdsi/webapi/OidcConfCreator.java
```diff
@@ -50,6 +53,12 @@ public OidcConfiguration build() {
         conf.setLogoutUrl(logoutUrl);
         conf.setWithState(true);
         conf.setUseNonce(true);
+
+        String scopes = "openid";
+        if (extraScopes != null && !extraScopes.isEmpty()){
+            scopes += extraScopes;
```
alex-odysseus, line 59: A blank in between "openid" and "extraScopes" is missing. It shouldn't work unless you start extra scopes value in the property file with a blank

alex-odysseus (changes requested):

chrisknoll:
If the default behavior here results in the same behavior pre-PR, then I'm fine with these changes.   Are you asking if you should provide unit tests for these changes?

chrisknoll:
Ok, I found some security tests (that make HTTP requests) [here](https://github.com/OHDSI/WebAPI/blob/master/src/test/java/org/ohdsi/webapi/test/SecurityIT.java).  Not sure if that makes a good model because, in the case of openid, we'd need some sort of testing context to mock an openid provider.   Having a test case for this would be ideal, but we've left testing confirmation to external parties and accepted PRs on their certification.

@alex-odysseus  or @anthonysena : do you have any OpenID environment set up that we could run an openID test using this new `scope` functionality?

anthonysena:
@alex-odysseus perhaps ELIXIR could be a suitable place to test this functionality? Our test environment has expired but wondering if you have access to one with OpenID?

chrisknoll (approved):

chrisknoll:
We'll accept this PR now because it appears to be low risk, and we can work on formalized unit tests for this functionality at a later date.

alex-odysseus:
We will check it internally with Google I believe. We can also give it a try with Life Science (aka ELIXIR AAI) though I remember there were a few complications when we were implementing OpenID Connect (OIDC) in ARACHNE for this particular Identity Provider when scopes are involved @chrisknoll @anthonysena

PR #2087, opened 2022-09-20
Impelement Phoebe 2.0

chrisknoll:
This PR contains the backend code to support OHDSI/ATLAS#2728.

New search option `isLexical` performs string searching on concept_synonym.
New endpoint {sourceKey}/lookup/recommended:  returns the set of recommended concepts given the input `concept[]`.
New permission: `"vocabulary:%s:lookup:recommended:post"`, which is assigned to roles based on source permissions.
Added i18n elements for Atlas UI.
Added some debug helpers.


anton-abushkevich:
@chrisknoll 
There are failed tests, please check 

1. For integration tests: it seems the temporal table "temp_migration" cannot be created since it is already creating in migration V2.8.0.20201022120031__concept_ancestor_and_descendants.sql 
2. org.ohdsi.webapi.service.VocabularyServiceTest - SQL comparison failure

src/main/java/org/ohdsi/webapi/service/VocabularyService.java
```diff
@@ -1051,6 +1131,56 @@ protected PreparedStatementRenderer prepareGetDescendantConceptsByList(String[]
...
+        public Void mapRow(ResultSet resultSet, int i) throws SQLException {
+          addRecommended(concepts, resultSet);
+          return null;
+        }
+      });
+
+      return concepts.values();
+
+    } catch (Exception e) {
+      if (e.getCause().getMessage().contains("concept_recommended") && e.getCause().getMessage().contains("does not exist")) {
```
anton-abushkevich, line 1157: In case of localized Postgres there will be no "does not exist" substring in the message. On my system I receive code 500 with "org.springframework.jdbc.BadSqlGrammarException" in the error description. The check needs to be less specific. I think `e.getCause().getMessage().contains("concept_recommended")` should be enough.

chrisknoll: Ok, thanks I didn't account for different locales. I'll address that.

chrisknoll:
I've made the update to drop the temp table after migration, so that should address that (thanks for fidning that) and we only look for the `concept_recommended` table when trying to determine if the table doesn't exist.

PR #2095, opened 2022-10-03
Adding REST endpoint javadoc comments for estimation and prediction endpoints

anthonysena:
@ablack3


anthonysena:
I'd also like to note that the EstimationController has an `@MireDotIgnore` tag since the estimation import object is very large and causes a problem for Miredot. We may want to investigate this further in a separate issue but for now I think this PR is ready.

PR #2103, opened 2022-10-04
Issue 1203 documentation updates

anthonysena:
Documentation updates for deprecated services, cohort analysis and concept set endpoints.


src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -510,6 +675,16 @@ public void unassignPermissionProtectedTag(@PathParam("id") final int id, @PathP
         unassignTag(id, tagId);
     }
 
+    /**
+     * Checks a concept set for diagnostic problems. At this time, 
+     * this appears to be an endpoint used to check to see which tags
+     * are applied to a concept set?
```
chrisknoll, line 681: Would not leave this question in the documentation.

chrisknoll (changes requested):

PR #2126, opened 2022-10-11
Issue 2019 hydra upgrade

ssuvorov-fls:


src/main/resources/resources/estimation/r/runAnalysis.R
```diff
@@ -56,7 +56,8 @@ tryCatch({
                 runAnalyses = TRUE,
                 packageResults = TRUE,
                 maxCores = maxCores,
-                minCellCount = 5)
+                minCellCount = 5,
+                verifyDependencies = FALSE)
```
alex-odysseus, line 60: Please add a comment what is the reason behind

pom.xml
```diff
@@ -22,7 +22,7 @@
     <bouncycastle.version>1.68</bouncycastle.version>
     <shiro.version>1.8.0</shiro.version>
     <dom4j.version>2.1.3</dom4j.version>
-    <hydra.version>0.2.0</hydra.version>
+    <hydra.version>0.3.0</hydra.version>
     <featureExtraction.version>3.2.0</featureExtraction.version>
     <tomcat.embed.version>8.5.81</tomcat.embed.version>
```
alex-odysseus, line 27: Merge with the remote rc-2.12.0 which already contains changes

alex-odysseus (changes requested):

src/main/resources/resources/estimation/r/runAnalysis.R
ssuvorov-fls: We can't verify dependencies because r-env.lock file if absent after installation of package into R repository

alex-odysseus (approved):

PR #2133, opened 2022-10-12
set default value for source.is_cache_enabled column to `false`

### LINKED ISSUE #2131: Null value in column "is_cache_enabled" of relation "source" violates not-null constraint due to missing default value
### Expected behavior

Initializing the WebAPI `source` table [by following the Wiki](https://github.com/OHDSI/WebAPI/wiki/CDM-Configuration#source-and-source_daimon-table-setup) instructions or using existing SQL scripts that used to work in version 2.11 should also work in WebAPI 2.12.0.

### Actual behavior

https://github.com/OHDSI/WebAPI/pull/2077 added a boolean `is_cache_enabled` column which is not-null but doesn't specify a default value. This causes `INSERT`s that don't specify a value for `is_cache_enabled` to fail.

I believe this might be best mitigated by adding a new migration that sets the default value for `is_cache_enabled` to `true`, like so:

```sql
ALTER TABLE ${ohdsiSchema}.source ALTER COLUMN is_cache_enabled SET DEFAULT true;
```

Which should preserve the pre version 2.12 caching behavior.

I can take a stab at a PR if the suggestion makes sense!

1. create an empty PostgreSQL database
2. launch the WebAPI
3. follow the Wiki instructions for setting up a source: https://github.com/OHDSI/WebAPI/wiki/CDM-Configuration#source-and-source_daimon-table-setup

chgl:
Closes #2131 

Two questions:

1. should the migration be versioned as v2.12.1 or v2.13.0?
1. out-of-scope, but while building the container image I noticed the `openjdk:8-jre-slim` base image is deprecated: <https://hub.docker.com/_/openjdk>
    > This image is officially deprecated and all users are recommended to find and use suitable replacements ASAP. 

    I switched to `eclipse-temurin:8-jre` and it seems to run fine. Since this PR is just one line so far, should I update the image as well while I'm at it?

### DIFF
--- a/src/main/resources/db/migration/postgresql/V2.12.1.202210120000__alter_source_set_cache_enabled_default.sql
+++ b/src/main/resources/db/migration/postgresql/V2.12.1.202210120000__alter_source_set_cache_enabled_default.sql
@@ -1,0 +1,1 @@
+ALTER TABLE ${ohdsiSchema}.source ALTER COLUMN is_cache_enabled SET DEFAULT false;


chrisknoll:
We'd like to focus PRs on one specific problem at a time. Since we're trying to resolve #2131, let's limit our scope to that.

There is cases where we have PRs solving multiple issues, but usually there's a common theme (like documentation, typos, etc).  In this case, it's a caching setting one one hand and a docker setting on the other. Let's stick with the former, and open an issue about the docker.

chrisknoll:
The migration version of 2.12.1 is fine.

chgl:
Sure, I'll keep the PR as-is then. Thanks for the feedback!

Issue #2112, opened 2022-10-05
Heracles reports do not insert final results when using Spark


TomWhite-MedStar:
### Expected behavior
After defining and  generating a cohort, it should be possible to use the Reporting tab to generate Quick or Full Analyses.

### Actual behavior
This works fine on SQL Server, but not on Spark (e.g. DataBricks). 

### Steps to reproduce behavior
Need an instance of OHDSI that has an OMOP data source connected via SPARK
1. Select any cohort
2. Run Generate on the cohort
3. Run "Quick Analysis" under the Reporting tab
After several minutes, the Job will show a Failed status.

### Root Cause
After all of the initial processing, the final query looks like  this:

`insert into result_schema.heracles_results (cohort_definition_id, analysis_id, stratum_1, stratum_2, stratum_3, stratum_4, count_value) select cohort_definition_id, analysis_id, cast(stratum_1 as STRING), cast(stratum_2 as STRING),
 cast(stratum_3 as STRING), cast(stratum_4 as STRING), count_value
from tmp.xrfc9243results_1
UNION ALL
...`

However, SPARK SQL does not support INSERT INTO with a list of fields that does not exactly match the full set of columns for the final table.

Modifying the query does work.  Either this:

`insert into omop_160101_to_220731_v813_results.heracles_results (cohort_definition_id, analysis_id, stratum_1, stratum_2, stratum_3, stratum_4, stratum_5, count_value, last_update_time)
select cohort_definition_id, analysis_id, cast(stratum_1 as STRING), cast(stratum_2 as STRING),
 cast(stratum_3 as STRING), cast(stratum_4 as STRING), '' as stratum_5, count_value, now()
from tmp.xrfc9243results_1`

or this:

`insert into omop_160101_to_220731_v813_results.heracles_results
select cohort_definition_id, analysis_id, cast(stratum_1 as STRING), cast(stratum_2 as STRING),
 cast(stratum_3 as STRING), cast(stratum_4 as STRING), '' as stratum_5, count_value, now()
from tmp.xrfc9243results_1`

Can the relevant SQL fragments be modified so that this works for Spark?

One of the relevant files appears to be:
https://github.com/OHDSI/WebAPI/blob/master/src/main/resources/resources/cohortanalysis/sql/selectHeraclesResults.sql

But it looks as though a Java file is used to orchestrate the sub-queries:
[...]

PR #2138, opened 2022-10-17
Fix issue #2112 - Heracles reports do not insert final results when using Spark

TomWhite-MedStar:


src/main/resources/resources/cohortanalysis/sql/selectHeraclesDistResults.sql
```diff
@@ -1,6 +1,6 @@
 select cohort_definition_id, analysis_id,
   cast(stratum_1 as varchar(255)), cast(stratum_2 as varchar(255)), cast(stratum_3 as varchar(255)), cast(stratum_4 as varchar(255)), cast(stratum_5 as varchar(255)),
   cast(count_value as bigint), cast(min_value as float), cast(max_value as float), cast(avg_value as float), cast(stdev_value as float),
-  cast(median_value as float), cast(p10_value as float), cast(p25_value as float), cast(p75_value as float), cast(p90_value as float)
+  cast(median_value as float), cast(p10_value as float), cast(p25_value as float), cast(p75_value as float), cast(p90_value as float), GETDATE()
 from #results_dist_@analysisId
```
chrisknoll, line 5: I was uner the impression that we would want to name each of these columns in the result as well, ie:  cast(stratum_1 as varchar(255)) as stratum_1.  I should look where we are using this select statement to pull the result.

In addition, if this is the query that is fetching the results, then I don't think we should be using GETDATE() here, shouldn't the result be the date stored in the row?   I haven't looked at these reports in quite a while, but I'm actually surprised that we store dates on each of the results:  these reports were based on Achilles, and we didn't do it there.

TomWhite-MedStar: @chrisknoll, I tried doing the insert without GETDATE(), and Spark threw an error about mismatched  # of columns, even though last_update_time has a default value of GETDATE() in the DDL.

Although I can change the select queries to include the column name, they did not start that way.  If you want me to make that change I can updated the pull request.

chrisknoll: Hi, I'm revisiting this now to understand some things:   the lack of column names form the casts is strange, but I need to see how it is being used in context to know for sure.   My gut says that when you select a cast() value the column won't be named, and that's going to be confusing.

I need to find out where the `selectHeraclesDistResults` is being referenced in code, and then understand how the columns are being accessed.

chrisknoll: Hi, I won't have time to dig deeply into this, so I will approve this with the assumption that this change works for you and other dialects.

chrisknoll (approved):

chrisknoll:
Hello, could someone confirm if this PR will work as intended? We can merge it in and get it into a hotfix.

PR #2145, opened 2022-11-01
Add tagging management through UI - Atlas#2773

### LINKED ISSUE #2149: Tagging Management API
To support Tagging Management UI (https://github.com/OHDSI/Atlas/issues/2773)

anton-abushkevich:
Resolves #2149
Server side for OHDSI/Atlas#2773


src/main/java/org/ohdsi/webapi/tag/TagService.java
```diff
@@ -68,7 +65,7 @@ public Tag create(Tag tag) {
                 .filter(Tag::isAllowCustom)
                 .count() == groups.size();
 
-        if (!allowCustom) {
+        if (!this.getPermissionService().isSecurityEnabled() && !SecurityUtils.getSubject().isPermitted("tag:management") && !allowCustom) {
```
anton-abushkevich, line 68: The logic here became that if security is disabled, than user cannot create tags in a group that does not allow custom tags - even in "Tag management" section. I assume what you wanted to add here is - if security is disabled, than we do not need to check the permission. In this case it should be `if (this.getPermissionService().isSecurityEnabled() && ...` - without `!`.

chrisknoll: Ok, you're right, oddly, this change still let me edit tags so there is a logical problem with isSecurityEnabled or elsewhere. I'll find it and push it up.

chrisknoll: I changed it to as follows:

```
if (this.getPermissionService().isSecurityEnabled() && !SecurityUtils.getSubject().isPermitted("tag:management") && !allowCustom)
```

However, in order for the exception to be thrown, all 3 things must be true: 
- security enabled
- not permitted to tag:management
- The group does not allows custom

But this means that you could be permitted but the group does not allow custom, and you won't throw the exception, so should the logic be:

```
securityEnabled AND (not permitted to tag:management OR not allows custom)
```

anton-abushkevich: I think, the logic is ok, and these 3 things indeed must be true to fail tag creation. Lets see:

1. security enabled = false. Obviously, if it is false we allow to create tags in any group. In this case, other checks do not relevant, so skip the exception.

If security enabled = true, then lets check other expressions:

2. if tag:management not permitted - if it is false, then we do not have any objection for user to create tags in any group, regardless of the `allowCustom` flag. But if it is true and user does not have this permission, then:

3. Last check - if the group allows custom tags? No - throw exception. Otherwise, all checks passsed.

chrisknoll: Ok, thank you, I wasn't sure if allowCustom would allow you to make something if tag:management wasn't granted.  Since you can still manage the tag wihtout permissions if allowCustom is enabled, then your initial logic was correct. Thanks.

chrisknoll (approved):

PR #2176, opened 2022-12-13
Add csv for comparison conceptsets

anton-abushkevich:
Backend for OHDSI/Atlas#2800


chrisknoll:
I tried to compare with the attached concept set CSV and got this error:

```
Caused by: org.postgresql.util.PSQLException: ERROR: stack depth limit exceeded
  Hint: Increase the configuration parameter "max_stack_depth" (currently 2048kB), after ensuring the platform's stack depth limit is adequate.
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2675)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2365)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:355)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:490)

```

Looks like this error isn't a java stack limit, but rather something in PGAdmin.

[concept_compare_1.csv](https://github.com/OHDSI/WebAPI/files/10304537/concept_compare_1.csv)

chrisknoll:
Just as an implementation note: I think this is a case where it may make more sense to do this in java-space:   by sorting the list and scanning through each result, you can determine which elements are absent/present from each side by scanning through the lists (something not easily done on sql side), but a sql solution could be done using a combinatino of joins (inner join will find 'both', left join will find in C1, and right join could find those in C2.

src/main/resources/resources/vocabulary/sql/getConceptByCodeAndVocabulary.sql
```diff
@@ -0,0 +1 @@
+select concept_id from @vocabulary_database_schema.CONCEPT where (concept_code, vocabulary_id) in (@conceptColumns)
```
anthonysena, line 1: From discussion on Atlas WG, this query could be problematic and the root cause of the issue reported in https://github.com/OHDSI/WebAPI/pull/2176#issuecomment-1365358488. This may require refactoring to use a temp table that is created using a `UNION ALL` with all of the concepts. Tagging @chrisknoll as he may be able to provide a code pointer to other places in the code base where this is done.

src/main/java/org/ohdsi/webapi/service/cscompare/ConceptSetCompareService.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        };
+
+
+        return executeInTransaction(source, callbackFunction);
+    }
+
+    private String getQuery(final ConceptSetExpression csExpression, final ExpressionType type,
+                            final Source source, final JdbcTemplate jdbcTemplate) {
+        final ConceptSetExpressionQueryBuilder builder = new ConceptSetExpressionQueryBuilder();
```
anton-abushkevich, line 87: Move the instantiation of the builder into 'else' block.

anton-abushkevich:
@chrisknoll, The fix is ready - please review.

chrisknoll (approved):

PR #2177, opened 2022-12-14
Add concept sets search

anton-abushkevich:
Backend for OHDSI/Atlas#2803

Solr core configuration for CS Search was added to `src/main/resources/solr/concept-sets`.
Solr core config for Concepts Search was moved from `src/main/resources/solr` to `src/main/resources/solr/vocabulary`.

### DIFF
--- a/src/main/resources/i18n/messages_en.json
+++ b/src/main/resources/i18n/messages_en.json
@@ -2465,7 +2465,7 @@
       "importUsers": "Import Users from LDAP/AD",
       "managePermissions": "Manage Permissions",
       "reindexCS": "Concept Sets Reindex",
-      "reindexCSStatus": "Concept Sets Reindex (<%=doneCount%> of <%=maxCount%>)"
+      "reindexCSStatus": "Concept Sets Reindex (<%=doneCount%> of <%=maxCount%>)",
       "tagManagement": "Tag Management"
     },
     "changeSourcePriorities": "Change source priorities in:",


chrisknoll (changes requested):
This is my first pass:

This PR is extremely large, touching 122 files,, and ~9k lines to review.   The initial issue (OHDSI/Atlas#2803) is very light on functional details about what the functionality is doing.

We had initial meetings about what the functionality was, and my understanding was at the time we would limit this type of search to find concepts by what was specified in the literal concept set, and not what may be referenced in an included concepts or mapped (source) concepts.   We also didn't include solr in the original discussions.

What is in this PR is much larger, more complex, and potentially extremely resource heavy on the database servers in order to perform re-indexing.   I have concerns about how this is actually going to perform, but at this point I have only given this the high-level review and have not executed the implementation locally.   The challenges to this is 1) a reasonable environment to do proper workload evaluation, and 2) local dev environment does not have SOLR enabled so I will need to enabled that locally in order to use this function.

I would ask that the initial issue (OHDSI/Atlas#2803) be amended to include more functional details about how this feature works:
- How is it invoked
- How much additional resources does supporting this feature put on a system?
- There is references to index refresh, but who does it?  Is it done every time an asset is saved, or some administrator need to kick off the job?
- How is it disabled?

That's a good start....more to come as this function is reviewed more...

ssuvorov-fls:
some changes already exist in https://github.com/OHDSI/WebAPI/pull/2196

anthonysena:
Moving to v2.14 per https://github.com/OHDSI/Atlas/issues/2803#issuecomment-1429777622

anthonysena:
Closing in favor of #2315

PR #2178, opened 2022-12-14
Add chart drilldown to concept set pages

anton-abushkevich:
Backend for OHDSI/Atlas#2805


chrisknoll (changes requested):
This function says it adds 'explore timeseries' to concept sets, but I do not see it when looking at a concept set inside a cohort definition (or any of the other 'embedded' concept set types.   I think you should be able to add this tab to the concept set editor that is presented for entities that have embedded concept sets.  Is this possible?

chrisknoll:
As I described in the UI PR (https://github.com/OHDSI/Atlas/pull/2806), I suggest removing the reports at the 'conceptset' level, but keeping the 'concept' drilldowns.   I believe that means that we don't need any WebAPI changes at all to show concept drilldowns in the UI.   I think most of this PR is invalidated because the use of SUM() and AVG() to combine concept statistics is incorrect/improper.

ssuvorov-fls:
@chrisknoll 
Hi
Should I close this PR?

chrisknoll:
If there are no requirements in this PR for https://github.com/OHDSI/Atlas/issues/2805, then yes, you can close.

PR #2179, opened 2022-12-16
Venn diagram

TitrS:
added locales to venn diagram


src/main/resources/i18n/messages_en.json
```diff
@@ -1704,7 +1704,9 @@
         "saveFromComparisonNameTail": " - From Comparison",
         "saveMessage": "Save New Concept Set From Selection Below",
         "saveWarning": "You must save the current concept set before you can perform this comparison.",
-        "viewCountMessage": "View database record counts (RC) and descendant record counts (DRC) for:"
+        "vennDiagram": "Venn Diagram",
+        "viewCountMessage": "View database record counts (RC) and descendant record counts (DRC) for:",
```
anton-abushkevich, line 1708: Comma at the end makes the JSON invalid. Remove it.

anton-abushkevich (changes requested):

src/main/resources/i18n/messages_en.json
```diff
@@ -1704,7 +1704,9 @@
         "saveFromComparisonNameTail": " - From Comparison",
         "saveMessage": "Save New Concept Set From Selection Below",
         "saveWarning": "You must save the current concept set before you can perform this comparison.",
+        "vennDiagram": "Venn Diagram",
         "viewCountMessage": "View database record counts (RC) and descendant record counts (DRC) for:"
+
```
anton-abushkevich, line 1709: Delete empty line. Line count in all localization file should be equal.

anton-abushkevich (changes requested):

anton-abushkevich (approved):

src/main/resources/i18n/messages_en.json
chrisknoll: Btw: while it's a good check to ensure localization lines should be equal across files, I noticed that the order of keys is out of order across files. This makes it difficult to use something like WinMerge to find differences in files, where we could find out that a key exists in one file and is missing in another (but this would only work if the keys were in the same order).

Just something to think about for a future update.

PR #2188, opened 2023-01-24
Dependencies update 2.13.0

alex-odysseus:


chrisknoll:
Can circe 1.10.1 be part of this PR?  https://github.com/OHDSI/circe-be/releases/tag/v1.10.1

alex-odysseus:
Chris, I asked Sergey to have a separate pull request for that and then cherrypick it to 'master-2.12'
If it is easier we can extend this one and do the same as this initial commit would be nice to cherrypick to 'master-2.12' as well
@chrisknoll @ssuvorov-fls

chrisknoll:
If this update is going to be also cherrypicked into 2.12.1, then let's put the circe update here as well.  If you need to have a separate commit so that you want to cherry pick circe-be back to 2.12.1, then yes, you'll want a separate PR. 

Thanks for the clarification.

chrisknoll (approved):

PR #1091, opened 2019-05-15
updates for solr based search

fdefalco:

PR #2196, opened 2023-01-30
solr functionalty was moved to sparate location

ssuvorov-fls:


src/main/java/org/ohdsi/webapi/extcommon/ConfigurationInfo.java
```diff
@@ -1,4 +1,4 @@
-package org.ohdsi.webapi.info;
+package org.ohdsi.webapi.extcommon;
```
chrisknoll, line 1: Don't move classes to new package in a minor release.   '`info` represents the namespace for classes that convey information about webapi to clients.  `extcommon` sounds like some sort of 'shared extension', which is very different, and implies some sort of functionality that isn't described.  Before making these types of architectural changes, they should be discussed and understood.

chrisknoll (changes requested):

anthonysena:
@ssuvorov-fls - we need an issue files for this PR to understand the motivation for these changes. In discussion with @alex-odysseus and @chrisknoll around this particular issue, there is a need to externalize the dependencies, interfaces, etc with the SOLR functionality. Can you please describe these in an issue and then we can link the issue to this PR? This is required otherwise we don't know how to evaluate the implementation.

anthonysena:
Linking to this PR: https://github.com/OHDSI/WebAPI/pull/1091 which is relevant to the changes that introduced SOLR functionality to the code base and also includes issues/discussions that are relevant to this set of changes too.

src/main/java/org/ohdsi/webapi/extcommon/vocabulary/SearchProvider.java
```diff
@@ -0,0 +1,12 @@
+package org.ohdsi.webapi.extcommon.vocabulary;
+
+import org.ohdsi.webapi.vocabulary.Concept;
+import org.ohdsi.webapi.vocabulary.SearchProviderConfig;
+
+import java.util.Collection;
+
+public interface SearchProvider {
+    boolean supports(String vocabularyVersionKey);
+    int getPriority();
```
chrisknoll, line 10: `getPriority` seems out of place here:   priority is a something in the context of Source and SourceDaimon.   Why would a SearchProvider have a priority?

src/main/java/org/ohdsi/webapi/extcommon/ConfigurationInfo.java
ssuvorov-fls: I've reverted moving classes to new package

ssuvorov-fls:
@chrisknoll I've restored moved classes
next steps for 2.14:

1. Creating two repositories:
1.1. Repository for common classes which will be used in both webapi and solr modules
This repository will include the following classes:
org.ohdsi.webapi.vocabulary.Concept;
org.ohdsi.webapi.vocabulary.SearchProviderConfig;
org.ohdsi.webapi.vocabulary.SearchProvider;
org.ohdsi.webapi.info.ConfigurationInfo - this classes can be excluded but in this case we won't be able to get information about Solr config parameters
1.2. Repository for solr modules
org.ohdsi.webapi.vocabulary.solr.SolrSearchProvider;
org.ohdsi.webapi.vocabulary.solr.SolrSearchClient;
org.ohdsi.webapi.vocabulary.solr.SolrConfigurationInfo;

2. Adding profile to pom.xml in WebAPI

src/main/java/org/ohdsi/webapi/vocabulary/SearchProvider.java
```diff
@@ -1,12 +1,12 @@
-package org.ohdsi.webapi.extcommon.vocabulary;
+package org.ohdsi.webapi.vocabulary;
 
 import org.ohdsi.webapi.vocabulary.Concept;
 import org.ohdsi.webapi.vocabulary.SearchProviderConfig;
 
 import java.util.Collection;
 
-public interface SearchProvider {
+public interface SearchProvider<T extends SearchProviderConfig> {
```
chrisknoll: Please justify the use of a parameterized type here.

ssuvorov-fls: It was removed in my last commit and there's no parameterized type now

ssuvorov-fls: Very strange, because this commit was made before your comment

chrisknoll: No problem, I don't know why the files looked different than your latest commit, but thank you for letting me know.

chrisknoll:
> This repository will include the following classes:
> org.ohdsi.webapi.vocabulary.Concept;
> 

Perhaps the Concept class should be moved to StandardAnalysisAPI since that repo is designed to store analytic classes,

Note: I don't mean create an interface Concept (so that everyone who wants to handle Concept data they have to create their own ConceptImpl class...)  The StandardAnalysisAPI can contain POJOs that represent data without behavior.

chrisknoll:
My understanding of this PR is to prepare the codebase to be easier to 'extract' the SOLR implementation into a separate library such that WebAPI can be built without SOLR libraries if it isn't required.

I'm seeing at least two places that changes were made but I do not see how it applies to this function, and wanted to get clarity:

1.  The supports() method of SearchProvider changed from VocabulyarsearchProviderType to String (a vocabularyVersionKey).   This seems to be adding a new feature where you can ask the search provider if it supports a vocabulary version?  Was it doing something like this before?  Is the change that there will be multiple SOLR cores installed with different vocabulary versions and getSearchProvider() will find the cores for that given vocabulary version?   What's strange to me about this is that now people may have to pay attention more to vocabulary versions of their cores, and I don't think they had to do that in the past: you just put a core for SOLR to use and whatever version is there it would query.
2. Related to the above, there's a hew function of 'getPriority().  Based on the above, it would seem that priority would be used to order the cores by priority, but looking at the implementation of getPriority(), it returns a constant `SOLR_PRIORITY`.  So, is this notion of a 'priority' something new to this PR?

Additional thoughts on externalizing the search providers:

We should probably say away from using WebAPI-specific terms from the provider and interface, specifically anything related to `Source`.  Source is something that exists only in the WebAPI domain, and to 'leak' Source-specific contexts will lead to unnecessary coupling between WebAPI and these external services.

ssuvorov-fls:
1. It is not functionality. It is taken from `VocabularySearchServiceImpl.getSearchProvider`. The Solr search can't be used when there is no index for version key
2. `GetPriority` is used to get appropriate search provider. Solr has higher priority and will be used instead of vocabulry search when Solr module is enabled

chrisknoll (approved):

ssuvorov-fls:
I removed it from code as I think that when we include solr starter then we need to get exception in case of lack of configuration.
But it seems that you are right and we need to check availability now
I'll create another pr with these changes

PR #2202, opened 2023-02-08
Bump up several libraries

ecruz165:
Removed the following unused libraries:
- org.ehcache:ehcache 3.8.1
- javax.cache:cache-api 1.0.0
- commons-httpclient:commons-httpclient 3.1

Upgraded the following to versions provided by SPring Boot 2.7 
-- org.apache.commons:commons-lang3 3.9 -> 3.12.0
-- com.google.code.gson:gson 2.8.9 -> 2.9.0
-- com.zaxxer:HikariCP 3.3.1 -> 4.0.3
-- com.microsoft.sqlserver:mssql-jdbc 9.2.1.jre8 -> 10.2.1.jre8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -705,15 +705,14 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-cache</artifactId>
     </dependency>
-    <dependency>
-      <groupId>javax.cache</groupId>
-      <artifactId>cache-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.ehcache</groupId>
-      <artifactId>ehcache</artifactId>
-      <version>3.8.1</version>
-    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>javax.cache</groupId>-->
+<!--      <artifactId>cache-api</artifactId>-->
+<!--    </dependency>-->
+<!--    <dependency>-->
+<!--      <groupId>org.ehcache</groupId>-->
+<!--      <artifactId>ehcache</artifactId>-->
+<!--    </dependency>-->
     <dependency>
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-core</artifactId>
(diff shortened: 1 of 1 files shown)


chrisknoll:
Our Spring Boot version seems to be 1.5 (at least according to this line in the pon):
```
<spring.boot.version>1.5.22.RELEASE</spring.boot.version>
```

Will these updates cause a conflict with the current version of spring boot?

ecruz165:
@chrisknoll I am aware we are on spring boot 1.5. Maybe the title of PR caused unnecessary concern.

As you might know, spring bundles third-party version numbers of libraries usually paired with spring boot applications to provide features. So I identified the libraries listed in WebAPI that can safely be bumped so that when we arrive at the goal of being on Spring Boot 2.7.x and then we can remove the explicit versioning and let the listed versions listed in Spring Boot handle the upgrading of libraries instead of bumping libs individually.
https://docs.spring.io/spring-boot/docs/2.7.0/reference/html/dependency-versions.html#appendix.dependency-versions

So yes, it is safe to update these libraries> I ran the unit test and the supplied integration tests and all ran to success.

chrisknoll:
Excellent! Thank you, pushing forward the dependencies to current versions is an important task.

chrisknoll (changes requested):
(deleted)

pom.xml
```diff
@@ -705,15 +705,14 @@
...
-      <version>3.8.1</version>
-    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>javax.cache</groupId>-->
+<!--      <artifactId>cache-api</artifactId>-->
+<!--    </dependency>-->
+<!--    <dependency>-->
+<!--      <groupId>org.ehcache</groupId>-->
+<!--      <artifactId>ehcache</artifactId>-->
+<!--    </dependency>-->
```
chrisknoll, line 715: Can we remove these lines?

chrisknoll (changes requested):

pom.xml
ecruz165: removed.  I did notice a manual approach to caching was implemented. Any reason why we chose manually cache implementation over Ehcache? I am very experienced in implementing Ehcache in case there is a future need.

anthonysena:
These changes are related to upgrading WebAPI to Spring Boot 2.x

anthonysena (approved):

PR #2209, opened 2023-02-15
Switch isNumeric() to anayisID test when casting.

### LINKED ISSUE #2187: Drilldown Reports cast issue
### Expected behavior
When Achilles has been executed with benchmark records, these records have decimal values stored in the stratum_1 fields of achilles_results and achilles_results_dist. These should not cause a casting issue for the [drilldown report SQLs](https://github.com/OHDSI/WebAPI/tree/master/src/main/resources/resources/cdmresults/sql/report/condition/drilldown).

### Actual behavior
In Redshift, we see this cause errors. As the drilldown queries utilize a casting of isnumeric, if the value is a decimal value, it will be numeric but not castable to int.

```
ERROR: Invalid digit, Value '.', Pos 2, Type: Integer 
  Detail: 
  -----------------------------------------------
  error:  Invalid digit, Value '.', Pos 2, Type: Integer 
  code:      1207
  context:   28.98
  query:     3749295
  location:  :0
  process:   query1_241_3749295 [pid=0]
  -----------------------------------------------
```

Run Achilles with benchmarks enabled. Try examining the drilldown reports in Data Sources. Observe no results, and then see the SQL errors in the WebAPI logs.

chrisknoll:
Fixes #2187.

These changes span a large number of files.   I did not change any files related to heracles reports (these should probably be deprecated).

The general strategy is moving away from isNumeric() and instead depending on the record's `analysis_id` to determine how the casting should perform.   The problem with isNumeric() is that numeric values like '123.45' were passign the isNumeric() test, but can not be casted to an integer.   By using analysis_id you can be certain that the cast you are doing is appropriate to the analysis that generated the record.

### DIFF
--- a/src/main/resources/resources/cdmresults/sql/report/condition/drilldown/ageAtFirstOccurrence.sql
+++ b/src/main/resources/resources/cdmresults/sql/report/condition/drilldown/ageAtFirstOccurrence.sql
@@ -11,9 +11,9 @@ SELECT
 FROM @results_database_schema.achilles_results_dist ard1
 INNER JOIN
 @vocab_database_schema.concept c1
-ON CAST(CASE WHEN isNumeric(ard1.stratum_1) = 1 THEN ard1.stratum_1 ELSE null END AS INT) = c1.concept_id
+ON CAST(CASE WHEN ard1.analysis_id = 406 THEN ard1.stratum_1 ELSE null END AS INT) = c1.concept_id
 INNER JOIN
 @vocab_database_schema.concept c2
-ON CAST(CASE WHEN isNumeric(ard1.stratum_2) = 1 THEN ard1.stratum_2 ELSE null END AS INT) = c2.concept_id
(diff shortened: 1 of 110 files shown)


chrisknoll:
@alondhe , this PR is in draft so that you can test this branch in your own env. If it works for you, we'll make it a formal PR.

alondhe:
Thanks @chrisknoll -- will review here and deploy in our test environment by early next week.

chrisknoll:
@alondhe, any results?

alondhe (approved):
Sorry for the delay, @chrisknoll! Been fighting with our tech stack and OS permissions. 

I tested by re-inserting our benchmark records into achilles_results and achilles_results_dist. In Atlas 2.12.1, the drilldowns throughout each domain would error out with the Redshift error I described. 

Using this branch, I tried the same tests and never came across the Redshift errors. I also checked that the values from the Achilles tables match what's in the report.

PR #2215, opened 2023-02-23
Refactored imports to depend on org.apache.commons:commons-lang3

ecruz165:
I saw we are using in code references to commons-lang and commons-lang3. I removed dependency on commons-lang, removed any unused imports in files i updated. And I introduced commons-text to support usage of WordUtils that was moved to that library.


anthonysena:
Discussion on this from the Atlas WG - we'll revisit this in the next release when the SOLR work is completed since there is a merge conflict on that file at the moment.

anthonysena:

PR #2217, opened 2023-02-23
Set job status to failed when exception is thrown in listener.

### LINKED ISSUE #2216: CohortGeneration job left in 'STARTED' state when error is thrown.
### Expected behavior
If there is a failure during cohort generation, the job should result in FAILED.

### Actual behavior
The exception thrown in GenerationJobExecutionListener is ignored by Spring Batch.

This is difficult to reproduce because you need to throw an exception during `GenerationJobExecutionListener.findBySourceId`.
However, the behavior is described [here](https://stackoverflow.com/questions/48210366/can-an-exception-in-a-spring-batch-listener-cause-the-job-to-fail).

The solution is to wrap the work done in afterJob (which contains the call to findBySourceId) in an exception handler such that the job step is set to failed:

Example code:
```
public class YourListener implements StepExecutionListener {

    @Override
    public ExitStatus afterStep(StepExecution stepExecution) {
        try {
            yourcode ///////
        } catch (Exception e) {

            //log your error 
            stepExecution.setStatus(BatchStatus.FAILED);
            stepExecution.addFailureException(e);
            return ExistStatus.FAILED.addDescription(e);
        }
        return ExitStatus.COMPLETED;
    }
}
```

chrisknoll:
Fixes #2216.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
+++ b/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerationJobExecutionListener.java
@@ -40,98 +40,117 @@ import java.util.Map;
 import java.util.Objects;
 
 import static org.ohdsi.webapi.Constants.Params.TARGET_DATABASE_SCHEMA;
+import org.springframework.batch.core.ExitStatus;
 
 /**
  * @author cknoll1
  */
 public class GenerationJobExecutionListener implements JobExecutionListener {
-    private static final Logger log = LoggerFactory.getLogger(GenerationJobExecutionListener.class);
-    private final SourceService sourceService;
-    private final CohortDefinitionRepository cohortDefinitionRepository;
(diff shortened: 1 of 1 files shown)


alex-odysseus:
Chris, I don't understand why GitHub doesn't show changes, anything wrong with formatting?

chrisknoll:
Shows it for me:
image

alex-odysseus:
Not on this tab I usually watch https://github.com/OHDSI/WebAPI/pull/2217/files

chrisknoll:
From the link you provided, I see the change:
image

chrisknoll:
Is the problem that you don't know about the 'hide whitespace' option?

image

chrisknoll:
I'd prefer it if we could get review on the material changes on this PR.

ssuvorov-fls (approved):

PR #2218, opened 2023-02-23
Hypersistence Optimizer Issues: Blocker: PooledSequenceOptimizerEvent

ecruz165:


chrisknoll:
Did you want to fold multiple changes to address Hypersistence Optimizer concerns into a single PR? I'm not sure if we're going to be fixing dozens or hundreds of messages (leading to the same number of PRs). Did you want to have this PR for issues of a certain severity (ie: critical and high?)

ecruz165:
This was only one marked blocker. CLose out this PR. It is a same and easy one. 

The report pointed out 165 issues were found: 1 BLOCKER, 80 CRITICAL, 40 MAJOR, 44 MINOR. I'll review the critical and group a PR by type of issue raised.

chrisknoll:
Ok, that's fine if this one covers all the 'blocker' category, we'll just leave the PR as-is and we'll merge it in as soon as we're done with our 2.13 release (which should be soon).

anthonysena (approved):

alex-odysseus:
@ecruz165 @chrisknoll @anthonysena @wivern @anton-abushkevich @ssuvorov-fls 

Don't you find it risky to have the algorithm changed? Would it be possible that a new sequence generator algorithm will generate an identifier which already exists?

https://github.com/OHDSI/WebAPI/commit/1a114fa9d9798d93035abb2c7a0fe16a17082455 (I read that from Hibernate 5 the default value is "true" but we set it explicitly to "false")

chrisknoll:
We marked it `false` in that commit because it messed with the 'implicit' auto generated columns in hibernate that we were using at the time.

Since then, we went to an explicit `sequence generator`, and so this option doesn't apply to us anymore.

PR #2219, opened 2023-02-24
Refactored imports to depend on com.google.guava:guava

ecruz165:
Currently, project has the child dependencies jersey-guava-2.25.1.jar and guava-28.2-jre.jar. Classes found in jersey-guava were moved to dedicated guava library.


anthonysena:
From discussion on the Atlas WG, this change is aimed at upgrading Spring Boot 2.x

anthonysena (approved):

PR #2220, opened 2023-02-24
Hypersistence Optimizer Issues: Critical: EagerFetchingEvent

ecruz165:
Consider using lazy fetching which, not only that is more efficient, but is way more flexible when it comes to fetching data. For more info about this event, check out this User Guide link - https://vladmihalcea.com/hypersistence-optimizer/docs/user-guide/#EagerFetchingEvent

I left one reported below as is because it seem eager is necessary here to avoid no session error i encountered.
The [daimons] attribute in the [org.ohdsi.webapi.source.Source] 
This article points out alternatives and i concluded keeping eager is right thing to do in this case.
https://www.baeldung.com/hibernate-initialize-proxy-exception

Reran report
BEFORE
165 issues were found: 1 BLOCKER, 80 CRITICAL, 40 MAJOR, 44 MINOR
AFTER
137 issues were found: 1 BLOCKER, 52 CRITICAL, 40 MAJOR, 44 MINOR

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortcharacterization/domain/CohortCharacterizationEntity.java
+++ b/src/main/java/org/ohdsi/webapi/cohortcharacterization/domain/CohortCharacterizationEntity.java
@@ -70,7 +70,7 @@ public class CohortCharacterizationEntity extends CommonEntityExt<Long> implemen
     @Column(name = "strata_only")
     private Boolean strataOnly;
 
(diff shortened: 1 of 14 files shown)


chrisknoll:
Eager also prevents N+1 fetches which can be very bad for performance.   I'm fairly certain where we use EAGER it is intentional and necessary, but I'll have to review the specific cases that you uncover.

anthonysena:

PR #2224, opened 2023-03-10
Fix SQL in procedure drilldown report

gebilaoman:
Here are two grammatical errors, the source of which was updated on February 16, 2023

### DIFF
--- a/src/main/resources/resources/cdmresults/sql/report/procedure/drilldownsummary/prevalenceByGenderAgeYear.sql
+++ b/src/main/resources/resources/cdmresults/sql/report/procedure/drilldownsummary/prevalenceByGenderAgeYear.sql
@@ -1,13 +1,13 @@
 SELECT
   c1.concept_id                                                          AS concept_id,
   c1.concept_name                                                        AS concept_name,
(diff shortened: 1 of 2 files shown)


chrisknoll (approved):

gebilaoman:
> 

yeah, I think you are right,thank you

gebilaoman:
> Thank you for finding this issue. I think we could accept this PR off of your master but I would like to caution you about working this way. Instead you should think about branching off master (even in your own fork) to fix changes. So, the workflow is:
> 
> 1. Identify the issue and make a issue out of (note the number of the issue here)
> 2. In your fork, create a branch named `issue-xxxx` or `issue-xxx-{short text about bug}`
> 3. Commit your changes into your branch, and create a PR that referenes your fork's branch with fix.
> 
> This may sound nit-picking, but it will help you out in the future if we are taking multiple PRs into the upstream `master` branch, your local master branch will not have those commits and I think it would get more complicated than if you just isolate your changes to a side branch.
> 
> @anton-abushkevich , I understand you guys are finalizing the release for 2.13. This is a bug that was introduced when fixing the casting problem based on numeric. I just malformed the query in the case of the procedure drilldown. Can this be merged in for 2.13 or did you want to get 2.13 done and fix this in a hotfix?

yeah, I think you are right,thank you

Issue #2230, opened 2023-03-16
webapi-docker profile still uses non-TLS url


alondhe:
### Expected behavior
As discussed here: #1825 we should use TLS based URLs in the pom.

### Actual behavior
This is not the case in the webapi-docker profile:

https://github.com/OHDSI/WebAPI/blob/master/pom.xml#L1345

### Steps to reproduce behavior
As we lack the port access for 8085, this causes Docker builds to fail in our site.

PR #2231, opened 2023-03-16
Changed non-TLS nexus to use TLS for docker

alondhe:
In docker-webapi profile, edited the nexus url to use TLS, same as in other profiles.

Issue #2230

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1342,7 +1342,7 @@
         <repository>
           <id>ohdsi</id>
           <name>repo.ohdsi.org</name>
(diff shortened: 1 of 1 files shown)


alondhe:
> Do you know why we re-declare the repositories node in the docker profile when I think it matches the general repository setting in the [main pom](https://github.com/OHDSI/WebAPI/blob/fe1920b1d4173a4e5cd4e3c023b9c53aeeba3ed5/pom.xml#L519)?

Ah good question. I'll test removing it so we don't need the separate entry here.

alondhe:
Withdrawing -- will revise.

PR #2235, opened 2023-03-27
Webapi docker pom tls

alondhe:
Issue #2230 

In this PR version, I've removed the extra maven and ohdsi repo entries for the webapi-docker profile. 

However, I have added a repo entry for ohdsi snapshots. This ensures that all snapshots from the ohdsi repo are handled with TLS enabled, and doesn't revert to using snapshots on the non-TLS sites.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1335,15 +1335,17 @@
           lower(email) = lower(?)</security.db.datasource.authenticationQuery>
       </properties>
       <repositories>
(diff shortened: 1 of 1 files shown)


anthonysena:
Tagging @chgl and @leeevans for any feedback on this PR.

leeevans (approved):
approved

Issue #2210, opened 2023-02-16
Column 'webapi.tag.created_by_id' is invalid in the select list because it is not contained in either an aggregate function


dsfrederic:
### Expected behavior

No errors in startup

### Actual behavior

Following error is popping up in the logs:
```
2023-02-16T20:17:48.825078196Z 2023-02-16 20:14:17.771 ERROR taskScheduler-1 org.hibernate.engine.jdbc.spi.SqlExceptionHelper - [] - Column 'webapi.tag.created_by_id' is invalid in the select list because it is not contained in either an aggregate function or the GROUP BY clause.
```

### Steps to reproduce behavior

WEBAPI_RELEASE=2.10.1

Context: I'm running my database on SQL server (on azure) for which I've translated the migration scripts from Postgresql to SQL server myself. There were some actions related to the webapi.tag table. But still this does not seem like something I have impact on.

chrisknoll:
Fixed in #2242.

Issue #2155, opened 2022-11-08
Application startup failed


dsfrederic:
👋🏻  I'm trying the setup of this webapi for the first time. I'm doing this on a Azure Container app my data source is an Azure SQL Server. 

I've deployed my container app with the following params

```
  container {
    name   = "broadsea-webtools"
    image  = "ohdsi/webapi:latest"
    cpu    = "0.5"
    memory = "1.5"

    ports {
      port     = 8080
      protocol = "TCP"
    }

    environment_variables = {
      DATASOURCE_DRIVERCLASSNAME = "com.microsoft.sqlserver.jdbc.SQLServerDriver",
      DATASOURCE_URL = #REDACTED#,
      DATASOURCE_USERNAME = #REDACTED#,
      DATASOURCE_PASSWORD = #REDACTED#
      DATASOURCE_OHDSI_SCHEMA =  "ohdsi",
      SPRING_JPA_PROPERTIES_HIBERNATE_DIALECT= "org.hibernate.dialect.SQLServer2012Dialect",
      SPRING_JPA_PROPERTIES_HIBERNATE_DEFAULT_SCHEMA= "ohdsi",
      SPRING_BATCH_REPOSITORY_TABLEPREFIX= "ohdsi.BATCH_",
      FLYWAY_DATASOURCE_DRIVERCLASSNAME= "com.microsoft.sqlserver.jdbc.SQLServerDriver",
      FLYWAY_DATASOURCE_URL=  #REDACTED#,
      FLYWAY_DATASOURCE_USERNAME = #REDACTED#,
      FLYWAY_DATASOURCE_PASSWORD = #REDACTED#
      FLYWAY_LOCATIONS= "classpath:db/migration/sqlserver",
      FLYWAY_PLACEHOLDERS_OHDSISCHEMA= "ohdsi",
      FLYWAY_SCHEMAS= "ohdsi",
      SECURITY_CORS_ENABLED= "true",
      SECURITY_ORIGIN = "*"
    }
  }
```

This gives me errors like 

```
org.springframework.context.ApplicationContextException: Unable to start embedded container; nested exception is org.springframework.boot.context.embedded.EmbeddedServletContainerException: Unable to start embedded Tomcat
```

```
Caused by: org.springframework.boot.context.embedded.EmbeddedServletContainerException: Unable to start embedded Tomcat
```

```
[...]

PR #2238, opened 2023-03-29
Translated postgresql migration scripts to SQL Server up to v2.10

dsfrederic:


chrisknoll (changes requested):

src/main/resources/db/migration/sqlserver/V2.9.0.20210423125133__assets_tags.sql
```diff
@@ -0,0 +1,172 @@
...
+    color                VARCHAR(MAX)                  NULL,
+    multi_selection      BIT                     NOT NULL DEFAULT 0,
+    permission_protected BIT                     NOT NULL DEFAULT 0,
+    mandatory            BIT                     NOT NULL DEFAULT 0,
+    allow_custom         BIT                     NOT NULL DEFAULT 0,
+    description          VARCHAR(MAX)                  NULL,
+    created_by_id        INTEGER,
+    created_date         DATETIME NOT NULL DEFAULT SYSDATETIMEOFFSET(),
+    modified_by_id       INTEGER,
+    modified_date        TIMESTAMP,
```
chrisknoll, line 98: I don't think modified date should be different datatime than created date. I believe in sqlserver, we made all date fields 'datetime'.

chrisknoll (approved):

chrisknoll:
Ok, can you confirm that these latest changes work in your environment, and i will proceed with merging the PR.

Side note:  in the future (not for this PR) you may want to consider creating a branch off of your fork's master branch. I'm not sure what it will look like when we merge this commit into OHDSI/WebAPI master branch and how that will impact your own local fork's master branch.  If you work off of branches, then there's no discrepancy between the commits that will be merged into OHDSI master vs. your fork's master (which will come from syncing with upstream).

dsfrederic:
Well, I'm experiencing some other issues but it does not seem related. 

https://github.com/microsoft/OHDSIonAzure/issues/111
https://github.com/microsoft/OHDSIonAzure/issues/89
https://github.com/OHDSI/WebAPI/issues/2155
https://github.com/OHDSI/WebAPI/issues/2210
https://forums.ohdsi.org/t/failing-webapi-atlas-implementation/18617/8

chrisknoll:
Ok, one thing we found was that Hibernate is yielding a query that is OK on PG, but not OK on MSSQL:   [this code](https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/tag/repository/TagRepository.java#L23) is defining a select/count/group by that is not doing the correct group-by clause.   On one hand, we only support PG, on the other, I think it's important to yield valid SQL (PG may break later).  So I think we should find a fix for this, but I don't think we're going to back-fix this back to the 2.10 release (since we're up to 2.13 now)

chrisknoll:
I created an issue that will address this, but it's targeted for v2.14 (it's not exactly a hotfix, we're changing a method type, but in a way that shouldn't incur backwards-breaking change.   If you are stuck on v2.10, did you want to try to make a local change on your local env to try to work around this?   Note: this is a PG vs. MSSql issue which is why we decided to limit ourselves to 1 platform. Another way to work around this is host WebAPI on PG while you can still host your patient-level-data on your Azure platform...although I think the context you are working on is an all-microsoft solution (the microsoft/ohdsiOnAzure repo) but maybe the pure microsoft solution can still include a PG component?

PR #2245, opened 2023-04-04
Filter cohorts and concepts by permission

rkboyce:


src/main/java/org/ohdsi/webapi/security/PermissionController.java
```diff
@@ -92,17 +92,22 @@ public List<RoleDTO> listAccessesForEntity(@QueryParam("roleSearch") String role
...
-    @Path("/access/{entityType}/{entityId}")
+    @Path("/access/{entityType}/{entityId}/{role}")
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     public List<RoleDTO> listAccessesForEntity(
             @PathParam("entityType") EntityType entityType,
-            @PathParam("entityId") Integer entityId
+            @PathParam("entityId") Integer entityId,
+	    @PathParam("role") String role
     ) throws Exception {
```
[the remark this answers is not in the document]
chrisknoll: Also, there's some overloading of terms 'role' where in this context, it is 'READ', 'WRITE' for 'role' but in other places we have 'Admin' as a role and 'Atlas User' as a role...but the latter roles maps to a shiro 'group' (which you assign permissions to'....we could probably interchange groups and roles, however, the AccessType is neither a group or a role...so...I don't have a solution here, just pointing out the differences in context.

It's almost like we're talking about a 'permission type'  .. permission to read, permission to write.

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -103,6 +104,9 @@ public class ConceptSetService extends AbstractDaoService implements HasTags<Int
     @Autowired
     private VersionService<ConceptSetVersion> versionService;
 
+    @Value("#{'${security.conceptsetsauthview}'.equals('True')}")
```
chrisknoll, line 107: we should look at places where we have 'enable' flags set, I'm not sure we look for the string 'True', but I could be wrong. let's find other examples and try to follow precedent.

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -131,15 +135,28 @@ public ConceptSetDTO getConceptSet(@PathParam("id") final int id) {
...
-        return getTransactionTemplate().execute(transactionStatus ->
-                StreamSupport.stream(getConceptSetRepository().findAll().spliterator(), false)
-                        .map(conceptSet -> {
-                            ConceptSetDTO dto = conversionService.convert(conceptSet, ConceptSetDTO.class);
-                            permissionService.fillWriteAccess(conceptSet, dto);
-                            return dto;
-                        })
-                        .collect(Collectors.toList())
-        );
+   	       if (conceptsetsauthview == false) { // don't filter based on read permissions
```
chrisknoll, line 138: Is this something we want on a case-by-case basis (ie: conceptsetauthview can be false but cohortdefinitionauthview can be set to true)?  Part of me thinks all-or-nothing makese sense and is easier to configure (ie: works the same everywhere) vs you do want to restrict one set of assets to only return those with perms vs. another asset you would let anyone see everything.

chrisknoll:
@rkboyce ,  I noticed when we did the walkthrough and we enabled the filter on the server, the asset list returned nothing until I created a new asset and it got the read perm assigned to it. I wanted to ask: have we thought about considering 'read access' if you have write? This way, if we enable it, anything I've been granted write access will appear, plus anything that someone grants me read access to.   Basically the filter becomes `if (canRead || canWrite|  then show item`.  Is that possible or was there a reason why we had to explicitly look for the granted permission?  The only reason why I bring this is that it seems if you can write to something then you can also read it...

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
rkboyce: This was changed to check for boolean false on a parameter that is more akin to others in the settings.xml

```  @Value("#{'${security.defaultglobalreadpermissions}'.equals(false)}")```

File: src/main/java/org/ohdsi/webapi/service/ConceptSetService.java (hunk 3 above)
Comment (rkboyce) [reply src/main/java/org/ohdsi/webapi/service/ConceptSetService.java]: The more general parameterized value is now defaultglobalreadpermissions - so, all or nothing

File: src/main/java/org/ohdsi/webapi/service/ConceptSetService.java [hunk 4]
```diff
@@ -131,15 +135,28 @@ public ConceptSetDTO getConceptSet(@PathParam("id") final int id) {
...
+		    return getTransactionTemplate().execute(
+					transactionStatus -> StreamSupport.stream(getConceptSetRepository().findAll().spliterator(), false)
+					                .filter(candidateConceptSet -> permissionService.hasReadAccess(candidateConceptSet))
+							.map(conceptSet -> {
+								ConceptSetDTO dto = conversionService.convert(conceptSet, ConceptSetDTO.class);
+								permissionService.fillWriteAccess(conceptSet, dto);
+								permissionService.fillReadAccess(conceptSet, dto);
+								return dto;
+							})
+					                .collect(Collectors.toList()));
```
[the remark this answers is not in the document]
rkboyce: Sorry about not implementing this way. It is a more elegant way to do things. It's my inexperience with reviewing comments on draft pull requests that caused me to miss this suggestion. I had carried through the less elegant coding approach throughout all of the other cases (IR, estimation, prediction, etc.) before seeing this.

src/main/java/org/ohdsi/webapi/security/PermissionController.java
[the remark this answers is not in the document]
rkboyce: I addressed this the overloading of the 'role' string by using 'permType' as the parameter. The function name is now listAccessesForEntityByPermType  used for the /access/{entityType}/{entityId}/{permType} endpoint with WRITE as the default permission type sent from the client. This distinguishes it from listAccessesForEntity  which is used for the /access/suggest endpoint.  I did not get a chance to test if the AccessType enum would work instead of string "READ" and "WRITE".

rkboyce:
In response to the last comment about "have we thought about considering 'read access' if you have write" -- yes, it is implemented that way and seems to function correctly based on my tests. 

This updated draft pull request implements the READ permissions filtering across all the following applications in Atlas:
- concept sets
- Cohort definition
- Cohort characterization
- Incident Rates
- Estimation
- Prediction

All have functioned as expected : things are not accessible to a user unless 1) they create the entity (which sets READ and WRITE), or 2) they are given READ permission to the entity

*NOTE:* These changes will need to come with a new system role that the admins could select which I am calling the 'Read Restricted Atlas User' role. I think the following query could be the basis for adding SQL to flyway to create this role: 

```
select distinct sp.*
from ohdsi.sec_role_permission srp 
  inner join ohdsi.sec_permission sp on srp.permission_id = sp.id 
where srp.role_id in (6,10) -- 'cohort reader', 'Atlas Users'
 and sp.value not in 
    (  
    	'conceptset:*:get',
    	'conceptset:*:expression:get',
    	'conceptset:*:version:*:expression:get',    	       
    	--
        'cohortdefinition:*:get',
        'cohortdefinition:*:info:get',
        'cohortdefinition:*:version:get',
        'cohortdefinition:*:version:*:get',        
        --        
	        'cohort-characterization:*:get',
                'cohort-characterization:*:generation:get',
		'cohort-characterization:generation:*:get',
		'cohort-characterization:design:get',
		'cohort-characterization:*:design:get',
		'cohort-characterization:design:*:get',
		'cohort-characterization:*:version:get',
		'cohort-characterization:*:version:*:get',
		--
		'pathway-analysis:*:get',
		'pathway-analysis:*:generation:get',
		'pathway-analysis:generation:*:get',
		'pathway-analysis:generation:*:result:get',
		'pathway-analysis:generation:*:design:get',
		'pathway-analysis:*:version:get',
		'pathway-analysis:*:version:*:get'
		--
		'ir:*:get',
		'ir:*:copy:get',
		'ir:*:info:get',
		'ir:*:design:get',
		'ir:*:version:get',
		'ir:*:version:*:get'
		--		
		'estimation:*:get',
		'estimation:*:copy:get',
		'estimation:*:download:get',
		'estimation:*:export:get',
		'estimation:*:generation:get'
               'comparativecohortanalysis:*:get',
		--
		'prediction:*:get',
		'prediction:*:copy:get',
		'prediction:*:download:get',
		'prediction:*:export:get',
		'prediction:*:generation:get',
		'prediction:*:exists:get',
                'plp:*:get'
    )
order by sp.value 
;
```

pom.xml
```diff
@@ -192,6 +192,11 @@
     <spring.datasource.hikari.register-mbeans>true</spring.datasource.hikari.register-mbeans>
     <spring.datasource.hikari.mbean-name>authDataSource</spring.datasource.hikari.mbean-name>
 
+    <!-- If defaultGlobalReadPermissions is set to true (default), then all users can see every artifact.  -->
+    <!-- If it is set to false, WebAPI will filter out the artifacts that a user does not explicitly have -->
+    <!-- read permissions to -->
+    <security.defaultGlobalReadPermissions>false</security.defaultGlobalReadPermissions>
```
chrisknoll, line 198: We should default this to true, and rely on settings.xml to override this setting. This will maintain current behavior between versions.

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
```diff
@@ -86,6 +89,9 @@ public class CcController {
     private CharacterizationChecker checker;
     private PermissionService permissionService;
 
+    @Value("#{'${security.defaultglobalreadpermissions}'.equals(false)}")
```
chrisknoll, line 92: I believe I will want this to be camelCase based on other params: defaultGlobalReadPermissions,   Also we should check if we can interpret this as a boolean value directly instead of a string compare.

src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
```diff
@@ -70,7 +71,10 @@ public class EstimationController {
   private final ScriptExecutionService executionService;
   private EstimationChecker checker;
   private PermissionService permissionService;
-
+  
+  @Value("#{'${security.defaultglobalreadpermissions}'.equals(false)}")
+  private boolean defaultglobalreadpermissions;
```
chrisknoll, line 76: Style: defaultGlobalReadPermissions

src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
```diff
@@ -97,14 +101,26 @@ public EstimationController(EstimationService service,
...
+      return StreamSupport.stream(service.getAnalysisList().spliterator(), false)
+	.filter(candidateEstimation -> permissionService.hasReadAccess(candidateEstimation))
+	.map(analysis -> {
+	    EstimationShortDTO dto = conversionService.convert(analysis, EstimationShortDTO.class);
+	    permissionService.fillWriteAccess(analysis, dto);
+	    permissionService.fillReadAccess(analysis, dto);
+	    return dto;
+	  })
+	.collect(Collectors.toList());
+    }
```
chrisknoll, line 123: I believe this was mentioned before but you can put the filter conditional in the stream expression:

```
      return StreamSupport.stream(service.getAnalysisList().spliterator(), false)
	.filter(!defaultglobalreadpermissions  ? candidateEstimation -> permissionService.hasReadAccess(candidateEstimation) : candidateEstimation -> true)
	.map(analysis -> {
	    EstimationShortDTO dto = conversionService.convert(analysis, EstimationShortDTO.class);
	    permissionService.fillWriteAccess(analysis, dto);
	    permissionService.fillReadAccess(analysis, dto);
	    return dto;
	  })
```

src/main/java/org/ohdsi/webapi/pathway/PathwayController.java
```diff
@@ -156,13 +162,29 @@ public PathwayAnalysisDTO importAnalysis(final PathwayAnalysisExportDTO dto) {
...
+			for (PathwayAnalysisEntity pa : newpage) {
+				if (permissionService.hasReadAccess(pa)) {
+					PathwayAnalysisDTO dto = conversionService.convert(pa, PathwayAnalysisDTO.class);
+					permissionService.fillWriteAccess(pa, dto);
+					permissionService.fillReadAccess(pa, dto);
+					dtolist.add(dto);
+				}
+			}
+			return new PageImpl<PathwayAnalysisDTO>(dtolist, pageable, dtolist.size());
+		}
```
chrisknoll, line 185: I believe this can be written same as in Estimation:

```
	return pathwayService.getPage(pageable)
		.filter(!defaultglobalreadpermissions  ? pa -> permissionService.hasReadAccess(pa) : pa -> true)
		.map(pa -> {
			PathwayAnalysisDTO dto = conversionService.convert(pa, PathwayAnalysisDTO.class);
			permissionService.fillWriteAccess(pa, dto);
			permissionService.fillReadAccess(pa, dto);
			return dto;
		});

```

src/main/java/org/ohdsi/webapi/prediction/PredictionController.java
```diff
@@ -93,26 +97,40 @@ public PredictionController(PredictionService service,
...
+					.filter(candidateAnalysis -> permissionService.hasReadAccess(candidateAnalysis))
+					.map(analysis -> {
+						CommonAnalysisDTO dto = conversionService.convert(analysis, CommonAnalysisDTO.class);
+						permissionService.fillWriteAccess(analysis, dto);
+						permissionService.fillReadAccess(analysis, dto);
+						return dto;
+					})
+					.collect(Collectors.toList());
+		}
+	}
```
chrisknoll, line 123: Same comment as above: use conditional filter to reduce redundancy.

src/main/java/org/ohdsi/webapi/prediction/PredictionController.java
```diff
@@ -93,26 +97,40 @@ public PredictionController(PredictionService service,
...
+	
+
+	/**
+	 * Check to see if a prediction design exists by name
+	 * 
+	 * @summary Prediction design exists by name
+	 * @param id The prediction design id
+	 * @param name The prediction design name
+	 * @return 1 if a prediction design with the given name and id exist in WebAPI and 0 otherwise
+	 */
```
chrisknoll, line 133: Fix indent of comments, should be aligned to the following function.

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -406,14 +410,26 @@ public GenerateSqlResult generateSql(GenerateSqlRequest request) {
...
+			return definitions.stream()
+					.filter(candidateCohortDef -> permissionService.hasReadAccess(candidateCohortDef))
+					.map(def -> {
+						CohortMetadataDTO dto = conversionService.convert(def, CohortMetadataImplDTO.class);
+						permissionService.fillWriteAccess(def, dto);
+						permissionService.fillReadAccess(def, dto);
+						return dto;
+					})
+					.collect(Collectors.toList());
+		}
```
chrisknoll, line 432: use conditional filter here.

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -131,15 +135,28 @@ public ConceptSetDTO getConceptSet(@PathParam("id") final int id) {
...
```
chrisknoll, line 158: use conditional filter here.

src/main/java/org/ohdsi/webapi/service/IRAnalysisService.java
```diff
@@ -341,17 +345,32 @@ private String getStrataTreemapData(int analysisId, int targetId, int outcomeId,
...
+	  return StreamSupport.stream(analysisList.spliterator(), false)
+	    .filter(candidateIRAnalysis -> permissionService.hasReadAccess(candidateIRAnalysis))
+	    .map(analysis -> {
+                IRAnalysisShortDTO dto = conversionService.convert(analysis, IRAnalysisShortDTO.class);
+                permissionService.fillWriteAccess(analysis, dto);
+		permissionService.fillReadAccess(analysis, dto);
+                return dto;
+              })
+	    .collect(Collectors.toList());
+	});
```
chrisknoll, line 372: use conditional filter here, will simplify code so that you only have the one `getTransactionTemplarte()` call, but within the stream you filter conditionally before mapping.

src/main/resources/application.properties
```diff
@@ -107,6 +107,7 @@ csrf.disable=true
 
 sparql.endpoint=http://virtuoso.ohdsi.org:8890/sparql?default-graph-uri=&query=
 
+security.defaultglobalreadpermissions=${security.defaultGlobalReadPermissions}
```
chrisknoll, line 110: Follow camelCase style, for example `security.cas.callbackUrl`.  So, set this to `security.defaultGlobalReadPermissions`

chrisknoll (changes requested):

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
rkboyce: I completed this change. ty

pom.xml
rkboyce: corrected now and tested that setting the variable false in the WebAPIConfig/settings.xml works

src/main/resources/application.properties
rkboyce: corrected

src/main/java/org/ohdsi/webapi/estimation/EstimationController.java
rkboyce: corrected

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcController.java
```diff
@@ -151,11 +157,28 @@ public CohortCharacterizationDTO copy(@PathParam("id") final Long id) {
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)
     public Page<CcShortDTO> list(@Pagination Pageable pageable) {
-        return service.getPage(pageable).map(entity -> {
-            CcShortDTO dto = convertCcToShortDto(entity);
-            permissionService.fillWriteAccess(entity, dto);
-            return dto;
-        });
+			if (defaultglobalreadpermissions == true) { // don't filter based on read permissions
```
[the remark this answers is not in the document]
rkboyce: corrected

rkboyce:
A detail worth mentioning  - were were curious what would happen if a person without read access knew the URL to an entity and tried a GET. I tested as follows:

1) logged in as an admin, created a brand new entity (did it for a few types e.g., cohort definition, estimation, etc)

2) copied the URL

3) logged out from the admin account and logged in to a user account without access to the entity

4) put the URL in the browser

The result is an empty page with no message at all. The server simply returns a 403 with an empty body in the response. Atlas renders nothing. So, I think it is Ok for now.

chrisknoll:
Yes, that's fine. Ok, if you're done with your part, I will merge this into a new feature branch under WebAPI so I can put in the final changes. Let me know if you are ready for me to do that.

rkboyce:
All good here!

PR #2254, opened 2023-04-14
OpenID Connect custom parameters

RowanErasmus:
Hi!

I would like to be able to pass the 'domain_hint' parameter with oidc authorization for 'a slightly more streamlined user experience' (https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow) with Azure Active Directory.

When using it it would look something like this "security.oid.customParams={\"domain_hint\":\"your.org\"}"

Would be happy to hear what you think :-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,7 @@
     <security.oid.url></security.oid.url>
     <security.oid.logoutUrl></security.oid.logoutUrl>
     <security.oid.extraScopes></security.oid.extraScopes>
+    <security.oid.customParams>{:}</security.oid.customParams>
     <security.oid.redirectUrl>http://localhost/index.html#/welcome/</security.oid.redirectUrl>
     <security.kerberos.spn></security.kerberos.spn>
     <security.kerberos.keytabPath></security.kerberos.keytabPath>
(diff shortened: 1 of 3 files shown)


chrisknoll:
We should add documentation to [security configuration](https://github.com/OHDSI/WebAPI/wiki/Security-Configuration)  to indicate the new OpenID config param `customParams` that takes a comma delim tuple `key1:value1,key2:value2`.

For Azure AD, the use case is domain_hint:your_org.

chrisknoll (approved):

chrisknoll:
We just received a report on the forums indicating a 'circular placeholder reference 'security.oid.customPrams in property definitions', and I'm wondering if there's some sort of cicular reference introduced in this PR.   This was when the app is run in a docker environment (IE: broadsea) and i'm not sure if there's some issue there.  I didn't see this error when I reviewed this PR, however, I also didn't run it under broadsea.

@RowanErasmus , any thoughts?

RowanErasmus:
Hi @chrisknoll this issue appeared during developing this feature once and then disappeared again, I'm going to take a closer look tomorrow. In the meanwhile you may want to revert the merge...

RowanErasmus:
I'm guessing it has to do with caching going belly up in the docker build, I've responded in the forum

Issue #2248, opened 2023-04-06
Atlas Characterizations, Cohort Pathways & Cohort Reports not working on Spark (Databricks)


TomWhite-MedStar:
### Expected behavior
Three Atlas functions that work fine on SQL Server are not working on Spark:  Characterizations, Cohort Pathways, and Cohort Reports (Heracles reports).  Characterizations worked fine in Spark/Databricks in the 2.11 release of Atlas and WebAPI, but do not work in 2.12.1

### Actual behavior
Each of those types of analyses fail - usually after significant runtime.  In each case the root cause appears to be SqlRender error in substituting <START_WITH> and <END_WITH>.  See[ issue # 330 ](https://github.com/OHDSI/SqlRender/issues/330 )in SqlRender for more detail.  There is also discussion on  [this ](https://forums.ohdsi.org/t/atlas-characterizations-error-org-ohdsi-webapi-exception/18639 )forum post.

#### Characterizations
Load the attached json into Chacterizations.  Select Executions and click Generate
[Lipitor Users vs Zocor Users.json.txt](https://github.com/OHDSI/WebAPI/files/11172664/Lipitor.Users.vs.Zocor.Users.json.txt)
There are multiple queries like `show columns in #analysis_ref` where the temporary table names are not substituted to something like `show columns in tmp.analysis_ref`.  However, those queries do not cause the job to fail.
Instead, it fails the first time there is a unprocessed <START_WITH> substitution.  The error file is below.
[Lipitor Users vs Zocor Users.error.txt](https://github.com/OHDSI/WebAPI/files/11172706/Lipitor.Users.vs.Zocor.Users.error.txt)

#### Cohort Pathways
Load the attached json into Cohort Pathways.  Select Executions, and click Generate
[Diabetic Treatment Pathway.json.txt](https://github.com/OHDSI/WebAPI/files/11172360/Diabetic.Treatment.Pathway.json.txt)
The error message generated is attached.
[Diabetic Treatment Pathway.error.txt](https://github.com/OHDSI/WebAPI/files/11172596/Diabetic.Treatment.Pathway.error.txt)

#### Cohort / Heracles Reports.
Load the attached json cohort.  Generate the cohort.  Then, under Cohort Reporting, select that source and choose Quick Analysis
[...]

PR #2257, opened 2023-04-25
Fix cdmresults drill-down reports

TomWhite-MedStar:
Three files had minor typos breaking drill-down report components.
Also fixes issue #2248 in Heracles Full Reports where analysis # 200 is concatenated to the next one (# 500).

### DIFF
--- a/src/main/resources/resources/cohortanalysis/heraclesanalyses/sql/200_201.sql
+++ b/src/main/resources/resources/cohortanalysis/heraclesanalyses/sql/200_201.sql
@@ -19,9 +19,8 @@ WHERE vo1.visit_start_date>=c1.cohort_start_date and vo1.visit_end_date<=c1.coho
 --}
 group by c1.cohort_definition_id,
 --{@CDM_version == '4'}?{
-vo1.place_of_service_CONCEPT_ID
+vo1.place_of_service_CONCEPT_ID;
 --}
 --{@CDM_version == '5'}?{
-vo1.visit_CONCEPT_ID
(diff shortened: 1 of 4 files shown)


chrisknoll (approved):

src/main/resources/resources/cohortanalysis/heraclesanalyses/sql/200_201.sql
```diff
@@ -19,9 +19,8 @@ WHERE vo1.visit_start_date>=c1.cohort_start_date and vo1.visit_end_date<=c1.coho
...
 group by c1.cohort_definition_id,
 --{@CDM_version == '4'}?{
-vo1.place_of_service_CONCEPT_ID
+vo1.place_of_service_CONCEPT_ID;
 --}
 --{@CDM_version == '5'}?{
-vo1.visit_CONCEPT_ID
+vo1.visit_CONCEPT_ID;
 --}
-;
```
chrisknoll, line 27: Slightly odd fix here:  to move the trailing `;` from the end into each of the branches of the SqlRender if-then block...but I don't think it's a problem.

TomWhite-MedStar: @chrisknoll , this looked like the safest solution.  Of the dozens of files that contains the --{, and --} syntax, this is the only one that has --} immediately before the closing semi-colon.  I tried adding extra newlines, but the parser for Spark seems to do a greedy match for whitespace after --}, so no matter how many newlines I added, I always got --; (but only in the Spark translation)

PR #2258, opened 2023-04-25
Enable PC and DPC for source and non-standard concepts

TomWhite-MedStar:
This, in combination with a pull request # [721 ](https://github.com/OHDSI/Achilles/pull/721)for Achilles, 
It solves the issue discussed in [this](https://forums.ohdsi.org/t/patient-counts-on-non-standard-concepts/18143/6) forum post.
I confirmed that this works on my Databricks instance.

### DIFF
--- a/src/main/resources/ddl/achilles/achilles_result_concept_count.sql
+++ b/src/main/resources/ddl/achilles/achilles_result_concept_count.sql
@@ -76,18 +76,28 @@ WITH counts AS (
 ), counts_person AS (
   SELECT stratum_1 as concept_id, MAX (count_value) agg_count_value
   FROM @results_schema.achilles_results
-  WHERE analysis_id IN (200, 400, 600, 700, 800, 900, 1000, 1300, 1800, 2100, 2200)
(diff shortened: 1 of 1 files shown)


TomWhite-MedStar:
@chrisknoll , per discussion, removed Achilles analysis # 40

anthonysena:
Thanks @TomWhite-MedStar for this contribution! I'd like to wait and see the outcome of OHDSI/Achilles#721. Assuming that is approved and the analyses you've added are included, we'll be able to merge this into WebAPI.

TomWhite-MedStar:
Thanks.  @anthonysena , or @chrisknoll , Do you need any more information from me about [OHDSI/Achilles#721](https://github.com/OHDSI/Achilles/pull/721)?

anthonysena:
I think we're set on the WebAPI side @TomWhite-MedStar. I just want to make sure that the analysis IDs referenced in the WebAPI query are the ones that are finalized in Achilles. I know that we could merge this in now but in the off-chance that the Achilles folks have any desire to change analysis IDs for some reason, we'd have to patch this up.

chrisknoll:
I think it'll be a low priority that they will object to the analysis_ids so how about we accept it here with that slight risk, and if they do change the analysis_ids, we can patch.

anthonysena:
That's good by me thanks @chrisknoll

chrisknoll (approved):

PR #2264, opened 2023-05-02
Added utility methods for synchronizing defined relationsips

ecruz165:
…ationships between tables.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/source/Source.java
+++ b/src/main/java/org/ohdsi/webapi/source/Source.java
@@ -17,10 +17,7 @@ package org.ohdsi.webapi.source;
 
 import com.odysseusinc.arachne.execution_engine_common.api.v1.dto.KerberosAuthMechanism;
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.Date;
-import java.util.List;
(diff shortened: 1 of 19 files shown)


src/main/java/org/ohdsi/webapi/source/Source.java
```diff
@@ -17,10 +17,7 @@
 
 import com.odysseusinc.arachne.execution_engine_common.api.v1.dto.KerberosAuthMechanism;
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.Date;
-import java.util.List;
-import java.util.Objects;
+import java.util.*;
```
chrisknoll, line 20: Lets leave this alone for now, since it's a little inconsistent now between the other explicit imports vs this one being wildcard.

chrisknoll (changes requested):

anthonysena:
@ecruz165 - we'll need to open an issue that describes the problem and then we can review work to ensure that it fixes the problem. We believe that this pull request corresponds to a tool's scan output so if we could include that in the issue it would be helpful for traceability. Thanks!

anthonysena:

PR #2267, opened 2023-05-04
Fix heracles 200 for spark

TomWhite-MedStar:
@chrisknoll , This is further fix for #2257 

The root cause may to be a "greedy" regex operator in SqlRender for Spark; but this fix ensures that the final semicolon is not trapped behind a single-line comment.


TomWhite-MedStar:
I tested the original #2257 WebAPI patch on our local instance, and it also dropped the terminal semicolon (so blended together queries 200 and 500).  The attached query renders correctly in https://data.ohdsi.org/SqlDeveloper/

chrisknoll (approved):

chrisknoll:
Thank you. The regex replaces in SqlRender is something we need to be careful about. Thanks for finding the issue and providing a fix.

TomWhite-MedStar:
@chrisknoll and @konstjar -  I tested Full Analysis on DataBricks using this patch.  The run completes without error, and all of the generated analyses behave as expected.  Thanks!

Issue #2256, opened 2023-04-20
Achilles DDL not working on Databricks Spark


TomWhite-MedStar:
### Expected behavior
The DDL generated by the [Achilles DDL WebAPI](https://atlas-demo.ohdsi.org/WebAPI/ddl/achilles?dialect=spark&schema=MY_RESULTS&vocabSchema=OMOP) for spark should run properly.

### Actual behavior
The generated DDL includes two WITH clauses within the same SQL statement, which throws an error.

### Steps to reproduce behavior
1. Generate DDL for spark from [Achilles DDL WebAPI.](https://atlas-demo.ohdsi.org/WebAPI/ddl/achilles?dialect=spark&schema=MY_RESULTS&vocabSchema=OMOP) - attached:
[achilles_ddl-spark.sql.txt](https://github.com/OHDSI/WebAPI/files/11290647/achilles_ddl-spark.sql.txt)
2. The error is on line 99, which reads `WITH insertion_temp AS (`
3. If that line is changed to `, insertion_temp AS (`, the query runs properly.

konstjar:
Closing since the issue was addressed in https://github.com/OHDSI/WebAPI/pull/2269 and merged.

PR #2269, opened 2023-05-09
Fix issue #2256

TomWhite-MedStar:
Fix issue #2256.  Confirmed now works on Databricks.
Root cause may be issue in SqlRender for Spark.

### DIFF
--- a/src/main/resources/ddl/achilles/achilles_result_concept_count.sql
+++ b/src/main/resources/ddl/achilles/achilles_result_concept_count.sql
@@ -17,8 +17,9 @@ CREATE TABLE @results_schema.achilles_result_concept_count
 /**********************************************/
 /***** Populate record/person count table *****/
 /**********************************************/
+DROP TABLE IF EXISTS #tmp_counts;
 WITH counts AS (
-  SELECT stratum_1 concept_id, MAX (count_value) agg_count_value
+  SELECT stratum_1 AS concept_id, MAX (count_value) AS agg_count_value
   FROM @results_schema.achilles_results
   WHERE analysis_id IN (2, 4, 5, 201, 225, 301, 325, 401, 425, 501, 505, 525, 601, 625, 701, 725, 801, 825,
     826, 827, 901, 1001, 1201, 1203, 1425, 1801, 1825, 1826, 1827, 2101, 2125, 2301)
@@ -58,7 +59,7 @@ WITH counts AS (
     */
   GROUP BY stratum_1
   UNION ALL
-  SELECT stratum_2 concept_id, SUM (count_value) AS agg_count_value
+  SELECT stratum_2 AS concept_id, SUM (count_value) AS agg_count_value
   FROM @results_schema.achilles_results
   WHERE analysis_id IN (405, 605, 705, 805, 807, 1805, 1807, 2105)
     /* analyses:
@@ -73,8 +74,15 @@ WITH counts AS (
         but this subquery only gets the type or unit concept_ids, i.e., stratum_2
     */
   GROUP BY stratum_2
-), counts_person AS (
-  SELECT stratum_1 as concept_id, MAX (count_value) agg_count_value
+)
+SELECT concept_id,
+  agg_count_value
+INTO #tmp_counts
+FROM counts;
+
+DROP TABLE IF EXISTS #tmp_counts_person;
+WITH counts_person AS (
+  SELECT stratum_1 AS concept_id, MAX (count_value) AS agg_count_value
   FROM @results_schema.achilles_results
   WHERE analysis_id IN (200, 240, 400, 440, 540, 600, 640, 700, 740, 800, 840, 900, 1000, 1300, 1340, 1800, 1840, 2100, 2140, 2200)
     /* analyses:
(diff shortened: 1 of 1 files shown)


src/main/resources/ddl/achilles/achilles_result_concept_count.sql
```diff
@@ -112,7 +112,10 @@ WITH counts AS (
   ) c
   left join @vocab_schema.concept_ancestor ca on c.concept_id = cast(ca.ancestor_concept_id as varchar(50))
 )
-INSERT INTO @results_schema.achilles_result_concept_count (concept_id, record_count, descendant_record_count, person_count, descendant_person_count)
+INSERT INTO @results_schema.achilles_result_concept_count 
+/* Fix bug in SqlRender for Spark.  When list variables for insert as below, incorrect SQL is generated
```
anthonysena, line 116: @TomWhite-MedStar - just a note here: we'd prefer not to remove the columns from the `INSERT` portion of the query. That said, we'll need to dig into why this translation is not working in Spark as it may be a bug in SqlRender.

chrisknoll: Agree with Sena, we should be explicit about col names because the order of values form the select may not be the inferred order of the insert.

TomWhite-MedStar: @chrisknoll and @anthonysena , I tested the as-is code on SqlRender 1.15.0 and it still has the problem shown in issue #2256.

chrisknoll: Ok, let me figure out who can focus on this issue on the SqlRender side, and see if we can get a 1.15.1 release so that your PR can reference it and we can have this all working.....I hate that it will delay this, but I think that's the proper way to approach it.

chrisknoll: @TomWhite-MedStar , I'm not able to find the issue in OHDSI/SqlRender so I think if we're going to get traction on this, we need a simple test case that demonstrates the multiple CTEs in the insert that get split into the double `WITH` clause (instead of with cteA as ( ) ... , cteB as ()...)

I'm sorry to bother you but could you provide a simple SQL statement with the two CTE expressions that gets translated over to 2 WITH statements?  We can feed this over into the SqlRender repo with steps to reproduce and can probably get addressed faster.  If you give me the problematic OHDSI-SQL statement that has problems translation, I will write the sample R script that will make a call to SqlRender to show the problem.

chrisknoll: @TomWhite-MedStar I'm going to grab your achilles DDL link you put in your issue as the test case, and possibly simplify it to make it easier to diagnose on the SqlRender side.   If I can replicate it in R then I can submit it as an issue over to SqlRender.

TomWhite-MedStar: @chrisknoll , excellent point.  I  refactored the query so that it runs fine in SqlRender for Spark.

src/main/resources/ddl/achilles/achilles_result_concept_count.sql
```diff
@@ -100,28 +108,44 @@ WITH counts AS (
...
+FROM #tmp_concepts concepts
+         LEFT JOIN #tmp_counts c1 ON concepts.ancestor_id = c1.concept_id
+         LEFT JOIN #tmp_counts c2 ON concepts.descendant_id = c2.concept_id
+         LEFT JOIN #tmp_counts_person c3 ON concepts.ancestor_id = c3.concept_id
+         LEFT JOIN #tmp_counts_person c4 ON concepts.descendant_id = c4.concept_id
 GROUP BY concepts.ancestor_id;
+
+DROP TABLE IF EXISTS #tmp_counts;
+DROP TABLE IF EXISTS #tmp_counts_person;
+DROP TABLE IF EXISTS #tmp_concepts;
```
chrisknoll, line 151: The IF EXISTS may not be valid syntax on all our supported platforms....and I'm not sure it's necessary since we just created it earlier.   And, if you have a typo in the table name, then it's not going to drop it because the typo-d table name won't actually exist and it won't drop the actual table you want. So, I suggest removing `IF EXISTS`.

chrisknoll (changes requested):

src/main/resources/ddl/achilles/achilles_result_concept_count.sql
TomWhite-MedStar: @chrisknoll , I found when  I ran this repeatedly, if failed is the tables already existed.
I fixed the syntax so that it matches what Sql Server expects and confirmed that it looks correct for all databases in SQLRender Developer.

chrisknoll: Ok, sounds good, it's odd that you would see tables exist when these should only be visible in the session (connection) that created them, or in cases where we use 'fake' temp tables via temp schemas, we prefix temp tables with a unique session id (meaning maybe the unique session ID isn't unique across connections, which would be odd).

In any case, the modification is more compatible, so thank you for that change.

chrisknoll (approved):

PR #2276, opened 2023-05-23
Added params to savePaths for spark dialect.

### LINKED ISSUE #2275: Non-fatal error in SQL construction for Pathway analysis on Spark (e.g. Databricks)
### Expected behavior
No generated queries throw errors

### Actual behavior
This query sent to Spark (e.g. Databricks):  `show columns in @target_database_schema.pathway_analysis_paths` without substituting `@target_database_schema`.
Note that it is a non-fatal error -- the Pathway analyses complete.  However, the generated SQL does not include the specified list of columns, which may make the process brittle.

If Atlas is pointing to a DataBricks SQL Warehouse, it logs all queries.  This query occurs immediately before an insert into pathway_analysis_paths.  However, sometimes the failed query is followed by a successful query.

The `show columns` command appears to be generated by the getMetaFields() function in [BigQuerySparkTranslate.java](https://github.com/OHDSI/SqlRender/blob/62616b12ca227d35f28e09d3b648f49db5051bdc/java/org/ohdsi/sql/BigQuerySparkTranslate.java#L2), which is called from sparkInsert().  getMetaFields() is called within a try/catch statement, so does not cause Pathway Analysis to fail when it fails.  Somehow the target_table parameter is not substituted sometimes.

chrisknoll:
Fixes #2275.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
+++ b/src/main/java/org/ohdsi/webapi/pathway/PathwayStatisticsTasklet.java
@@ -31,6 +31,7 @@ import java.util.concurrent.FutureTask;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
+import org.ohdsi.sql.SqlRender;
 import org.ohdsi.webapi.common.generation.CancelableTasklet;
 import org.ohdsi.webapi.util.PreparedStatementRendererCreator;
 import org.springframework.jdbc.core.PreparedStatementCreator;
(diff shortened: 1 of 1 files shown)


chrisknoll:
@tomwhite , @TomWhite-MedStar , please let me know if this PR addresses your issue. I believe the issue was that the SqlRender params weren't being set for the schema and that led to the issue.

TomWhite-MedStar:
@chrisknoll , Thanks.  Yes, I tested this patch and it solved the issue for Pathway analysis.

Oddly, I'm still seeing issues with SqlRender (and show columns) for other modules (esp. for Feature Extraction in Characterization).  However,  the patch does solve the issue with Pathway analysis.

chrisknoll:
Ok, thanks, please let us know where those other issues are, they probably follow a similar pattern of:

```
if (sparkDb) then { handleSparkInsert (sql) }
else {  prepareRenderTanslate(sql, params, values) }
```

We need to find those places that are doing the special spark handling and ensure their sql statements have their parameters filled.

PR #2277, opened 2023-05-24
Wildcard is not used now by default. User has to set it explicitly if needed

### LINKED ISSUE #1790: Whitespace wildcards in AD group search
### Expected behavior
When searching for groups in the Import LDAP function, search results return.

### Actual behavior
Due to the usage of WhitespaceWildcardsFilter, the search string has wildcards on both sides, which causes the search to time out in larger AD systems.

https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java#L31

Connect to AD, import LDAP groups, click on a role and try to search for an AD group using a part of the full group name.

I worked around this by making a modified version of the WhitespaceWildcardsFilter class that just adds the wildcard at the end. I do think perhaps this field shouldn't add wildcards, rather, let the user (which is most likely an admin) explicitly state the search string (with or without wildcards).

ssuvorov-fls:
fixes #1790

### DIFF
--- a/src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java
+++ b/src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java
@@ -8,11 +8,14 @@ import org.springframework.ldap.core.AttributesMapper;
 import org.springframework.ldap.core.CollectingNameClassPairCallbackHandler;
 import org.springframework.ldap.core.LdapTemplate;
 import org.springframework.ldap.filter.AndFilter;
-import org.springframework.ldap.filter.WhitespaceWildcardsFilter;
+import org.springframework.ldap.filter.EqualsFilter;
+import org.springframework.ldap.support.LdapEncoder;
 import org.springframework.ldap.support.LdapUtils;
 
 import java.util.List;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static org.ohdsi.webapi.user.importer.providers.OhdsiLdapUtils.getCriteria;
 import static org.ohdsi.webapi.user.importer.providers.OhdsiLdapUtils.valueAsString;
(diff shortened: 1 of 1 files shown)


src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java
```diff
@@ -24,12 +27,49 @@ public abstract class AbstractLdapProvider implements LdapProvider {
...
+        if (i == wildcardIndexes.size() - 1 && wildcardIndex < searchString.length() - 1) {
+          String substring = searchString.substring(wildcardIndex + 1, searchString.length());
+          buff.append(LdapEncoder.filterEncode(substring));
+        }
+        previousWildcardIndex = wildcardIndex;
+      }
+    }
+
+    return buff.toString();
   }
```
chrisknoll, line 73: Couldn't this all be replaced with:

```
List<String> tokens = StringUtils.split(searchString, wildcard);
List<String> encodedTokens = tokens.replaceAll(LdapEncoder::filterEncode);
return StringUtils.join(encodedTokens, wildcard);
```

chrisknoll (changes requested):

src/main/java/org/ohdsi/webapi/user/importer/providers/AbstractLdapProvider.java
ssuvorov-fls: Hi
I've checked it and it doesn't work as expected
For `*` it returns empty string
For `user*` it returns "`user`
For `user*cn*` it returns "`user*cn`
My code will return correct string without any limitations of wildcard position

chrisknoll: Understood, I didn't think through the corner cases, but your implementation is pretty complicated and I think I can modify the suggested with a few tweaks to hit the corner cases:

```
if (searchString.isEmpty() || wildCard.equals(searchString)) return searchString; // nothing to encode

List<String> tokens = StringUtils.split(searchString, wildcard);
List<String> encodedTokens = tokens.replaceAll(LdapEncoder::filterEncode);
String encodedSearchString = StringUtils.join(encodedTokens, wildcard) + searchString.endsWith(wildcard) ? wildcard : "";

```

Issue #2288, opened 2023-06-15
API access on OIDC secured WebAPI from other application


RowanErasmus:
Hi,

We have WebAPI running secured with OpenID Connect and would like to access WebAPI directly from another application (so not through the UI). As far as I can tell (and of what I have tested) the current OIDC implementation only allows the code flow (indirect client) through which the user gets redirected and must enter their credentials, this is not workable for connecting from another app.

The [pac4j documentation](https://www.pac4j.org/4.0.x/docs/clients/openid-connect.html) suggest to use a 'HeaderClient' for this scenario, I have implemented and tested this and it seems to work. 

However, before opening a pull request etc. I though maybe to check if anybody had a way to work this without having to make code changes.

Thanks!

konstjar:
Alternative option would be to enable additional authentication type on WebAPI (eg "db") method.

Issue #1884, opened 2021-05-15
Implement API Key for programmatic access to WebAPI without authentication


chrisknoll:
### Expected behavior
It would be nice to be able to access WebAPI (security enabled) by passing an user-owned API key that would bypass authentication and impersonate the person that issued the API key when calling functions.  This would avoid problems where if you authenticate twice, your previous session is closed and you are logged out of the previous session.  This makes it hard to work in something like the Atlas UI and an R session.

### Actual behavior
Only authentication is supported, where you pass the authentication header that you got from a login request, and then use that to access WebAPI.  If you log in again, the session is closed leading to confusing errors about 'access denied' when you accidently log into a new session.

### Implementation Notes

The functional requirements are as follows (open to discussion):
- Any user can create an API Key.  
- A user may create one or more API Keys.
- When you create a key, you can specify an expriation days such that the key will automatically expire after a fixed number of days from creation.
- A user may opt to 'refresh' an API key which will reset the expiration date to a new date.
- A user may delete an API key at any time which will prevent anyone who has a copy of it from using it to access WebAPI as the user.
- WebAPI should accept an alternative key `WEBAPI_KEY` in the request header that when present, it will attempt to resolve the key to the issuing user, and if successful, will continue the request under the context of the user that issued the key.
- This feature can be disabled by WebAPI configuration setting.
- The key should be encrypted in the database (ie: not stored in plain-text)

### Database Details

Probably only 1 new table is required: API_key

|column name|type|description
|---|---|----|
|key_id|integer|System generated key identifier|
|user_id|integer|The system userID for this user|
|api_key|varchar| The API key, which is a formatted GUID: {XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX}, but encrypted in DB|
|expire_days|integer|the number of days from creation/refreshing that the key will expire|
|expire_date|datetime|The date the key expires.  not user defined, but calcualted from the user-input of expire_days|

### Service Details
Simple CRUD and list operations to manage the 1 (user)-to-many (API keys) relationships.

### User interface details
Fairly simple: a simple form primpting the user to create or delete a collection of issued API keys for the user.  Some helper functions would be nice (ie: copy to clipboard, refresh, etc)

anthonysena:
Linking this issue to #1473 as this issue is aiming to describe both the use case and implementation for developing the personal access token for use with WebAPI/Atlas.

anthonysena:
[...]

Issue #1473, opened 2020-03-04
Personal Access Token for security-enabled WebAPI 


anthonysena:
### Problem Statement

When WebAPI `AtlasRegularSecurity` is enabled, requests to WebAPI require an `Authorization` token as part of the request as described on the wiki: https://github.com/OHDSI/WebAPI/wiki/Security-Implementation#signing-requests-to-protected-methods. This is cumbersome when using the WebAPI from applications other than ATLAS since obtaining the `Authorization` token is only possible when accessing WebAPI via ATLAS today. As an example, we have R code that aims to use WebAPI to access design elements, etc which cannot participate in the authentication process enabled by ATLAS/WebAPI

### Background 

To obtain the `Authorization` requires the security handshake with one of the configured security providers to ensure that authentication is completed. The authentication process may vary between providers (i.e. AD, oAuth, etc) and may require redirection to 3rd party sites for authentication, etc. Doing this in the context of R is not practical since it would require a browser to do this work.

### Ideas to move forward

I like the idea of having a personal access token as is done for accessing the GitHub API and is described here in these articles: 

- https://happygitwithr.com/github-pat.html. 
- https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line

The advantage of this is that we can insist on the proper security mechanisms through ATLAS and use that as mechanism to create & store a personal access token.

Please consider this as a working idea vs. a solution - I think the more we can find a workable solution to this for 2.8 it will help us think through a more robust solution for 3.0.

anthonysena:
Recap of a discussion on this topic with @gklebanov, @YaroslavTir, @wivern, @chrisknoll 

---------------

- In general, we want to avoid storing any sensitive information about identity (i.e. password) in the WebAPI database. So, the notion of having a personal access token stored in the WebAPI database is undesirable and should be avoided if possible.
- We reviewed the list of supported authentication providers in WebAPI and put together a list of actions for the team to investigate. The idea is that we'll detail the way that we might use R to interact with WebAPI using each authentication mechanism and what (if any) pre-requisite steps are required. 

#### Auth Providers
- [ ] Active Directory @anthonysena, @chrisknoll. Since this may have some variations based on platform, we broke this into subtasks
  - [ ] Windows - @chrisknoll 
  - [ ] Linux – Look at using Keystore and/or Kerbros - @wivern Linking to issue #1374
- [ ] OpenID @anthonysena
  - Anthony to prove this out with Elixir. 
[...]

PR #2289, opened 2023-06-16
Enable OIDC authentication with a direct client

RowanErasmus:
The OpenID Connect story continues.

As per this issue #2288 we are trying to access WebAPI programmatically and the current OIDC implementation only allows the code flow through the UI, this adds the option to authenticate with a token supplied by the identity provider.

I think, as far as the needs of EMC go, with this we are now at a fully functional OIDC implementation for WebAPI :-)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -953,6 +953,17 @@
     </dependency>
     <dependency>
       <groupId>org.pac4j</groupId>
+      <artifactId>pac4j-http</artifactId>
+      <version>${pac4j.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.core</groupId>
+          <artifactId>jackson-databind</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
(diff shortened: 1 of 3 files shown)


anthonysena:
@RowanErasmus - from discussion on Atlas WG, we'd like to consider these changes in light of the discussion on #1473 and #1884. If your team needs to use this, please use this side-branch you've developed here but we'll want to consider these requirements when we develop the design for those issues.

anthonysena:
> I think we also discussed that this functionality may have a use-case, so we can incorporate this change for use when you want to authenticate via OID, and the separate functionality of API Keys (which can bypass authentication) which is discussed in the different PR.
> 
> @anthonysena , if you agree, can we move forward with merging this PR?

Agreed - I'll approve this PR with the note that it is handling a specific use-case vs the other referenced issues which is a broader solution.

chrisknoll:
I don't believe so. It addes a new path `/user/login/openidDirect` which can be used for direct API access with OpenID.

PR #2290, opened 2023-06-19
Fixed PLE editor text for i18n.

chrisknoll:
Replaced umberOfRepetitionsOfData with minDataCrossValidation.


chrisknoll:
I do not have translations for the other dialects, so we can accept this PR with english and accept a PR for the language-specific terms later.

PR #2301, opened 2023-07-31
Filter cohorts and concepts by permission (#2245)

chrisknoll:
Enable WebAPI to do filtering based on READ permissions. The new property is called security.defaultGlobalReadPermissions.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
+++ b/src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
@@ -86,6 +86,7 @@ import org.springframework.batch.core.configuration.annotation.JobBuilderFactory
 import org.springframework.batch.core.configuration.annotation.StepBuilderFactory;
 import org.springframework.batch.core.job.builder.SimpleJobBuilder;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.jdbc.core.RowMapper;
@@ -203,6 +204,9 @@ public class CohortDefinitionService extends AbstractDaoService implements HasTa
 
 	@Autowired
 	private VersionService<CohortVersion> versionService;
+
+        @Value("#{'${security.defaultGlobalReadPermissions}'.equals(false)}")
+	private boolean defaultGlobalReadPermissions;
 
 	private final MarkdownRender markdownPF = new MarkdownRender();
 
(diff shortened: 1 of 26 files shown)


chrisknoll:
@rkboyce : please pull this branch and test on your local environment.

rkboyce:
All of the following tests passed:

### Configuration 1:  

- WebAPI: <security.defaultGlobalReadPermissions>false</security.defaultGlobalReadPermissions> 
- Atlas: configLocal.enablePermissionManagement = true; 
- A test user  with the following roles: 
```
Atlas users
Moderator
Source user (<some source>)
admin
cohort creator
cohort reader
concept set creator
public
<user specific role>
```
- A test user with the following roles but who does not have the permissions that follow the listed roles:

```
-- Roles
Source user (<some source>)
cohort creator
concept set creator
public
<user specific role>

-- The user does not have these permissions 

    	      'conceptset:*:get',
    	      'conceptset:*:expression:get',
    	      'conceptset:*:version:*:expression:get',    	       
    	--
              'cohortdefinition:*:get',
              'cohortdefinition:*:info:get',
              'cohortdefinition:*:version:get',
              'cohortdefinition:*:version:*:get',        
        --        
	      'cohort-characterization:*:get',
               'cohort-characterization:*:generation:get',
	 	'cohort-characterization:generation:*:get',
		'cohort-characterization:design:get',
		'cohort-characterization:*:design:get',
		'cohort-characterization:design:*:get',
		'cohort-characterization:*:version:get',
		'cohort-characterization:*:version:*:get',
		--
		'pathway-analysis:*:get',
		'pathway-analysis:*:generation:get',
		'pathway-analysis:generation:*:get',
		'pathway-analysis:generation:*:result:get',
		'pathway-analysis:generation:*:design:get',
		'pathway-analysis:*:version:get',
		'pathway-analysis:*:version:*:get'
		--
		'ir:*:get',
		'ir:*:copy:get',
		'ir:*:info:get',
		'ir:*:design:get',
		'ir:*:version:get',
		'ir:*:version:*:get'
		--		
		'estimation:*:get',
		'estimation:*:copy:get',
		'estimation:*:download:get',
		'estimation:*:export:get',
		'estimation:*:generation:get',
		'comparativecohortanalysis:*:get', -- might only apply to older versions of Atlas/WebAPI
		--
		'prediction:*:get',
		'prediction:*:copy:get',
		'prediction:*:download:get',
		'prediction:*:export:get',
		'prediction:*:generation:get',
		'prediction:*:exists:get',
		'plp:*:get' -- might only apply to older versions of Atlas/WebAPI
```

*Test 1: expected behavior - filtering of listed entities based on READ permissions by WebAPI*

User logs in and can view all of the entities that the user  has READ permissions to (concept sets, cohort definitions, characterizations, cohort pathways, incidence rates, estimation, prediction)  - _Passed_

*Test 2:  ability to add READ/WRITE permissions to any entity that the user has WRITE permissions to*

User creates an entity or opens an existing entity that they have WRITE permissions for and can add  READ/WRITE permissions to that entity for another user. The other user will be able to view (if given READ permissions) and edit (if given WRITE permissions)  - _Passed_ 

*Test 3:  ability to remove READ/WRITE permissions to any entity that the user has WRITE permissions to*

User opens an existing entity that they have WRITE permissions for and can remove the READ/WRITE permissions to that entity for another user. The other user will not be able to view (if READ permissions are removed) nor edit (if WRITE permissions are removed) that entity  - _Passed_

anthonysena (approved):

src/main/java/org/ohdsi/webapi/service/CohortDefinitionService.java
```diff
@@ -406,11 +410,12 @@ public GenerateSqlResult generateSql(GenerateSqlRequest request) {
 	@Transactional
 	public List<CohortMetadataDTO> getCohortDefinitionList() {
 		List<CohortDefinition> definitions = cohortDefinitionRepository.list();
-
 		return definitions.stream()
+						.filter(!defaultGlobalReadPermissions ? entity -> permissionService.hasReadAccess(entity) : entity -> true)
```
pieterlukasse, line 414: @chrisknoll @rkboyce I just came across this change and was wondering why the permissionService is only checked here. Should it not also be used in `getCohortDefinitionRaw`(https://github.com/OHDSI/WebAPI/pull/2301/files#diff-f0d513a6643cd34ac8cee922ef4e560fe9cbc270929d9054520ccc7055fbad10R478-R487) for example?

chrisknoll: no, the purpose of this filter is to remove non-permissioned entities (like cohort definitions) from the calls to get the list of cohort definitions (or concept sets) from the server.  The getCohortDefinitionRaw (and other method) fetches a single cohort definition from the service, which would already be 'permission denied' by the security layer...so we don't need to filter in this method as this method will be rejected if you don't have permissions.

PR #2305, opened 2023-08-28
Update Dependencies and address defaultGlobalReadOnly param

chrisknoll:
Updated circe to version 1.11.1
Updated StandardAnalysisUtils to 1.4.0.
Fixed the assignment of defaultGlobalReadOnly field.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
 
     <commons-fileupload.version>1.3.3</commons-fileupload.version>
 
-    <circe.version>1.11.0</circe.version>
+    <circe.version>1.11.1</circe.version>
     <jersey.version>2.14</jersey.version>
     <SqlRender.version>1.13.1</SqlRender.version>
     <hive-jdbc.version>3.1.2</hive-jdbc.version>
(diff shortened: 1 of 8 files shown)


chrisknoll:
@rkboyce , just FYI, I found something odd in how the permissions were being specified...although I did the previous PR and tested the functionality, I didn't pay close enough to what the value of the paramter was vs. what it represented (ie, was testing turning it on and off to test the difference but didn't look closely at the name).

In this case, We are talking about the setting `security.defaultGlobalReadPermissions`, which sounds like if it is true, you would be granted global read permissions by default.  But what I found in code was a sort of reversal:

```
	@Value("#{'${security.defaultGlobalReadPermissions}'.equals(false)}")
	private boolean defaultGlobalReadPermissions;
```

So it's storing the field 'defaultGlobalReadPermissions as the opposite value that was specified in the config...which feels wrong. I've updated it here:

```
	@Value("${security.defaultGlobalReadPermissions}")
	private boolean defaultGlobalReadPermissions;
```

I am pinging you here because you may have this deployed in your own environment where you are reversing the value (true vs. false) in your own config, and I just wanted to let you know of the correction I'm making here.

anthonysena (approved):

PR #2315, opened 2023-10-11
Add concept sets search (mergeable)

alex-odysseus:


anthonysena:
Linking to #2177. 

From discussion on the Atlas/WebAPI WG call: This set of changes for advanced concept set search is complex and is currently lacking information on how the functionality is intended to work. As an example, is concept set indexing something that run synchronously and could this delay saving of a concept set? It is also unclear how we can run this branch locally to test the functionality since it is unclear how to configure SOLR to make the functions of this branch work properly.

We're proposing to move this to v2.15 but need some feedback on this @alex-odysseus

PR #2341, opened 2024-02-02
Permission Performance Optimizations

chrisknoll:
Implemented permission lookup to bypass hibernate.
Moved PermissionsDTO to PermissionManager
Removed GSON to use the app-wide ObjectMapper from Jackson

Required for OHDSI/Atlas#2912.

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/UserService.java
+++ b/src/main/java/org/ohdsi/webapi/service/UserService.java
@@ -1,5 +1,6 @@
 package org.ohdsi.webapi.service;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.odysseusinc.logging.event.*;
 import org.eclipse.collections.impl.block.factory.Comparators;
 import org.ohdsi.webapi.shiro.Entities.PermissionEntity;
@@ -50,6 +51,7 @@ public class UserService {
     public String login;
     public String name;
     public List<Permission> permissions;
+    public JsonNode permissionIdx;
 
     public User() {}
 
@@ -114,6 +116,8 @@ public class UserService {
     user.login = currentUser.getLogin();
     user.name = currentUser.getName();
     user.permissions = convertPermissions(permissions);
+    user.permissionIdx = authorizer.queryUserPermissions(currentUser.getLogin()).permissions;
+    
 
     return user;
   }
--- a/src/main/java/org/ohdsi/webapi/shiro/PermissionManager.java
+++ b/src/main/java/org/ohdsi/webapi/shiro/PermissionManager.java
@@ -1,11 +1,8 @@
 package org.ohdsi.webapi.shiro;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.google.gson.JsonArray;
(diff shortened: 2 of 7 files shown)


src/main/java/org/ohdsi/webapi/shiro/PermissionManager.java
```diff
@@ -1,5 +1,11 @@
 package org.ohdsi.webapi.shiro;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
```
anthonysena, line 8: com.google.gson was removed from pom.xml

anthonysena (commented):
Looks good from review on the Atlas WG call - just one comment for review to make sure we've removed the `com.google.code.gson` reference.

src/main/java/org/ohdsi/webapi/shiro/PermissionManager.java
```diff
@@ -305,7 +328,51 @@ public Set<PermissionEntity> getUserPermissions(UserEntity user) {
 
     return permissions;
   }
+  
+  public PermissionsDTO queryUserPermissions(final String login) {
```
pieterlukasse, line 332: @chrisknoll should `getUserPermissions` be deprecated in favor of this method?

src/main/java/org/ohdsi/webapi/service/UserService.java
```diff
@@ -114,6 +116,8 @@ public User getCurrentUser() throws Exception {
     user.login = currentUser.getLogin();
     user.name = currentUser.getName();
     user.permissions = convertPermissions(permissions);
+    user.permissionIdx = authorizer.queryUserPermissions(currentUser.getLogin()).permissions;
```
pieterlukasse, line 119: does this mean `user.permissions` will no longer be needed?

chrisknoll: It's unclear. When I added the code, I was concerned with not breaking backwards compatibility.  While I did do a search on where the user.permissions field was referenced, it doesn't find those uses by downstream apps like Atlas (ie: the value is populated, returned to Atlas via webAPI call, so removing it will break it there).

When we move into the 3.0 WebAPI changes, ( #2354) we can be more aggressive about removing obsolete/unused code.

src/main/java/org/ohdsi/webapi/shiro/PermissionManager.java
chrisknoll: Depends:  
the `getuserPermissions` is returning PermissionEntity...which is the JPA Entity Bean that represents the row of data in the DB, and fetchign permissions this way involves getting the roles from the user, iterating through each role and adding each permission from those roles into a final Set of `PermissionEntity`.  The use of Set de-dupes permissions across roles.

`queryUserPermissions` provides more of a 'view' of distinct permissions associated with a user bypassing the JPA interface of sweeping through roles.  It also structures the result to be 'indexed' such that you can quickly navigate to specific permissions that begin with a certain term.  

The inputs to each of these are a bit different too:  the former wants a ID or UserEntity, the latter works off of a login, and this is just based on the use-case of when you'd use one or the other.   When we get to the point that we can clean up the API/object model in the 3.x release we can normalize and simplify a lot of this to a more simple, straight forward implementation.

PR #2316, opened 2023-10-26
Flyway read restricted atlas user 2

rkboyce:
This request is to add the new 'read restricted Atlas Users' role (system role, ID 15) to the sec_role table with the associated permissions in the sec_role_permission table. This will allow users of 2.14 WebAPI and Atlas who want to restrict READ permissions to have an appropriate role that they can assign (as admin) to user accounts.

Issue #2222, opened 2023-02-28
Idea : filter items shown to users in summary lists (e.g., concept sets, cohort definitions, etc)  based on permissions 


rkboyce:
Use case: 
Dozens of users in a data commons are using Atlas to create concept sets and cohort definitions. Many of them are just learning how build these artifacts and some have obtained mastery. Users do want to view/browse concept sets and cohort definitions that could be useful for their analyses (reusability) but often other users would prefer that their artifacts not be listed until they give permission (e.g., because they are in draft mode or private to a project or team).

Current behavior:
All users will see up to hundreds of concept sets and cohort definitions when they click on  the `#/conceptsets` and `#/cohortdefinitions` endpoints in Atlas. This can be very confusing to the users as they will see numerous things that are described only by title and often with little context to help them know if its relevant to their project. They might not have a way of knowing know if the artifacts are in draft mode or have been tested. Also, many users would prefer the ability to give permission to share the viewing/listing of their artifacts with either specific individuals, groups, or the community of users as a whole. Reasons for wanting to control access include a desire to wait until the artifact is solidly tested prior to contributing it to the community, wanting the the project to be complete and published before they share the work, or other reasons. 

Proposal: 
Modify the WebAPI to use a new property setting that will cause the WebAPI instance to return only concept sets, cohort definitions, or other artifacts (e.g., incidence rate analysis, characterizations, etc) in the response provided by endpoints such as  `#/conceptsets`, `#/cohortdefinitions`, etc.  that a user has explicit permission to view based on the sec_role_permission and sec_permission tables in the security schema. 

Justification for doing this in WebAPI: 
It seems more appropriate from an authorization standpoint for the service to filter artifact lists based on the authenticated user's permissions rather than having the client application (Atlas) do so. In the latter case, the client app still receives a listing that someone could sniff using developer tools within the browser. Note that if this idea is accepted, it will likely make since to add the ability of an author to edit read permissions using Atlas "Configure access". This will add a new ability given that the current Configure Access focuses only on  whom can edit an artifact.

anthonysena:
From discussion on the Atlas WG today, some notes from discussion:

- At the moment, WebAPI does not have the notion of a "read" permission for entities - they are assumed to be read-only for all users in the system. This is problematic in the scenario described above as it creates a mess of "work in progress" entities. It also does not provide the fine-grained controls required by data commons such that work can be done in private until such a time that it can be shared.
[...]

PR #2342, opened 2024-02-02
Fix/flyway atlas read only user

rkboyce:
This is a fix to the Atlas read-restricted role permissions SQL. The prior SQL permissions select statement did not adequately address multiple schemas for vocabulary and sources. As a result, some the intended read access would not work as expected for some WebAPI configurations. This script does a better job of accounting  for all sources and schemas relevant to the user.


chrisknoll:
With the update for WildCard permissions in #2355, we should be able to just use * permissions to grant permission to any source. ie: `vocabulary:*:concept:*:get` will let you query concepts from any source vocab.   Also, this will work with any existing or future source because new source keys will match with *.

@rkboyce: would you try to perform a test on your env with the branch from #2355, and possibly will need to make a new source for testing because if this PR was deployed to your WebAPI DB, then you have all the source-specific permissions granted already and you want to test it on a new source (so the * takes effect).

src/main/resources/db/migration/postgresql/V2.14.1.20240130082100__fix_atlas_read_restricted_role_and_permissions.sql
```diff
@@ -0,0 +1,226 @@
...
+ from (
+ select *
+ from (values 
+		('vocabulary:')
+	) t1 (l)
+ cross join 
+	( select source_key 
+	  from vocab_source
+	) t2 (m)
+ cross join
```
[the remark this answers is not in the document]
pieterlukasse: one issue with this is that permissions like `vocabulary:*:search:*:get` do not exist in the `sec_permission` table yet, so these would first have to be added as well (maybe as a first step of this migration script).

src/main/resources/db/migration/postgresql/V2.14.1.20240130082100__fix_atlas_read_restricted_role_and_permissions.sql
```diff
@@ -0,0 +1,226 @@
...
+		'cohortdefinition:*:check:get',
+		'cohortdefinition:*:check:post',
+		'cohortdefinition:*:copy:get',
+		'cohortdefinition:*:exists:get',
+		'cohortdefinition:*:export:conceptset:get',
+		'cohortdefinition:*:tag:*:delete',
+		'cohortdefinition:*:tag:post',
+		'cohortdefinition:*:version:*:createAsset:put',
+		'cohortdefinition:*:version:*:delete',
+		'cohortdefinition:*:version:*:put',
```
pieterlukasse, line 86: these * permissions should not be part of role 15, as they enable users to access cohortdefinition data that should not be accessible to them. See also https://github.com/uc-cdis/WebAPI/pull/142

chrisknoll: Hi @pieterlukasse ,

I skimmed through the referenced issue you linked, but didn't digest all of it (I did leave a comment tho), but in the context of your concern, the role is supposed to grant global read permission to anyone who has the role, so, in your context, wouldn't hte solution be to NOT assign role 15 to any users who you want to restrict those actions on cohort definitions?  I don't think the role is wrong, it's just who you would apply the role permissions to (which in your case, sounds like no one).

pieterlukasse: @chrisknoll thanks for reviewing and adding the comment here and in the linked ticket!

Please note that the role 15 is supposed to replace the "Atlas user" role (role 10) and ensure all artifacts are visible only to their creator (see also https://github.com/OHDSI/WebAPI/issues/2222 and https://github.com/OHDSI/WebAPI/wiki/Read-restricted-Configuration). This means that permissions like `cohortdefinition:*:get` should _not_ be part of role 15 (as correctly done in https://github.com/OHDSI/WebAPI/pull/2316), as well as any other permission that enables the user to get access to  a cohortdefinition by calling other endpoints (meaning permissions such as `cohortdefinition:*:copy:get` should also _not_ be part of role 15). 

@rkboyce please correct above if anything is wrong!

pieterlukasse: Thanks @chrisknoll, I think we are aligned.

> one where you would have wild-card permissions for all the :get permissions. The read access role would grant you read permission to everything.

correct.  

> where's copy?

I've been asking that myself. That is why I fixed it here as well: https://github.com/uc-cdis/WebAPI/pull/142/files#diff-b0ddc2ab44ac01c9f5a2769cbfc9b8e54224dc0ea8a6c03b6113bc89b45487e4R23, https://github.com/uc-cdis/WebAPI/pull/142/commits/fa2eafe0266feeab5f229169900145c81d80cddf

>  we need to change it in the 3.X line

I agree. It would be great to simplify this part. As this discussion exemplifies, it is al too easy to get lost in the many permissions, and wrong permissions can slip into the configuration more easily.

pieterlukasse:
@rkboyce I made 2 PRs above that address the issues for this PR.

rkboyce:
Though not well documented, this pull request actually tried to address two issues: 1) adding read permissions for the 'atlas read restricted' role to existing sources, and 2) addressing some missing permissions in the WebAPI permissions templates needed for read permissions to work properly. 
 
- The consensus is that having a flyway script to add source READ permissions is not a good idea b.c it will give a false sense of security to users that the user will have read access any new sources added after the WebAPI update.  At today's call (09/17), the idea was proposed to move any management of source 'read' access to the options provided in the Admin section of Atlas.
- Addressing some missing permissions in the WebAPI permissions templates needed for read permissions to work properly is an important consideration that parts of this pull request might have addressed

The action decided for now was to create an issue ticket that enumerates the issues above and references this pull request. We might remove the flyway parts of the pull request and retain the WebAPI permission templates but this will require further review and testing. The issue created is: https://github.com/OHDSI/WebAPI/issues/2398

pieterlukasse:
Thanks for the summary @rkboyce , I think that makes sense. An extra point to mention is that the flyway script, as proposed here, does remove all * permissions from role 15, and this potentially breaks some of the functionality for some of the features in the Atlas UI. These * permissions need to be translated into their respective artifact specific versions of the permission first, before they can be fully dropped from role 15.

PR #2355, opened 2024-02-27
Permission Checks now use Wildcard semantics.

### LINKED ISSUE #2353: has[Write|Read]Access does not use wildcard permissions
### Expected behavior

Wildcard (`*`) permissions should be honored when checking read/write access.

### Actual behavior

When checking entity permissions via `hasWriteAccess` or `hasReadAccess` it is checking the `entitySchema` (which is a string template of permissions for the given entity) for an exact string match for the specific entity ID.  This means that permission granted to the user via wildcard permissions (such as a role that grants write permission to everything via `*`) is not honored.

Ie: checking `cohortdefinition:1:get` against a grant of `cohortdefinitiion:*:get` will fail, when the wildcard should allow any ID to pass.

This is a symptom of #2323 , in so far as the wildcard permission to `cohortdefinition:*:get` should be seen as read permission.  and in that case only the specific `cohortdefinition:{id}:get` is recognized as the read permission.   But this is a general symptom that we're not using WildCard permissions when checking perms in some cases but we are in other (such as URL checks).

### Technical Considerations

As I was researching how permissions were applied, I was greatly confused by the approach taken in the code, and will try to describe it here so that the original devs (or people closer to the code) can comment and clarify.

The [hasWriteAccess()](https://github.com/OHDSI/WebAPI/blob/f258186ce9955ff5ee1cef220de88da7a9049de2/src/main/java/org/ohdsi/webapi/security/PermissionService.java#L255) is doing something particularly strange:   I would have expected that in order to determine if you have write permissions, you get the set of permissions from [getTemplatesForType()](https://github.com/OHDSI/WebAPI/blob/f258186ce9955ff5ee1cef220de88da7a9049de2/src/main/java/org/ohdsi/webapi/security/PermissionService.java#L111) and just check that the user has all necessary permissions in order to determine if the user has `WRITE` access.  However, the code builds a permission cache in [preparePermissionCache()](https://github.com/OHDSI/WebAPI/blob/f258186ce9955ff5ee1cef220de88da7a9049de2/src/main/java/org/ohdsi/webapi/security/PermissionService.java#L182) which does the following:

```
    public void preparePermissionCache(EntityType entityType, Set<String> permissionTemplates) {
        if (permissionCache.get().get(entityType) == null) {
            final ConcurrentHashMap<String, Set<RoleDTO>> rolesForEntity = new ConcurrentHashMap<>();
            permissionCache.get().put(entityType, rolesForEntity);

            List<String> permissionsSQLTemplates = permissionTemplates.stream()
                    .map(pt -> getPermissionSqlTemplate(pt))

            Map<Long, RoleDTO> roleDTOMap = new HashMap<>();
            permissionsSQLTemplates.forEach(p -> {
                Iterable<PermissionEntity> permissionEntities = permissionRepository.findByValueLike(p, PERMISSION_ENTITY_GRAPH);
                for (PermissionEntity permissionEntity : permissionEntities) {
                    Set<RoleDTO> roles = rolesForEntity.get(permissionEntity.getValue());
                    if (roles == null) {
                        rolesForEntity.put(permissionEntity.getValue(), new HashSet<>());
                    }
                    Set<RoleDTO> cachedRoles = rolesForEntity.get(permissionEntity.getValue());
                    permissionEntity.getRolePermissions().forEach(rp -> {
                        RoleDTO roleDTO = roleDTOMap.get(rp.getRole().getId());
                        if (roleDTO == null) {
                            roleDTO = conversionService.convert(rp.getRole(), RoleDTO.class);
                            roleDTOMap.put(roleDTO.getId(), roleDTO);
                        }
                        cachedRoles.add(roleDTO);
                    });
                }
            });
        }
    }
```

in short: this will find all the roles that contain the specified permissions (from the template), and back in `getRolesHavingPermissions` it performs this work to return `RoleDTO`s that contain all the specified permissions in the template:

```
    public List<RoleDTO> getRolesHavingPermissions(EntityType entityType, Number id) {
        Set<String> permissionTemplates = getTemplatesForType(entityType, AccessType.WRITE).keySet();
        preparePermissionCache(entityType, permissionTemplates);

        List<String> permissions = permissionTemplates.stream()
                .map(pt -> getPermission(pt, id))
        int fitCount = permissions.size();
        Map<RoleDTO, Long> roleMap = permissions.stream()
                .filter(p -> permissionCache.get().get(entityType).get(p) != null)
                .flatMap(p -> permissionCache.get().get(entityType).get(p).stream())
                .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
        List<RoleDTO> roles = roleMap.entrySet().stream()
                .filter(es -> es.getValue() == fitCount)
                .map(es -> es.getKey())
        return roles;
    }
```

Note: using WildCard semantics, it's possible a person may match 2 permissions for the same check (the * check and the literal {id} check), throwing off the 'fitCount' check which says we want the roles that have exactly the correct permissions.

Back in `hasWriteAccess`:

```
                } else {
                    EntityType entityType = entityPermissionSchemaResolver.getEntityType(entity.getClass());

                    List<RoleDTO> roles = getRolesHavingPermissions(entityType, entity.getId());

                    Collection<String> userRoles = authorizationInfo.getRoles();
                    hasAccess = roles.stream()
                            .anyMatch(r -> userRoles.stream()
                            .anyMatch(re -> re.equals(r.getName())));
                }
```

This block appears to be saying : ' if the user has any role that matches any role that has the required write permissions matches any role that is assigned to the user.

This seems like a wildly round-about way to answer the question 'do you have write access'.  If write access is determined by if they have the required permission, a permission grant of `*` should match all 5 of the required permissions and thus grant the user TRUE for write access.   

However, I  may be missing some underlying concept of this logic that is not clear.  For example, maybe associating back to the role-level permissions handles some corner case...or maybe it's faster to check a set of roles a user has compared to a set of permissions (although we still need to check all the permissions to find the roles these permissions belong to).   It also seems to force that all permissions must be found in a single role in order to 'count', but I'm not sure why that needs to be the case here:  If RoleA grants some the required WRITE perms, and RoleB grants the rest, wouldn't a person assigned to RoleA and RoleB then have write perms to the entity?

I would appreciate if @alex-odysseus and members of @OHDSI/odysseus could speak on these considerations because they were responsible for the initial implementation.

chrisknoll:
Removed role cache from Permission Service.
Changed read/write permission checks to use permissions instead of roles. 
Permission checks are using Subject.isPermitted() which honors wildcard semantics. 
Altered JwtAuthRealm to filter user permissions to either * or first element of permission to check for speed. 
Changed permission index from JsonNode to Map<>.  Serializes same way, but map semantics are simpler to navigate. 
Altered AuthrizationInfo to contain index of Permissions and store Wildcard perms.

General cleanup of unused imports and removed unused dependencies (ie: Autowired fields were removed if no longer needed).

Fixes #2353.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -761,18 +761,6 @@
       <artifactId>xml-security-impl</artifactId>
       <version>1.0</version>
     </dependency>
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-test</artifactId>
-      <version>${spring.boot.version}</version>
-      <scope>test</scope>
-        <exclusions>
-            <exclusion>
-                <groupId>com.vaadin.external.google</groupId>
-                <artifactId>android-json</artifactId>
-            </exclusion>
-        </exclusions>
-    </dependency>
     <!-- It is overriding a transitive dependency from the dependency above -->
     <!-- After migrating to Spring Boot 2.x we should get rid of it -->
     <!-- It is currently inefficient as com.nimbusds:nimbus-jose-jwt has hard bounds [1.3.1,2.3] -->
@@ -1030,24 +1018,6 @@
       <version>1.9.2</version>
     </dependency>
     <dependency>
-      <groupId>org.dbunit</groupId>
-      <artifactId>dbunit</artifactId>
-      <version>2.7.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.github.springtestdbunit</groupId>
-      <artifactId>spring-test-dbunit</artifactId>
-      <version>1.3.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>pl.pragmatists</groupId>
-      <artifactId>JUnitParams</artifactId>
-      <version>1.1.0</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcprov-jdk15on</artifactId>
       <version>${bouncycastle.version}</version>
@@ -1209,18 +1179,6 @@
       <version>0.15.2</version>
     </dependency>
     <dependency>
-      <groupId>com.opentable.components</groupId>
-      <artifactId>otj-pg-embedded</artifactId>
-      <version>0.13.1</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-        <groupId>com.github.mjeanroy</groupId>
-        <artifactId>dbunit-plus</artifactId>
-        <version>2.0.1</version>
-        <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.dom4j</groupId>
       <artifactId>dom4j</artifactId>
       <version>${dom4j.version}</version>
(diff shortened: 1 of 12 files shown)


chrisknoll:
Leaving this PR as draft as there may be additional code to be removed from the codebase as a result of this change:  things like sql templates and entity graphs that are used to find roles for the user may no longer be necessary (I'm not sure where we would ever use roles to determine a permission, except for 'Admin').

As we code review, feel free to identify blocks of code that may be redundant or no longer necessary.  We are always making efforts to simplify the codebase.

rkboyce:
@chrisknoll  - I tested this change in my environment where have the read-restricted feature enabled and everything worked as expected. Great job!

chrisknoll:
The last commit adds tests.

For people on the call, DB Unit has several modes of inserting described here:
https://www.dbunit.org/components.html

The code I demonstrated did a CLEAN_ALL at first which blows away all the data in a table, which would include 'system roles' so it was rightfully pointed out about that issue.

I attempted a quick demo to show how the test could be changed, and I switched it to DatabaseOperation.INSERT thinking it would insert the data, but from the docs, DatabaseOperation.INSERT will fail if the table exists (ie: INSERT requires a table creation).   The correct operation I wanted to use is REFRESH, which will leave existing rows alone and add new records based on the test data set.

@anthonysena this is ready for review.

anthonysena (approved):

Issue #2370, opened 2024-05-09
Proposal: Profiles section - compatibility with bigints types


luisleon1894:
Hi! I am developer in IOMED (link: [Home](https://www.iomed.health/)). At our company company, one of the services we use is Atlas + WebPAPI. Recently , we encountered a limitation in the `Profiles` section related to the observartion_period.

### Explanation

We are aware that in the [official DDL](https://github.com/OHDSI/CommonDataModel/blob/3b84ef61144a9a92f04443d98fbbdda35bea885e/ddl/5.4/postgresql/OMOPCDM_postgresql_5.4_ddl.sql#L24) the `observation_period_id` column is of integer type. However in our environment we have set this column as bigint.

Upon reviewing the traceback we notice that the function used currently is `getInt`. Is there any chance to use the `getLong` instead?

### Steps to reproduce behavior & Actual behavior

1. Have a `observation_period_id` bigger than a Integer, for example:

|observation_period_id|person_id|observation_period_start_date|observation_period_end_date|period_type_concept_id|
|---------------------|---------|-----------------------------|---------------------------|----------------------|
|389686681703353278|12319690674377349|2007-09-15|2023-06-16|44814724|

2. Search in `Profiles` the `person_id` 

![image](https://github.com/OHDSI/WebAPI/assets/26791006/83e7adc0-687b-4d0b-9178-294ddebbec80)

3. See logs from web api

```
Caused by: org.postgresql.util.PSQLException: Bad value for type int : 389686681703353278
	at org.postgresql.jdbc.PgResultSet.toInt(PgResultSet.java:3218) ~[postgresql-42.3.7.jar!/:42.3.7]
	at org.postgresql.jdbc.PgResultSet.getInt(PgResultSet.java:2408) ~[postgresql-42.3.7.jar!/:42.3.7]
	at org.postgresql.jdbc.PgResultSet.getInt(PgResultSet.java:2836) ~[postgresql-42.3.7.jar!/:42.3.7]
	at org.ohdsi.webapi.service.PersonService$2.mapRow(PersonService.java:103) ~[classes/:?]
	at org.ohdsi.webapi.service.PersonService$2.mapRow(PersonService.java:95) ~[classes/:?]
	at org.springframework.jdbc.core.RowMapperResultSetExtractor.extractData(RowMapperResultSetExtractor.java:93) ~[spring-jdbc-4.3.25.RELEASE.jar!/:4.3.25.RELEASE]
[...]

PR #2371, opened 2024-05-09
Update PersonService.java

luisleon1894:
Issue: https://github.com/OHDSI/WebAPI/issues/2370


chrisknoll:
Ok, after discussion on the working group call:

While we acknowledge this isn't part of the CDM spec, the ObservationPeriod class that was implemented as `long id` so it does make sense that we would read the value from the query via `getLong`,

So we will accept this PR as-is. Thank you for your contribution!

chrisknoll (approved):

Issue #2359, opened 2024-04-09
Provide configuration option to disable source connection checks


anthonysena:
The following line of code will attempt to check the connection to a configured data source upon startup:

https://github.com/OHDSI/WebAPI/blob/d90d653d1449affd8ef272141cf0af59e635b115/src/main/java/org/ohdsi/webapi/source/SourceService.java#L108

This is problematic for data sources, such as DataBricks, where the database server takes time to activate upon first query which results in WebAPI hanging until the DataBricks environment starts up. We'd like to be able to control this behavior through configuration - either overall for the WebAPI deployment or on an individual data source level.

ganisimov:
@anthonysena @fdefalco I'm looking into this one and going to prepare a PR. Please, let me know in case of concerns or additional requirements. I'm going to add the flag to skip connection check both on deployment and individual data source level. Data source level setting has a precedence over deployment level. This covers simple cases and allows fine tuning with minimum effort in terms of configuration.

anthonysena:
Thanks @ganisimov - it is probably cleaner to just have this setting on the data source and not on the deployment. Presumably you'll have a flyway migration that adds a new column for the data source connection check and we'd have it default to "true" and then we can optionally turn it off where needed. 

So please ignore my suggestion of having an application-level configuration for this setting.

Issue #2350, opened 2024-02-14
Priority call causes application hang in initial load for ATLAS


fdefalco:
### Expected behavior
WebAPI responds or async handles requests for data source priority during initial application loading.

### Actual behavior
WebAPI will block until a 'select 1' is executed against all data sources configured. This is problematic in cases where database platforms have idle/standby modes where upon receiving a query they take several minutes to initialize before being able to respond to the query. (ie, databricks, redshift serverless, spark serverless)

![image](https://github.com/OHDSI/WebAPI/assets/6586384/badf6c80-e57d-4aa4-ae9a-8fad5b82861d)

This check connection appears to be the cause of the issue:
https://github.com/OHDSI/WebAPI/blob/master/src/main/java/org/ohdsi/webapi/source/SourceService.java#L108

For responding to priority it might be better to only make use of the data in the daimon tables without performing a connection test.

anthonysena:
Closed via #2374

PR #2374, opened 2024-05-23
Allow to skip connection check for certain sources

ganisimov:
Addresses #2359

### DIFF
--- a/src/main/java/org/ohdsi/webapi/source/Source.java
+++ b/src/main/java/org/ohdsi/webapi/source/Source.java
@@ -113,6 +113,10 @@ public class Source extends CommonEntity<Integer> implements Serializable {
   @Column(name = "is_cache_enabled")
   private boolean isCacheEnabled;
 
+  @Column(name = "check_connection")
+  private boolean checkConnection;
+
+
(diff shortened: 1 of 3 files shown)


chrisknoll (approved):
Looks good. Is there any other modifications you would want to make before we merge this PR?

ganisimov:
@chrisknoll , the checkConnection() is used in API call /source/connection/{key} here:
https://github.com/OHDSI/WebAPI/blob/6a43da50ab2bf10df18306df658933ec6a49235e/src/main/java/org/ohdsi/webapi/source/SourceController.java#L348
So, I'm not sure if it's desired to suppress connection check here in any case. If the change was requested to optionally skip the check only when dealing with daimons (#2350), I would handle the setting closer to that code and not in checkConnection(). Do you have any thoughts on this?

chrisknoll:
Seems like if the setting on the source is to disable check connections, then it's reasonable that all calls to check connection will just bypass.  So, in the case you show here, you would want to bypass the check in all cases.

ganisimov:
OK then I'm good with the changes, thanks!

PR #2380, opened 2024-07-26
Add check_connection flag to SourceRequest

ganisimov:
Related to #2359

### DIFF
--- a/src/main/java/org/ohdsi/webapi/service/converters/SourceRequestToSourceConverter.java
+++ b/src/main/java/org/ohdsi/webapi/service/converters/SourceRequestToSourceConverter.java
@@ -29,7 +29,7 @@ public class SourceRequestToSourceConverter implements Converter<SourceRequest, 
         source.setKeyfileName(request.getKeyfileName());
         source.setKrbAdminServer(request.getKrbAdminServer());
         source.setKrbAuthMethod(KerberosAuthMechanism.getByName(request.getKrbAuthMethod()));
-        if (null != request.isCheckConnection()) {
+        if (request.isCheckConnection() != null) {
             source.setCheckConnection(request.isCheckConnection());
         }
         return source;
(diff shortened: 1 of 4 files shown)


src/main/java/org/ohdsi/webapi/service/converters/SourceRequestToSourceConverter.java
```diff
@@ -29,6 +29,9 @@ public Source convert(SourceRequest request) {
         source.setKeyfileName(request.getKeyfileName());
         source.setKrbAdminServer(request.getKrbAdminServer());
         source.setKrbAuthMethod(KerberosAuthMechanism.getByName(request.getKrbAuthMethod()));
+        if (null != request.isCheckConnection()) {
```
anthonysena, line 32: @ganisimov can you change the evaluation such that is reads: request.isCheckConnection() != null? The other way is a bit confusing to me at least.

src/main/java/org/ohdsi/webapi/source/SourceController.java
```diff
@@ -238,6 +238,9 @@ public SourceInfo updateSource(@PathParam("sourceId") Integer sourceId, @FormDat
       }
       setKeyfileData(updated, source, file);
       transformIfRequired(updated);
+      if (null == request.isCheckConnection()) {
```
anthonysena, line 241: Same for this evaluation

anthonysena (changes requested):
Looks good with a few minor comments about the `if` statements

src/main/java/org/ohdsi/webapi/service/converters/SourceRequestToSourceConverter.java
ganisimov: Agree, this doesn't make any sense for Java. Thanks for looking into this!

chrisknoll:
Looks good to me, thanks for addressing those comments.

Issue #2329, opened 2023-12-18
Application usage statistics


alex-odysseus:
### Business requirements
There should be a way to know which application functionality is used frequently by users so that any future feature enhancements can be additionally supplemented with some raw data of actual application usage in the relevant areas

Answering to the following questions is essential:
* how often a user performs an Analysis Execution on a particular Data Source
* what are trends of a user accessing a particular application functionality

### Design Ideas and Functional Requirements

As the Audit Trail (https://github.com/OHDSI/WebAPI/pull/1913) contains the most comprehensive data entries of by whom/when/which endpoints have been invoked the newly designed endpoints will rely on those

The first endpoint will be tailored to the particular functionality - Analysis Execution (including Cohorts) on Data Sources
The second endpoint will be more common and may receive an arbitrary list of analyzed endpoints of interest with a notion of placeholders

By default both endpoints will be available for the administrators only (additionally can be administered by assigning the corresponding permissions to a desired user role) and should have a possibility to provide the results both in JSON and CSV formats for its easy processing by the client applications. As user identifiers are usually treated as personal identifiable information (PII) there should be a possibility to exclude their propagation to the results

@mrechkem @Sathyaraos

anthonysena:
Raising a few questions with this requested feature:

> Answering to the following questions is essential:
> 
> 1. how often a user performs an Analysis Execution on a particular Data Source
> 2. what are trends of a user accessing a particular application functionality

For #1, can't we obtain this from the `jobs` table(s) in the WebAPI database? Any analysis that runs is captured in this table and should include the data source.

For #2, can this live outside of the ATLAS DB? Since the audit log trail produces .txt files with this info, wouldn't a simple log parser suffice? Is this really necessary to surface in the the application?

PR #2387, opened 2024-08-27
Application usage statistics

alex-odysseus:
Addressing #2329 

Business logic is based on parsing Audit Trail log entries. As some of the Analysis Execution entries were identified as duplicates a filter has been introduced to count each Analysis Execution precisely once

Usage examples (for the second one {} is used in the 'urlPattern' parameter to indicated a dynamic part of the URL):

```
curl --location --request POST 'localhost:8080/WebAPI/statistic/executions' \
--header 'Content-Type: application/json' \
--header 'Cookie: JSESSIONID=398******************EB0' \
--data-raw '{
"startDate":"2023-11-01",
"endDate":"2023-11-10",
"sourceKey":"SYNPUF_110k"
}'

```

```
curl --location --request POST 'localhost:8080/WebAPI/statistic/accesstrends' \
--header 'Content-Type: application/json' \
--header 'Cookie: JSESSIONID=398************************EB0' \
--data-raw '{
"startDate":"2024-03-12",
"endDate":"2024-03-12",
"endpoints":[{"method":"POST","urlPattern":"/WebAPI/conceptset/check"},{"method":"GET","urlPattern":"/WebAPI/i18n/locales"},{"method":"GET","urlPattern":"/WebAPI/cdmresults/{}/dashboard"}],
"responseFormat":"CSV",
"showUserInformation":true
}'
```

### DIFF
--- a/src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
+++ b/src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
@@ -2,7 +2,7 @@ package org.ohdsi.webapi.statistic.controller;
 
 import com.opencsv.CSVWriter;
 
-import org.apache.commons.lang3.tuple.Pair;
+import org.ohdsi.webapi.shiro.TokenManager;
 import org.ohdsi.webapi.statistic.dto.AccessTrendDto;
 import org.ohdsi.webapi.statistic.dto.AccessTrendsDto;
 import org.ohdsi.webapi.statistic.dto.EndpointDto;
@@ -11,9 +11,10 @@ import org.ohdsi.webapi.statistic.dto.SourceExecutionsDto;
 import org.ohdsi.webapi.statistic.service.StatisticService;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Controller;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
 
 import javax.ws.rs.Consumes;
-import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
@@ -26,6 +27,8 @@ import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 @Controller
@@ -33,6 +36,10 @@ import java.util.stream.Collectors;
 @ConditionalOnProperty(value = "audit.trail.enabled", havingValue = "true")
 public class StatisticController {
     private StatisticService service;
+
+    public enum ResponseFormat {
+        CSV, JSON
+    }
 
     private static final List<String[]> EXECUTION_STATISTICS_CSV_RESULT_HEADER = new ArrayList<String[]>() {{
         add(new String[]{"Date", "Source", "Execution Type"});
@@ -56,11 +63,16 @@ public class StatisticController {
     @Consumes(MediaType.APPLICATION_JSON)
     public Response executionStatistics(ExecutionStatisticsRequest executionStatisticsRequest) {
         DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+        String userID = executionStatisticsRequest.isShowUserInformation() ? extractUserID() : null;
         
         SourceExecutionsDto sourceExecutions = service.getSourceExecutions(LocalDate.parse(executionStatisticsRequest.getStartDate(), formatter),
-                LocalDate.parse(executionStatisticsRequest.getEndDate(), formatter), executionStatisticsRequest.getSourceKey());
+                LocalDate.parse(executionStatisticsRequest.getEndDate(), formatter), executionStatisticsRequest.getSourceKey(), userID);
 
-        return prepareExecutionResultResponse(sourceExecutions.getExecutions(), "execution_statistics.zip");
+        if (ResponseFormat.CSV.equals(executionStatisticsRequest.getResponseFormat())) {
+            return prepareExecutionResultResponse(sourceExecutions.getExecutions(), "execution_statistics.zip");
+        } else {
+            return Response.ok(sourceExecutions).build();
+        }
     }
 
     /**
@@ -73,11 +85,16 @@ public class StatisticController {
     @Consumes(MediaType.APPLICATION_JSON)
     public Response accessStatistics(AccessTrendsStatisticsRequest accessTrendsStatisticsRequest) {
         DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-        
-        AccessTrendsDto trends = service.getAccessTrends(LocalDate.parse(accessTrendsStatisticsRequest.getStartDate(), formatter),
-                LocalDate.parse(accessTrendsStatisticsRequest.getEndDate(), formatter), accessTrendsStatisticsRequest.getEndpoints());
+        String userID = accessTrendsStatisticsRequest.isShowUserInformation() ? extractUserID() : null;
 
-        return prepareAccessTrendsResponse(trends.getTrends(), "execution_trends.zip");
+        AccessTrendsDto trends = service.getAccessTrends(LocalDate.parse(accessTrendsStatisticsRequest.getStartDate(), formatter),
+                LocalDate.parse(accessTrendsStatisticsRequest.getEndDate(), formatter), accessTrendsStatisticsRequest.getEndpoints(), userID);
+
+        if (ResponseFormat.CSV.equals(accessTrendsStatisticsRequest.getResponseFormat())) {
+            return prepareAccessTrendsResponse(trends.getTrends(), "execution_trends.zip");
+        } else {
+            return Response.ok(trends).build();
+        }
     }
 
     private Response prepareExecutionResultResponse(List<SourceExecutionDto> executions, String filename) {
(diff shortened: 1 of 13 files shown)


src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
```diff
@@ -0,0 +1,255 @@
...
+import java.io.StringWriter;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Controller
+@Path("/statistic/")
+@ConditionalOnProperty(value = "audit.trail.enabled", havingValue = "true")
```
alex-odysseus, line 31: If Audit Trail feature is not enabled the Controller is not efficient

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
```diff
@@ -0,0 +1,255 @@
...
+public class StatisticController {
+    private StatisticService service;
+
+    public enum ResponseFormat {
+        CSV, JSON
```
alex-odysseus, line 36: Both CSV and JSON output formats should be supported

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
```diff
@@ -0,0 +1,255 @@
...
+            csvWriter.flush();
+            baos.write(sw.getBuffer().toString().getBytes());
+
+            return Response
+                    .ok(baos)
+                    .type(MediaType.APPLICATION_OCTET_STREAM)
+                    .header("Content-Disposition", String.format("attachment; filename=\"%s\"", filename))
+                    .build();
+        } catch (Exception ex) {
+            throw new RuntimeException(ex);
```
alex-odysseus, line 132: We may add some logging here

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+
+    private void updateExecutionStatisticsHeader(boolean showUserInformation) {
```
alex-odysseus, line 136: Both methods to be replaced by extra two static constants!

src/main/java/org/ohdsi/webapi/statistic/dto/SourceExecutionDto.java
```diff
@@ -0,0 +1,50 @@
+package org.ohdsi.webapi.statistic.dto;
+
+import java.time.Instant;
+import java.time.LocalDate;
+
+public class SourceExecutionDto {
+    private String sourceName;
+    private String executionName;
+    private String executionDate;
+    private String userID;
```
alex-odysseus, line 10: userId

src/main/java/org/ohdsi/webapi/statistic/service/StatisticService.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Service
+@ConditionalOnProperty(value = "audit.trail.enabled", havingValue = "true")
+public class StatisticService {
+    protected final Logger LOG = LoggerFactory.getLogger(getClass());
+
+    @Value("${audit.trail.log.file}")
+    // TODO remove value
```
alex-odysseus, line 40: To be configured via properties

src/main/resources/db/migration/postgresql/V2.14.0.20231206190600__fixing_statistics_enpoint_permission_method.sql
```diff
@@ -0,0 +1,2 @@
+UPDATE ${ohdsiSchema}.sec_permission SET value='statistic:executions:post'
```
alex-odysseus, line 1: We may combine both scripts into one

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
anthonysena: Is the intent here to make the `/statistic/` end-point available based on the `audit.trail.enabled` property? Seems a little confusing to make an endpoint conditional vs. having it always available and to have it return based on the status of the configuration of the application.

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
anthonysena: When will this be added?

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
anthonysena: When will this be replaced?

src/main/java/org/ohdsi/webapi/statistic/dto/SourceExecutionDto.java
anthonysena: When do you plan to change this?

src/main/java/org/ohdsi/webapi/statistic/service/StatisticService.java
anthonysena: So will you remove this then?

src/main/resources/db/migration/postgresql/V2.14.0.20231206190600__fixing_statistics_enpoint_permission_method.sql
anthonysena: Yes, please do and make sure the prefix of the file is v2.15...

anthonysena (changes requested):

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
alex-odysseus: I personally see no problem that some of the endpoints are not available if an appropriate configuration is not at place

chrisknoll:
Docker build fails due to (presumably) a JDK 11 docker config.   We discussed and possible least-friction solution is to check the stream syntax on regex Matcher in the stream API.   Failing that, may need to find an ARM compatible JDK 8 for docker.

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
chrisknoll: I agree with @anthonysena, it is confusing because people will look at api docs and think they can call an endpoint, but it won't exist, which is confusing.

I don't see any documentation that describes that this end-piont doesn't exist when the configuration is not set.

We have rules about opt-in flags for default behavior (such as logging should be turned on/off, or auditing is enabled/distables) but the invocation of a end-point is the opt-in request of the user.   If it's not enabled, it would be better to raise an error message saying 'auditing is not enabled' then return a 404 that a conditional-endpoint does not exist.

alex-odysseus: Changed as suggested

src/main/java/org/ohdsi/webapi/statistic/controller/StatisticController.java
```diff
@@ -92,6 +98,9 @@ public Response executionStatistics(ExecutionStatisticsRequest executionStatisti
     @Produces(MediaType.APPLICATION_JSON)
     @Consumes(MediaType.APPLICATION_JSON)
     public Response accessStatistics(AccessTrendsStatisticsRequest accessTrendsStatisticsRequest) {
+    	if (!auditTrailEnabled) {
+    		throw new NotFoundException("Audit Trail functionality should be enabled (audit.trail.enabled) to serve this endpoint");
```
chrisknoll, line 104: NotFound results in a 404, correct? We said that it would be confusing if someone looked at an api and attempted to call an endpoint and it was not found.  I think we mentioned raising an error, such that it could return an HTTP 500 or some other HTTP code that represents an error other than 'not found'?

alex-odysseus: I misread your passage, corrected to the internal server error

chrisknoll (approved):

pieterlukasse:
hi @alex-odysseus , @chrisknoll sorry for the late review comment here, but I just wanted to raise a flag here. If I understand these changes correctly, the `audit.trail.enabled` setting already existed before. The behavior of this config setting was that it would ensure things got logged to some local `audit.log` file. I can imagine many installations might have turned this ON. Now, with this PR, the behavior changes significantly, by opening up new endpoints. This will happen silently, and therefore I think it is a potential security issue. Instead, I think this new feature should be controlled by a _new_ config option that is set to `false` by default. Can you please consider changing this?

alex-odysseus:
Pieter, newly added endpoints should be available for the administrators only by default @pieterlukasse

pieterlukasse:
@alex-odysseus thanks for the clarification. 
Still, I think the change might come as a surprise to some.
As an admin, I would not like endpoints (albeit admin ones) to be silently opened. 
Then again, I believe all work you are doing is still on a dev branch and nothing is officially release yet, correct?

chrisknoll:
Thanks @pieterlukasse , it's an interesting perspective.   From a WebAPI perspective, we introduce new endpoints quite a bit.  Never as a result of a hotfix release (ie x.x.1 -> x.x.2) but minor releases which add new non-backwards-breaking features to the app will bring along new endpoints (usually a new feature is exposed as a new endpoint).  The hint to the admins is that if they are installing a new version of WebAPI (minor or major version increment) they should expect new endpionts.

Now, have we been proactive about announcing which new endpoints exist?  Other than the automatic REST documentation that is generated that gives all the service endpoints, I don't think we make a habbit of itemizing each new endpint in our release notes, we usually just declare the new feature, point it to the PR, and then if you want to know the technical details, it's there.

Could we do more than that? Possibly, given the resources.   Should anyone be surprised about new endpoints in new releases?  Probably not, that should almost be the expectation.

In this specific case, a couple endpoints were added to provide reporting on the audits that are enabled via config.  So, the release notes will say something about exposing audit train reports, and pointing to the PR that implements.

pieterlukasse:
Thanks for clarifying @chrisknoll . It helps me understand better how you are dealing with this. If this is mentioned in the release notes and points to the PR, then this sounds like a reasonable approach to me. It could be useful to add a line to this PR description stating that an existing config setting is being reused/partially repurposed to do something new.

Issue #2330, opened 2023-12-18
Tools module


alex-odysseus:
### Business Requirements

A new Tools module will allow users to easily navigate to tools/applications configured by administrators. They may refer to popular internal application resources and the external ones

Some of the tools/applications may be configured as currently unavailable for the standard users 

### Design Ideas and Functional Requirements

A new tile view to be designed which can be reached by each user clicking on the "Tools" link on the left side panel
![image](https://github.com/OHDSI/WebAPI/assets/75131044/72a37ceb-b2ba-4872-a6ce-6c9556aed0db)

Administrators will be able to add new entries, edit the existing ones and delete those which are no longer required in the system. Standard CRUD operations endpoints to be added. There will be an attribute which will allow to hide an entry from the tiles view

A Tool entity should have name, description (long text), author, created timestamp, visibility flag. When the entity is updated there should be information about who and when

Standard users will be able to view the Tool entries available for visualization only

Clicking on a tile should warn with a dialogue that a user will be redirected to a URL in a separate tab

When the number of tools/applications doesn't fit into one screen there should be a way to navigate through pages

### Open Questions
- an idea to place the Tools view on the Home page instead of the block for GitHub issues is not necessary ideal as some users pass it very fast as being not very informative as of December 2023 (might be configured as optional for visualization)

@mrechkem @Sathyaraos

alondhe:
Not sure if this is the same, but in Broadsea, we have a landing page. It's a Quarto document that is generated based off of the env file.

<img width="1314" alt="image" src="https://github.com/OHDSI/WebAPI/assets/8037742/f3d2c2f1-abed-4e4c-acac-ffafc21f08be">

PR #2389, opened 2024-08-30
Tools module functionality

alex-odysseus:
Addressing #2330

- CRUD operations implemented
- a separate Permission Schema has been created


src/main/java/org/ohdsi/webapi/tool/Tool.java
```diff
@@ -0,0 +1,93 @@
...
+    @Column(name = "name")
+    private String name;
+
+    @Column(name = "url")
+    private String url;
+
+    @Column(name = "description")
+    private String description;
+    @Column(name = "is_enabled")
+    private Boolean isEnabled;
```
alex-odysseus, line 38: The property name should be renamed to 'enabled'

src/main/java/org/ohdsi/webapi/tool/ToolController.java
```diff
@@ -0,0 +1,55 @@
+package org.ohdsi.webapi.tool;
+
+import org.ohdsi.webapi.tool.dto.ToolDTO;
+import org.springframework.stereotype.Controller;
+
+import javax.ws.rs.*;
```
alex-odysseus, line 6: Imports with * should be eliminated

src/main/java/org/ohdsi/webapi/tool/converter/ToolConvertor.java
```diff
@@ -0,0 +1,82 @@
...
+        if (isNewTool) {
+            setCreationDetails(tool, currentInstant);
+        } else {
+            setModificationDetails(tool, currentInstant);
+        }
+        updateToolFromDTO(tool, toolDTO);
+        return tool;
+    }
+
+    private void setCreationDetails(Tool tool, Instant currentInstant) {
```
alex-odysseus, line 40: I don't think we need two separate methods

src/main/java/org/ohdsi/webapi/util/DateUtils.java
```diff
@@ -0,0 +1,16 @@
+package org.ohdsi.webapi.util;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class DateUtils {
```
alex-odysseus, line 7: The purpose of this class is unclear SimpleDateFormat usage should be sufficient

src/main/resources/db/migration/postgresql/V2.13.1.20231213141052__create_table_tools.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS ${ohdsiSchema}.tool (
```
alex-odysseus, line 1: We can combine all migration scripts into one

src/main/java/org/ohdsi/webapi/security/model/ToolPermissionSchema.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+public class ToolPermissionSchema extends EntityPermissionSchema {
+    private static Map<String, String> writePermissions = new HashMap<String, String>() {{
+        put("tool:%s:delete", "Delete Tool with id = %s");
```
alex-odysseus, line 11: I am afraid that the notation doesn't coincide with what is inserted by the migration scripts

src/main/resources/db/migration/postgresql/V2.13.1.20231213141052__create_table_tools.sql
chrisknoll: I'd prefer if you did this

chrisknoll: In addition, please mark these migrations attached to V2.15

chrisknoll:
Hello Everyone.  Once the migration scripts are combined and the changes @alex-odysseus requests have been implemented, I can pull down the branches to check the functionality.

src/main/java/org/ohdsi/webapi/security/model/ToolPermissionSchema.java
anthonysena: Checking this without security enabled seemed to work fine but your comment seems correct and this should be corrected before we merge this work.

anthonysena (changes requested):

src/main/java/org/ohdsi/webapi/security/model/ToolPermissionSchema.java
```diff
@@ -9,8 +9,8 @@
 public class ToolPermissionSchema extends EntityPermissionSchema {
     private static Map<String, String> writePermissions = new HashMap<String, String>() {{
         put("tool:%s:delete", "Delete Tool with id = %s");
-        put("tool:%s:put", "Update Tool with id = %s");
-        put("tool:%s:post", "Create Tool with id = %s");
+        put("tool:put", "Update a Tool");
+        put("tool:post", "Create a Tool");
```
chrisknoll, line 13: This won't work: what will happen is that when the Tool asset is deleted, all these permissions will be deleted and everyone will lose the ability to create and update a tool.

This is an example of the complication of the permission implementation we have in webapi that we'll want to resolve in 3.x or 4.x.  

For details on how the non-entity specific permissions caused issues, you can refer to this pr: https://github.com/OHDSI/WebAPI/pull/2413.

Essentially, the only permissions that go into PermissionSchema are those with entity-id context.  The general ones can't be there because deleting one asset will delete the entire (non-asset specific) set of permissions from everyone.

chrisknoll: @alex-odysseus , please provide update to this issue.

PR #2393, opened 2024-09-12
Re-introduce Caching

chrisknoll:
Implemented caching for cohort def list, concept set list, permissions.
Implementation uses ehCache 3.9.
Added test for CohortDefinitionService caching.

Notes:
This implementation diverges from the sandbox:   I decided to leverage `@CacheEvict` on methods instead of JPA Entity Listeners to simplify the implementation.

Currently getting cache info is unprotected, but cache/clear is protected.


chrisknoll:
@amarsan , would you be willing to do a review on this. I changed one of the functions of the clear cache such that the clear cache button now just clears the WebAPI cache, and the individual 'refresh' buttons next to sources will clear the caches on the specific cdms.  We could introduce a button somewhere for refreshing all cdm caches, but I seriously doubt anyone would need that function.

src/main/java/org/ohdsi/webapi/source/SourceService.java
```diff
@@ -160,10 +179,9 @@ public SourceInfo getPriorityVocabularySourceInfo() {
         return new SourceInfo(source);
     }
 
-    public void invalidateCache() {
-
-        this.cachedSources = null;
-    }
+	@CacheEvict(cacheNames = CachingSetup.SOURCE_LIST_CACHE, allEntries = true)
+	public void invalidateCache() {
+	}
```
amarsan, line 184: I see that you have @CacheEvict here, but also on the methods in the SourceController that call this method. Is that redundant?

src/main/java/org/ohdsi/webapi/source/SourceController.java
```diff
@@ -157,6 +157,7 @@ public SourceDetails getSourceDetails(@PathParam("sourceId") Integer sourceId) {
   @POST
   @Consumes(MediaType.MULTIPART_FORM_DATA)
   @Produces(MediaType.APPLICATION_JSON)
+	@CacheEvict(cacheNames = SourceService.CachingSetup.SOURCE_LIST_CACHE, allEntries = true)
```
amarsan, line 160: Does this invalidate the cache even if this is called when security is disabled?

src/main/java/org/ohdsi/webapi/source/SourceService.java
chrisknoll: I don't think it's redundant, but I did have trouble understanding where the line between the service and the controller was being drawn, so there were certain places where cache evict happens at the controller level and the service level in others.   In this function, there was a service-level function to invalidate the cache so that's where the evict annotation applies here.  I think the reason why it's at the service level is that the service components are typically the ones that are dependency-injected into other components, so if some other component needs to signal a cache invalidation for some reason, it would call out to this service method.

I'm glad you're poking at this because it is revealing some of the inconsistency and coding confusion that we have in the codebase that we could try to address when we have more flexibility to change things (Ie 3.x and 4.x)

src/main/java/org/ohdsi/webapi/source/SourceController.java
chrisknoll: Yes, the caching is independent of security being enabled.

amarsan: Just so I understand how this CacheEvict works...does it execute *after* the method executes, and only if it executes without throwing an exception?

chrisknoll: It will fire after the method, and if the method throws an exception, it will not evict.

It can be set to evict the cache before invoking the method by using the `beforeInvocation` option on the annotation.

In this case, I think we want the default behavior, if it fails to create the source for some reason, then there is no need to cache evict.

amarsan (approved):

PR #2396, opened 2024-09-16
Fix for compile error seen for Eclipse (issue 2390) (#2395)

chrisknoll:


anthonysena (approved):

mvn
anthonysena, line 1: I'd suggest removing this empty file - not sure if this was intentional or not.

Issue #2318, opened 2023-11-08
Add metadata to Concept Set Items


alex-odysseus:
### Functional Requirements

When a Concept is added to a Concept Set as a Concept Set Item there should be an annotation automatically added which encompasses any contextual parameters being used to display a Concept

There are multiple screens from which it is possible to add a Concept into a Concept Set, the metadata might be the following:

- Search listing (the search term itself, advanced options (Domains, Vocabularies), filters)
- Concept Details (previous page context)
- Related Concepts (previous page context, filters)
- Concept Hierarchy (previous page context, filters)
- Concept Set - Included Concepts (previous page context, filters)
- Concept Set - Included Source Codes (previous page context, filters)
- Concept Set - Recommend (previous page context, filters)
- Concept Set - Import - Concept Identifiers (previous page context, filters)
- Concept Set - Import - Source Codes (previous page context, filters)
- Concept Set - Import - Concept Set (previous page context)
- Concept Set - Import - Repository (previous page context)
- any other places?!

The previous page context to be assessed if possible to add as metadata to a Concept Set Item annotation

A new tab "Metadata" (naming to be defined) should be added after the "Compare" tab to display annotations for each Concept Set Item, by default an annotation after being added will contain a templated text which can be edited after

The annotations listing should contain information when an annotation was added, who added it, the annotation text itself (editable) and when it was updated and by whom

The feature should be optional, effective by an explicit application property switch

### Opened Questions

- there should be no intention to extend Circe with the Concept Set Item annotations so that this information is transferred between WebAPI / ATLAS installations

chrisknoll:
If you could add the motivation to including this feature (ie: the problem it's trying to solve) that would be helpful.

PR #2403, opened 2024-10-29
Extending Concept Set Items with Annotations

alex-odysseus:
Addressing #2318

A new Concept Set Annotation entity has been added being associated with a Concept Set to which the Concept Set Annotations relate to

When Concept Set is copied its Concept Set Annotations are preserved so that it will be known that some of the Concept Set Annotations were originated from a particular Concept Set of a specific Version

When a Concept Set deleted it is deleted together with the Concept Set Annotations

Concept Set Annotations can be deleted by the Administrators only or by those who have a specific permission assigned


src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -857,4 +876,153 @@ private ConceptSetVersion saveVersion(int id) {
...
+                ConceptSetAnnotation conceptSetAnnotation = new ConceptSetAnnotation();
+                conceptSetAnnotation.setConceptSetId(conceptSetId);
+                try {
+                    AnnotationDetailsDTO annotationDetailsDTO = new AnnotationDetailsDTO();
+                    annotationDetailsDTO.setId(newAnnotationData.getId());
+                    annotationDetailsDTO.setConceptId(newAnnotationData.getConceptId());
+                    annotationDetailsDTO.setSearchData(newAnnotationData.getSearchData());
+                    conceptSetAnnotation.setAnnotationDetails(mapper.writeValueAsString(annotationDetailsDTO));
+                } catch (JsonProcessingException e) {
+                    throw new RuntimeException(e);
```
alex-odysseus, line 911: Let's add a log entry

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -857,4 +876,153 @@ private ConceptSetVersion saveVersion(int id) {
...
+        return true;
+    }
+    private void removeAnnotations(int id, SaveConceptSetAnnotationsRequest request){
+        if (CollectionUtils.isNotEmpty(request.getRemoveAnnotation())) {
+            for (AnnotationDTO annotationDTO : request.getRemoveAnnotation()) {
+                this.getConceptSetAnnotationRepository().deleteAnnotationByConceptSetIdAndConceptId(id, annotationDTO.getConceptId());
+            }
+        }
+    }
+    private ConceptSetAnnotation copyAnnotation(ConceptSetAnnotation sourceConceptSetAnnotation, int sourceConceptSetId, int targetConceptSetId){
```
alex-odysseus, line 933: Please move it down below the public method where it is used

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -857,4 +876,153 @@ private ConceptSetVersion saveVersion(int id) {
...
+        if (conceptSetAnnotation != null) {
+            getConceptSetAnnotationRepository().deleteById(id);
+            return Response.ok().build();
+        } else throw new NotFoundException("Concept set annotation not found");
+    }
+
+    @PUT
+    @Path("/update/{id}/annotation")
+    @Produces(MediaType.APPLICATION_JSON)
+    public AnnotationDTO updateConceptSetAnnotation(@PathParam("id") final int id, AnnotationDTO annotationDTO) throws IOException {
```
alex-odysseus, line 1018: This method is obsolete at this moment, if in the future it will be necessary to add/update a comment to a specific Concept Set Annotation it might be added

src/main/resources/db/migration/postgresql/V2.14.0.202407161000__conceptset_annotation.sql
```diff
@@ -0,0 +1,46 @@
+CREATE SEQUENCE ${ohdsiSchema}.concept_set_annotation_sequence;
```
alex-odysseus, line 1: Let's combine one migration script

src/main/resources/i18n/messages_en.json
```diff
@@ -1786,7 +1791,8 @@
         "includedConcepts": "Included Concepts",
         "includedSourceCodes": "Included Source Codes",
         "versions": "Versions",
-        "messages": "Messages"
+        "messages": "Messages",
+        "metadata": "Metadata"
```
alex-odysseus, line 1795: To be removed

src/main/java/org/ohdsi/webapi/conceptset/ConceptSet.java
```diff
@@ -36,7 +37,7 @@
  *
  * @author fdefalco
  */
-@Entity(name = "ConceptSet")
+@Entity
```
chrisknoll, line 40: Please revert this as we have a convention to name the entity, you can see this in CohortDefinition, Source, etc.

src/main/java/org/ohdsi/webapi/conceptset/ConceptSetGenerationInfo.java
```diff
@@ -21,7 +21,7 @@
  * @author Anthony Sena <https://github.com/ohdsi>
  */
 
-@Entity(name = "ConceptSetGenerationInfo")
+@Entity
```
chrisknoll, line 24: Same as above, please revert this change.

src/main/java/org/ohdsi/webapi/conceptset/ConceptSetItem.java
```diff
@@ -29,7 +29,7 @@
  * @author fdefalco
  */
 
-@Entity(name = "ConceptSetItem")
+@Entity
```
chrisknoll, line 32: Same as above, we name Entities.

src/main/java/org/ohdsi/webapi/security/model/ConceptSetAnnotationPermissionSchema.java
```diff
@@ -0,0 +1,25 @@
...
+    private static Map<String, String> writePermissions = new HashMap<String, String>() {
+        {
+            put("conceptset:annotation:%s:delete", "Delete Concept Set Annotation with ID %s");
+        }
+    };
+
+    private static Map<String, String> readPermissions = new HashMap<String, String>() {
+        {
+        }
+    };
```
chrisknoll, line 20: This is odd to me: 

Shouldn't the writePermissions contain all permissions related to creating annotations and the read have the ones related to reading annotations?

src/main/java/org/ohdsi/webapi/service/ConceptSetService.java
```diff
@@ -25,6 +26,11 @@
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
```
chrisknoll, line 33: Just might want to check this import (collections4 is something I'm not familiar with).  is there an existing package we depend on that has the function in CollectionUtils that you are using?

Just trying to keep extra dependencies down....when we get to the 3.x line we're going to be combing through the code locating redundant dependcies (ie: libraries that support string manipulation will be de-duplicated)  So if we can remove a source of additional duplication, we should try to do it here.

src/main/resources/db/migration/postgresql/V2.14.0.202407161000__conceptset_annotation.sql
chrisknoll: Yes, let's combine into 1 script and also this is targeting v2.15 so if we do create 1 migrations cript let's just make a new one with all the others combined, call it V2.15 ...

src/main/resources/db/migration/postgresql/V2.14.0.202407261000__add_conceptset_version_to_annotation.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE ${ohdsiSchema}.concept_set_annotation ADD concept_set_version VARCHAR;
```
chrisknoll, line 1: Are we sure VARCHAR is the correct type here?  That will allow 2GB of text

chrisknoll:
Hi, Everyone. I added comments and if you could address those I will work on pulling the branch down to test.

Specifically, the migration scripts being combined and re-labeled to 2.15 is soemthing I'd like to have changed before pulling it down, because otherwise I'd have to manually roll back the stuff in the 2.14 migrations and then let them re apply from the 2.15 migration, and it gets to be a headache.   Once that's done I can play-test the functionality locally and provide feedback.

chrisknoll:
Thanks for addressing those requested changes. I'll pull it down and take it for a spin.

src/main/java/org/ohdsi/webapi/security/model/ConceptSetPermissionSchema.java
```diff
@@ -11,17 +11,21 @@ public class ConceptSetPermissionSchema extends EntityPermissionSchema {
     private static Map<String, String> writePermissions = new HashMap<String, String>() {{
         put("conceptset:%s:put", "Update Concept Set with ID = %s");
         put("conceptset:%s:items:put", "Update Items of Concept Set with ID = %s");
+        put("conceptset:*:annotation:put", "Create Concept Set Annotation");
```
chrisknoll, line 14: These should not be * permissions. What happens is that when you delete the entity, it finds the write permission schema and deletes it, thus deleting the wildcard permission' from all users, and then preventing access to the endpiont for all users.

However, this is what happens when you extend the 'CommonEntity' which you didn't do in this case, however, ou are using a permissionSchema as if it was a common entity, so there may be a little confusion here.

But, in any case, we shouldn't see * permissions in any writePermission schema because these permission schemas are targeting a specific entity ID.

chrisknoll (changes requested):

chrisknoll:
Hi, I am getting an error when I restore to a prior version of the DB and the migration scripts were applied.  

Steps to reproduce:

1. Click Create concept Set (repository concept set)
2. Name it
3. Save it (empty concept set)
4. Observe error:  There was an error saving your concept set'
5. Close editor with X to return to concept set list
6. Observe the new concept set was actually created.
7. Open Concept set
8. Search for a concept
9. Add concept to concept set
10. Try to save again: Observe same error 'There was an error saving your concept set'
11 close and re-open the concept set, and the concept set seems to be saved.

Note: annotation tab is 'disabled' in that clicking it does not swatch tab.
Also, there are no errors in the browser console, and no exceptions seen on the server log, making it hard to diagnose the issue.

chrisknoll:
I've confirmed that I do not get this error when switching to master/master (atlas/webapi) so there is something related to this branch leading to the described error.

Issue #2296, opened 2023-07-08
Atlas 2.13.0 returns wrong values for RC/DRC and PC/DPC


TomWhite-MedStar:
### Expected behavior
When you use Search for a concept and select a data source, Atlas will show the correct counts for RC/DRC and PC/DPC.

### Actual behavior
When we first select a new datasource (let's call it source A) after doing a Search (or anywhere that RC and PC are shown - including in global concept sets or those within cohort definitions), the correct query will be run (confirmed by reviewing developer tools for API calls and monitoring queries issued to database), but once the cursor stops spinning, Atlas will show wrong values (sometimes 0, sometimes wrong values) for RC and PC for those concepts.  If you then switch to another data source (let's call it source B) and then switch back to source A, the query will not be re-run (confirmed by developer tools and database logs), but Atlas _will_ then show the correct RC and PC counts for those concepts.  Those correct values are stored in the results.achilles_result_concept_count table - but Atlas displays the wrong values until one switches away from and then back to the desired data source.

This has been an issue since 2.13.0 was released.

Thus, it appears that there is an internal caching issue.  The right WebAPI command is called to trigger the query to get the RC and PC values, but those final values are not then available when the DC/PC results are displayed until the user switches away from the data source and then returns to it.

### Steps to reproduce behavior
We have confirmed this repeatedly on our 2.13.0 instance, which happens to be connected to Databricks.  Since we can confirm that the appropriate queries against Databricks are run (by reviewing its query logs) and that the right row counts are returned, I doubt this a Databricks-specific issue.  Atlas also worked fine for Databricks prior to version 2.13.0.

### Update on 1/1/2024

With the additional information, we think the steps to reproduce are as follows:

1. Clear `cdm_cache` (truncate) so that there are no cached record counts in the cache table.
2. Perform a vocab search (or open a concept set that will fetch included concepts) which will trigger a row count lookup on server
3. Observe that the record count REST endpoint is invoked, and this should lead to a record fetch over to the CDM source's acchilles_results_concept_count.
4. Observe that the RC/DRC columns should appear as 0 (provided this bug is happening as we expect).  We should check the REST payload to see what the values were in the response. If they are 0, then the UI is presenting the info that it was given. If it is not zero, then the UI is presenting zeros when the payload has actual values for some reason.  **(See below:  Tom reports that the counts did come back as 0 on the first request)**
5. Switch the record count datasource to another source.
[...]

PR #2405, opened 2024-11-07
Bug Fix: Return correct count values for RC/DRC and PC/DPC

amarsan:
Closes https://github.com/OHDSI/WebAPI/issues/2296

The issue is that we were modifying an array of concept IDs when going to the DB for their counts and caching them. I wrote an integration test cover this scenario.

I had some trouble getting the integration tests to run in my environment, and ended up having to make some changes to unrelated files in order to get the project to build.


pom.xml
```diff
@@ -769,6 +769,12 @@
       <artifactId>json-smart</artifactId>
       <version>2.4.9</version>
     </dependency>
+    <dependency>
+      <groupId>net.minidev</groupId>
+      <artifactId>asm</artifactId>
+      <version>1.0.2</version>
+      <scope>test</scope>
+    </dependency>
```
amarsan, line 777: I needed to add this dependency in order to get the project to build and run in my test runner.

chrisknoll: should be fine, the scope is test so won't be part of the final bundle.

src/main/java/org/ohdsi/webapi/cdmresults/service/CDMCacheService.java
```diff
@@ -145,9 +145,9 @@ private void cacheRecordsById(Source source, List<Integer> ids) {
       PreparedStatementRenderer psr = getPartialPreparedStatementRenderer(source, idsSlice);
       Set<Integer> cachedIds = loadCache(source, psr, mapper, jdbcTemplate);
       // in this batch, need to save any concepts that were not found when loading cache
-      idsSlice.removeAll(cachedIds);
-      if (!idsSlice.isEmpty()) { // store zeros in cache
-        List<CDMCacheEntity> zeroConcepts = idsSlice.stream().map(id -> {
+      List<Integer> notFoundIds = new ArrayList<Integer>(CollectionUtils.subtract(idsSlice, cachedIds));
+      if (!notFoundIds.isEmpty()) { // store zeros in cache
+        List<CDMCacheEntity> zeroConcepts = notFoundIds.stream().map(id -> {
```
amarsan, line 150: This is the bug fix (all the other files are for adding the integration test). The `idsSlice` variable is a subList of the `ids` that were passed into this method. In Java, an array subList references the parent list that it came from, and thus any structural changes to it affect the parent list. We were removing the `cachedIds` from slice, which also removed them from the parent `ids` list. After this method is called, the calling method uses the `ids` list to determine what to return the front end, so what was happening was that any counts we had to get from the DB were not being returned.

chrisknoll (approved):

src/main/java/org/ohdsi/webapi/cdmresults/service/CDMCacheService.java
chrisknoll: Very good explanation. Thank you for researching this and providing a patch.

Issue #2348, opened 2024-02-13
Non-Standard Concepts mapped to Standard Concepts to be replaced by Standard Concepts by the "Optimize" functionality of a Concept Set 


alex-odysseus:
### Business Requirements

If a Concept Set contains all non-Standard Concepts that are comprised of a Standard Concept they should be replaced with the Standard Concept as the result of Concept Set optimization routine 

### Design Ideas and Functional Requirements

An initial idea is:

An additional flag to be added on the "Concept Set Optimization" screen which will extend the existing optimization business logic with a possibility to replace non-Standard Concepts being mapped to the Standard Concepts by the appropriate Standard Concepts

Based on Anthony's comment the following solution design is proposed:

- a new button "Resolve mappings" to be added in between the "Optimize" and the Delete buttons, it is enabled only if at least one non-standard Concept is present as a Concept Set Item
![image](https://github.com/OHDSI/WebAPI/assets/75131044/ec7519e2-5b40-44f4-812d-15d7b0fe38e4)
- after clicking on the button there will be a popup window displayed with the title "Non-standard Concepts Mapping Resolution" which will contain three visually distinguished blocks
![image](https://github.com/OHDSI/WebAPI/assets/75131044/f62eec2c-28cb-49f3-b99b-cbab135d8755)
- 'Concept Set with resolved mappings' will contain all Standard and Classification Concepts which were present before clicking on the 'Resolve mappings' button
-- additionally it will contain Standard Concepts counterparts (without duplicates) for those non-Standard Concepts where the mapping is straightforward
-- 'Remove Selected' will allow to remove some of the items which may not be needed
- 'Non-standard Concepts with multiple Standard Concepts mappings' will be grouped in a way that Standard Concepts with 'Non-standard to Standard map (OMOP)' will be followed by non-standard Concepts for which the ambiguity is present
-- a User will be able to select both Standard and non-Standard Concepts and add them to the 'Concept Set with resolved mappings' block after selecting the appropriate ones and clicking on "Add to Concept Set" button
-- by means of the 'Preview...' button it will be possible to preview a Concept Set and how it will look like after adding the selected Concepts to 'Concept Set with resolved mappings' (additionally displaying 'Included Concepts' and 'Included Source Codes' listings)
- 'Original Concept Set' displays the version of Concept Set Items before any mapping resolution and User's manipulations after
[...]

PR #2407, opened 2024-11-12
Non-Standard to Standard Concepts mapping API

alex-odysseus:
Addressing #2348

Adding an endpoint to get Standard Concepts into which non-Standard Concepts can be mapped

The Related Concepts SQL statement has been used as the basis


src/main/resources/resources/vocabulary/sql/getRelatedStandardMappedConcepts.sql
anthonysena, line 1: This query seems a bit off - why does it require a GROUP BY when there is no aggregation done in the outermost query? Should the `1 as RELATIONSHIP_DISTANCE` be summed?

anthonysena (commented):
Had a question regarding the query you are using to get the related concepts but otherwise seems good.

src/main/resources/resources/vocabulary/sql/getRelatedStandardMappedConcepts.sql
chrisknoll: It could work the same as a select distinct...

chrisknoll (approved):

PR #2439, opened 2025-04-23
Dependabot updates

chrisknoll:
This PR groups the individual dependabot updates into a single PR.

One point of note: the jackson databind from 2.12.7 to 2.12.7.1 is not a global (jackson-wide) version change, so in that case we're not using the POM variable to assign the version. We have to explicitly add the version in that case.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,12 +28,12 @@
 
     <commons-fileupload.version>1.5</commons-fileupload.version>
 
-    <circe.version>1.12.0</circe.version>
+    <circe.version>1.12.1</circe.version>
     <jersey.version>2.14</jersey.version>
     <SqlRender.version>1.19.1</SqlRender.version>
     <hive-jdbc.version>3.1.2</hive-jdbc.version>
     <pac4j.version>4.5.5</pac4j.version>
-    <jackson.version>2.12.7.1</jackson.version>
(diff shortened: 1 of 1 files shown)


chrisknoll:
Tagging @oleg-odysseus to review dependency updates for any concerns.

oleg-odysseus:
Hi @chrisknoll , I`ve started a review on one of the lines of code once you`ve assigned me to the PR, but looks like it did not trigger any notification to you. My comment was the following (for this line, jackson-databind dependency,  <version>2.12.7.1</version>): "Is it expected to be different from ${jackson.version} which is 2.12.7? If not then I would propose to keep the single version for all the jackson dependencies. I`ve tested this branch on my local build with 4 core analyses - works ok, at first glance all those version changes do not cause any observable issues and I assume are OK to merge squashed."

chrisknoll:
Yes, that's what I put into the PR description:  Putting it in the `jackson.version` property, it makes all jackson dependencies 2.12.7.1, but the 2.12.7.1 only applies to jackson-databind (but not jackson-core, for example). Therefore, we had to specify a dependency-specific version for that one element, and let the jackson.version apply in the other cases.  

Eventually when jackson goes to something like 2.12.8, we can probalby go back to pulling the version from the property for all cases.

PR #2458, opened 2025-08-07
Temp schema prefix to be present in the cohortTable argument in CCQueryBuilder

oleg-odysseus:
…

### DIFF
--- a/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
+++ b/src/main/java/org/ohdsi/webapi/cohortdefinition/GenerateCohortTasklet.java
@@ -132,7 +132,7 @@ public class GenerateCohortTasklet extends CancelableTasklet implements Stoppabl
       final Integer sourceId = Integer.valueOf(jobParams.get(SOURCE_ID).toString());
       final Source source = sourceService.findBySourceId(sourceId);
 
-      final String cohortTable = jobParams.get(TARGET_TABLE).toString();
+      final String cohortTable = String.format("%s.%s", SourceUtils.getTempQualifier(source), jobParams.get(TARGET_TABLE).toString());
       final String sessionId = jobParams.get(SESSION_ID).toString();
 
       final String tempSchema = SourceUtils.getTempQualifier(source);


chrisknoll:
@oleg-odysseus , are you sure this is correct, i'm looking in other parts of the code and i see thigns like this:

```
    final Source source = sourceService.findBySourceId(sourceId);
    final String resultsQualifier = SourceUtils.getResultsQualifier(source);
    final String tempQualifier = SourceUtils.getTempQualifier(source, resultsQualifier);
    final String translatedSql = SqlTranslate.translateSql(sql, source.getSourceDialect(), null, tempQualifier);
    Arrays.stream(SqlSplit.splitSql(translatedSql)).forEach(jdbcTemplate::execute);
```

In other words, it handles replacing the temp table reference with the temp schema in the translateSql step.

In your PR, you have this code:

```
final String cohortTable = String.format("%s.%s", SourceUtils.getTempQualifier(source), jobParams.get(TARGET_TABLE).toString());
```

This presumes that the temp table is going to require a temp schema qualifier, which isn't always the case (like you don't need that for postgres).

So, the crux of my qustion is: does this work correctly with both 'virtual temp tables' that need a temp schema vs. other dialects where you can reference the temp table without a schema (such as #myTempTable)

chrisknoll:
Hi @oleg-odysseus . I don't think we need to make sure that there's a temp placeholder, I think sqlTranslate will translate temp tables into temp-schema prefixed...I think the idea behind putting into sqlTranslate (vs. sqlRender) is that translate will transform from a source dialect to target. So `#myTempTable` gets translated to `myTempTable` in postgres, `{session_id}_myTempTable` for virtual temp tables, etc.

We don't have to dive to deep on this, if you are certain this will resolve the immeate issue, that's fine, but as we move into Atlas 3.x we should probalby follow a standard convention for handling temp tables in our queries.

oleg-odysseus:
Hi @chrisknoll . To confirm on the fix - I ran the cohort generation on master without the fix yesterday (postgres) and it failed. And having cherry-picked the fix into it - the analysis went fine. Double checked today on a non-cached cohort 
image
Best Regards.

PR #2466, opened 2025-10-19
Preserve supplied oracle temp schema in SqlRenderService

watilde:
Fix: Preserve custom oracleTempSchema instead of overwriting with default

- Only set default placeholder when oracleTempSchema is not provided
- Add test case for custom schema preservation


watilde:
Let me close this and investigate as I see a difference between my local and CI.

Issue #2461, opened 2025-09-02
Snowflake Authentication from userid/password to key-pair


davidhcar:
Snowflake recently announced that they are discontinuing the usage of userid/password authentication method in the coming months, the userid/password connections will not be supported from November 1st, 2025. Our organization is moving towards Key-pair based authentication that would avoid using password in the connection string for Snowflake connections.

Requesting the Snowflake connector to support the key-pair authentication in place of userid/password feature with Atlas WebAPI.

Announcement: https://docs.snowflake.com/en/user-guide/security-mfa-rollout?_ga=2.164037754.470643457.1756141295-382733811.1692124158

Key-Pair Authentication doc at snowflake: https://docs.snowflake.com/en/user-guide/key-pair-auth#supported-snowflake-clients

@chrisknoll @anthonysena

davidhcar:
@konstjar Appreciate any update on this, it sounds like no code changes necessary to make this work. if so, would you please provide instructions on how to configure?

konstjar:
@davidhcar 

There are 2 alternatives to username password:

1. Key-Pair works out of box right now. You can just follow this instructions:
https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-configure#using-key-pair-authentication-and-key-rotation

2. PAT. Its support was added in the recent JDBC driver version and requires update. 
There is PR for this: https://github.com/OHDSI/WebAPI/pull/2464

davidhcar:
Thanks @konstjar ! Where do we put the privatekey file, is there a specific location/folder path in the source code or somewhere else? Please let me know. Thank you for looking into this.

konstjar:
@davidhcar 
As I recall, JDBC Driver version 3.13.x got key-pair support back in 2022. 
WebAPI v2.13 has Snowflake JDBC driver in [pom.xml](https://github.com/OHDSI/WebAPI/blob/v2.13.0/pom.xml#L1824C35-L1824C42) with v3.13.22 which supports key-pair authentication.

davidhcar:
@konstjar What would be the krb_auth_method value in Source table. The default is PASSWORD and I think it should be different for Key/Value pair authentication.

konstjar:
@davidhcar I do not think krb_auth_method is used in the Snowflake connection flow. So, you can set anything.
Regarding `private_key_file_pwd`, you may avoid put this parameter directly in the connection string, but use environment variable instead. That can resolve your special characters issue and secure way to store the password. 

```
[...]

PR #2469, opened 2025-10-24
Added snowflake key-pair authentication support

davidhcar:


chrisknoll:
Hi, @davidhcar ,

Thanks so much for contributing this PR. Would you be able to provide a little description of what's involved in this PR?

I scanned through the PR and saw the code changes, which I think I understand what is being done there (such as looking for the driver associated with the JDBC url, checking for properties, etc) but it isn't clear the motivation or purpose of doing it. 

Could you provide a brief summary of what the problem you solved was, and a small description of the approach you to to solve?

Thank you!

davidhcar:
Hi @chrisknoll,
This is related to the issue https://github.com/OHDSI/WebAPI/issues/2461 the current code does not support storing the private key password in a secure way, so wanted to leverage the existing username/password fields. But I did not mean to create this PR yet, sorry about that.

PR #2473, opened 2025-12-14
add jar build

p-hoffmann:
This pull request modernizes the OHDSI WebAPI project by switching its primary packaging from a deployable WAR to an executable Spring Boot JAR with embedded Tomcat, and updates the build, Docker, and documentation accordingly. The changes simplify deployment (no external app server required), streamline configuration, and improve the developer experience. WAR packaging is still available via a Maven profile for legacy use cases.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,9 +4,9 @@ name: CI
 # Run in master and dev branches and in all pull requests to those branches
 on:
   push:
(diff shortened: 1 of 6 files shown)


konstjar:
The change will break existing Dockerfile and build procedure. It should be updated:
https://github.com/OHDSI/WebAPI/blob/webapi-3.0/Dockerfile#L29

PR #2486, opened 2026-01-24
fix docker build

p-hoffmann:


chrisknoll:
I think we want this PR to go into my external config feature branch, so that we can collect all the changes together. I think if we merge this PR into webapi-3.0 it will take my incomplete work.

If I have this correctly, I think you should reassign this PR to go to my branch (upgrade/external-config).  You had it this way initially, not sure what it was changed.

p-hoffmann:
I've created this PR only to run the CI. The CI will only be triggered if the PR has webapi-3.0 or master as a base.

PR #2498, opened 2026-03-25
Fix anonymous auth, login error handling, and data for CI tests

p-hoffmann:
This pull request updates the test and integration Docker environments and related SQL setup scripts to migrate from the legacy JDBC authentication configuration to a new database authentication mechanism. It also improves test coverage for JWT authentication and role/permission assignments, and updates Postman integration tests to match new API response formats.


src/main/java/org/ohdsi/webapi/security/authc/JwtAuthConfig.java
```diff
@@ -264,36 +256,29 @@ public SecurityFilterChain apiChain(HttpSecurity http,
...
-    } else {
-      // DisabledSecurity: permit all requests, no JWT validation
-      http.authorizeHttpRequests(auth -> auth.anyRequest().permitAll());
-    }
-
-    // Fallback to anonymous if JWT not present
-    http.anonymous(anon -> anon
+        .httpBasic(AbstractHttpConfigurer::disable)
+        // Public endpoints that don't require authentication
+        .authorizeHttpRequests(auth -> auth
```
chrisknoll, line 264: We shouldn't need to do this:  we are configuring our authentication endpints in separate security filter chains with a @order tag to have them match before the generic 'everythign else' rest paths.  We don't want to maintain a list in the JwtSecurityFilterChain that knows about the other security filter chains (it's a maintenance nightmare).

src/main/java/org/ohdsi/webapi/security/authc/JwtAuthConfig.java
```diff
@@ -264,36 +256,29 @@ public SecurityFilterChain apiChain(HttpSecurity http,
...
+            .requestMatchers("/info", "/auth/**", "/user/login/**", "/user/oauth/**",
+                             "/.well-known/**", "/actuator/**").permitAll()
+            .anyRequest().authenticated())
+        // Return 401 JSON for unauthenticated requests
+        .exceptionHandling(ex -> ex.authenticationEntryPoint((req, resp, authEx) -> {
```
chrisknoll: I think there is some confusion with your understanding of the intent of the security in webAPI 3.0:  you can access methods unauthenticated, and when you do, you get an annonymous user identity which has permissions to do things (or can be configured with no permissions for anything).  So, we only need to set authentication required on the login endpoints filter chains.

this filter chain can accept a JWT (and will validate) or nothing in the auth header (and will assume anonymous identity).

src/main/java/org/ohdsi/webapi/security/spring/SpringSecurityConfig.java
```diff
@@ -10,21 +9,12 @@
...
-  @ConditionalOnProperty(name = "security.provider", havingValue = "AtlasRegularSecurity")
-  @EnableMethodSecurity
-  public static class MethodSecurityConfig {
-
-    @Bean
-    public MethodSecurityExpressionHandler methodSecurityExpressionHandler(
-        AuthorizationService authorizationService) {
-      return new WebApiMethodSecurityExpressionHandler(authorizationService);
-    }
+  @Bean
```
chrisknoll, line 15: This is a strange added lines block because i'm looking at my branch and this content is already included in my branch, so not sure why this is looking like added lines.

src/test/java/org/ohdsi/webapi/test/WebApiIT.java
```diff
@@ -129,6 +138,25 @@ public static void before() throws IOException {
...
 
+    @Before
+    public void setUpAuthentication() {
+        // Ensure anonymous user has admin role for test permissions
+        jdbcTemplate.execute(
+            "INSERT INTO public.sec_user_role (id, user_id, role_id, origin) " +
+            "SELECT nextval('public.sec_user_role_sequence'), -1, 2, 'SYSTEM' " +
+            "WHERE NOT EXISTS (SELECT 1 FROM public.sec_user_role WHERE user_id = -1 AND role_id = 2)");
+
+        // Generate a JWT for the anonymous user so HTTP requests are authenticated
```
chrisknoll, line 149: This is good that you're handling the test case, thanks for updating this.

I don't think we need to make an annonymous JWT token because you will assume an anonymous identity when you connect to the webAPI service.  However, it is definitely the case where we would need to insert permissions for the anonymous user into the DB so it has 'full permissions' to do all the tests.   Not sure if we need to do that on a per-test basis or if we should set up the test DB once for the entire test db and have all anonymous test examples run with full permissions.  I don't know the right path here because we may want to run certain tests based on certain user permissions, but in that case I'd say we'd set up a series of test users with their associated permissions (and in that case, you would need to mint a JWT). So maybe for this test, you should register a new user for your test, grant it permisisons/roles and then run the test. That way we aren't overloading the notion of an 'anonymous user' as 'someone who logs in anonymously.

chrisknoll:
Also, is this PR intended to go into webapi-3.0 or is it to merge into my authz-update branch?

docker/auth-test/setup-test-users.sql
```diff
@@ -6,8 +6,8 @@ INSERT INTO webapi.sec_role (id, name, system_role) VALUES (1001, 'Atlas users',
 INSERT INTO webapi.sec_role (id, name, system_role) VALUES (1002, 'Moderator', true) ON CONFLICT (id) DO NOTHING;
 
 -- Anonymous user (required for public endpoints)
-INSERT INTO webapi.sec_user (id, login, name) VALUES (1, 'anonymous', 'anonymous') ON CONFLICT (id) DO NOTHING;
-INSERT INTO webapi.sec_user_role (id, user_id, role_id, origin) VALUES (1, 1, 1, 'SYSTEM') ON CONFLICT (id) DO NOTHING;
+INSERT INTO webapi.sec_user (id, login, name) VALUES (1, 'anonymous', 'anonymous') ON CONFLICT DO NOTHING;
```
chrisknoll, line 9: Looking further on this, isn't this already handled in the baseline migration found here:

https://github.com/OHDSI/WebAPI/blob/a123a4f80a72ce3b84e1428128733eab497a82b2/src/main/resources/db/migration/postgresql/B3.0.0__webapi_baseline.sql#L2043

Also, the anonymous id is hard-coded to -1 (to avoid the conflict from the 2.x user IDs in case they exist).  Here's the part of the baseline flyway:

```
-- Anonymous User and anonymous role:
INSERT INTO ${ohdsiSchema}.sec_user (id, login, name, origin)
VALUES (-1, 'anonymous', 'Anonymous', 'SYSTEM');

INSERT INTO ${ohdsiSchema}.sec_role (id, name, system_role)
VALUES (-1, 'anonymous', false);

INSERT INTO ${ohdsiSchema}.sec_user_role (id, user_id, role_id, origin)
VALUES (nextval('${ohdsiSchema}.sec_user_role_sequence'), -1, -1, 'SYSTEM');
```

src/main/java/org/ohdsi/webapi/security/authc/JwtAuthConfig.java
p-hoffmann: Yes, these were temporary changes to get the tests running until I've figured out how the proper fix works

PR #2499, opened 2026-03-27
fix SpEL syntax errors

p-hoffmann:
This pull request makes minor adjustments to security annotations in the backend service code, primarily correcting permission checks in method-level security expressions.


chrisknoll (commented):
Looks good, tho i think there's some open questions about the permissions such as: can you generate without needing read access to a cohort definition?

In any event, glad you found these typos and shocked I didn't catch them the first time around as I was testing the cohort generation.

PR #2502, opened 2026-04-07
fix tests and other issues

p-hoffmann:


chrisknoll:
Thank you for these fixes, i relied a bit too heavily on LLM on a couple things which I think I should revisit to see if the permission checks make sense for some of those cases.  I see this is a draft, if there is more work you want to do, let me know otherwise, I can merge it in.

src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -618,31 +618,40 @@ public Page<CcShortDTO> getPage(final Pageable pageable) {
         UserAuthorizations authz = authorizationService.getCurrentUserAuthorizations();
         boolean globalRead = authorizationService.isPermitted("read:cohort-characterization");
         boolean globalWrite = authorizationService.isPermitted("write:cohort-characterization");
+        boolean needsFiltering = !defaultGlobalReadPermissions && !globalRead;
```
chrisknoll, line 621: Good change, paging based on db vs in memory is a cool optimization.

The defaultGlobalReadPermissions function is something that we're going to revisit in 3.0.  But instead of an app-wide setting, it may become a paramater passed from the client to decide if they should see the things they don't have read-access to.  For example, creating a cohort characterization and getting a list of cohort definitions to include in your characterization, i don't see a reason to not show cohort definitions to characterize even if you can't see the actual cohort definition.   Conversely, the cohort definition list that lets you view and edit does make sense to restrict seeing only things you can at least read.

I have to check in with @rkboyce on this function, it may be the case we want an application-wide option to hide non-readable assets from lists.  If so, we should just change this flag to requireReadAccess and then if that is set, then we do the additional filtering of lists when requireReadAccess is TRUE.

pom.xml
```diff
@@ -280,6 +280,10 @@
       <name>repo.ohdsi.org</name>
       <url>https://repo.ohdsi.org/nexus/content/groups/public</url>
     </repository>
+    <repository>
```
chrisknoll, line 283: we should probably ask @leeevans to add this repo as a proxy from the nexus repo for OHDSI because some environments (speaking from personal experience) we have restricted access to 'allowed' repositories and the work around is to proxy through one of the approved sources.

chrisknoll:
I took these updates and merged it into my next characterizaton branch, so please leave this open to add more fixes you find if necessary but I can pull in these changes as part of my characterizaiton work.

PR #2504, opened 2026-04-07
Re-introduce Pathway Characterization

chrisknoll:
Added Pathway analysis classes.
Applied PreAuth permissions to controller endpoints.
Incorporated @p-hoffmann fixes for Characterization.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,10 +41,10 @@ COPY --from=builder /code/opentelemetry-javaagent.jar .
 COPY --from=builder /code/target/WebAPI.jar .
 
 RUN mkdir -p /tmp/trexsql && \
-    unzip -j WebAPI.jar 'BOOT-INF/lib/trexsql-ext-*.jar' -d /tmp && \
-    unzip -j /tmp/trexsql-ext-*.jar 'libtrexsql_java.so_linux_amd64' -d /tmp/trexsql 2>/dev/null || true && \
-    mv /tmp/trexsql/libtrexsql_java.so_linux_amd64 /tmp/trexsql/libtrexsql_java.so 2>/dev/null || true && \
-    rm -f /tmp/trexsql-ext-*.jar
+    (unzip -j WebAPI.jar 'BOOT-INF/lib/trexsql-ext-*.jar' -d /tmp 2>/dev/null && \
+    unzip -j /tmp/trexsql-ext-*.jar 'libtrexsql_java.so_linux_amd64' -d /tmp/trexsql 2>/dev/null && \
+    mv /tmp/trexsql/libtrexsql_java.so_linux_amd64 /tmp/trexsql/libtrexsql_java.so 2>/dev/null && \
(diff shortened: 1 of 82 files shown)


p-hoffmann:
### Code review

Found 3 issues:

1. **Copy-paste bug in `hasEntityAccess`: PATHWAY_ANALYSIS case reads from wrong access map.** The switch case for `PATHWAY_ANALYSIS` in `hasEntityAccess` queries `authz.incidenceRateAccess` instead of `authz.pathwayAccess`. The `isOwner` method directly above (line 309) correctly uses `authz.pathwayAccess`. This means all `@PreAuthorize` annotations that call `hasEntityAccess(id, PATHWAY_ANALYSIS, ...)` will always return `false`, breaking per-entity access control for pathways.

https://github.com/OHDSI/WebAPI/blob/915c6c88661d78da6cfedfc0a0f9598c86674257/src/main/java/org/ohdsi/webapi/security/authz/AuthorizationService.java#L356-L361

2. **`PathwayServiceImpl.getPage()` ignores per-entity access grants.** Unlike `CcServiceImpl.getPage()` (also in this PR), which correctly consults `authz.cohortCharacterizationAccess` for entity-level grants, `PathwayServiceImpl.getPage()` only checks global permissions and sets `dto.setReadAccess(globalRead)` without ever calling `getCurrentUserAuthorizations()` or reading `pathwayAccess`. Users granted entity-level read access to specific pathway analyses (but not global `read:pathway`) will see an empty list. Additionally, lines 243-244 check `"read:pathway-analysis"` / `"write:pathway-analysis"` but all `@PreAuthorize` annotations and DB seeds use `"read:pathway"` / `"write:pathway"` -- the permission strings don't match.

https://github.com/OHDSI/WebAPI/blob/915c6c88661d78da6cfedfc0a0f9598c86674257/src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java#L241-L265

3. **`getById()` uses `.orElseThrow()` making downstream null checks unreachable.** `PathwayServiceImpl.getById()` at line 287 calls `.orElseThrow()` which throws `NoSuchElementException` (mapped to HTTP 500). Callers in `PathwayController` (lines 207, 287, 422, 445) then call `ExceptionUtils.throwNotFoundExceptionIfNull()` expecting a possible null return -- but this null check is dead code since the exception already fired. The result is a 500 instead of a proper 404 for missing pathway analyses. Same issue in `getGeneration()` at line 517 and `cancelGeneration()` at line 494.

https://github.com/OHDSI/WebAPI/blob/915c6c88661d78da6cfedfc0a0f9598c86674257/src/main/java/org/ohdsi/webapi/pathway/PathwayServiceImpl.java#L285-L289

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

chrisknoll:
Great report, made the adjustments as well as adopted the 'page load from db' logic in the case that we can page in-db.

Going to merge this in.

PR #2520, opened 2026-06-20
add additional jvm support

p-hoffmann:
This pull request introduces several improvements and refactorings across the codebase, focusing on enhanced configuration management, improved cache and JPA setup, better OIDC authentication toggling, and codebase simplification.


src/main/java/org/ohdsi/webapi/cohortcharacterization/CcServiceImpl.java
```diff
@@ -324,6 +324,12 @@ public CohortCharacterizationEntity createCc(final CohortCharacterizationEntity
         return saveCc(entity);
     }
 
+    @Override
+    protected UserEntity getCurrentUser() {
```
chrisknoll, line 328: This should come from the Authorization manager.

src/main/java/org/ohdsi/webapi/cohortdefinition/CohortDefinitionService.java
```diff
@@ -255,6 +255,14 @@ public void setTransactionManager(PlatformTransactionManager transactionManager)
 		this.transactionTemplate = new TransactionTemplate(transactionManager);
 	}
 
+	@Override
+	public TransactionTemplate getTransactionTemplate() {
+		if (this.transactionTemplate == null && this.transactionManager != null) {
```
chrisknoll, line 260: Shouldn't the transaction template be managed by Spring and injected as a singleton?  Or is transaction manager based on the lifecycle of the CohortDefinitionService?  The latter seems wrong...

PR #2521, opened 2026-06-20
Enforce authorization on every endpoint (default-deny + complete @PreAuthorize coverage)

p-hoffmann:
## Summary
Hardens WebAPI authorization so no endpoint is reachable without an explicit check. Before this PR, the API filter chain was effectively permit-all and several handlers were without per-operation authorization. 

## Changes
- The API filter chain no longer permits all requests. `security.anonymousAccess.enabled` (default `true`) decides how token-less requests are handled: admitted as the anonymous user, then denied by per-endpoint `@PreAuthorize` on protected endpoints
- **Per-operation `@PreAuthorize` across the API**, using existing 3.0 permissions:
  - admin operations: `admin:cache`, `admin:tags`, `admin:tools`, `admin:security` (user/role/permission + user-import), `admin:source`, admin for statistics
  - source-scoped reads/writes: `read:source`/`hasSourceAccess` on cdmresults, cohort-sample, vocabulary, evidence, ddl, sqlrender
  - per-domain `read:<x>`, entity owner/access expressions for cohort, concept-set, characterization, pathway, feature-analysis, reusable
  - generic `read` for tag/tool/job reads; `isAuthenticated()` for self-scoped endpoints (`/user/me`, notifications)


chrisknoll:
I don't believe these changes are correct.  Please review the authrization mechanics described in the sandbox:

https://github.com/OHDSI/AtlasWebAPISandbox/tree/main/spring-security-authn

It looks like you are requiring global read access in order to get lists of objects from the service.

We discussed this and decided that listing options doesn't require their own permissions. We can filter items by if someone has read permission to the list, but 'give me the list of things' is not something we have a permission for.

src/main/java/org/ohdsi/webapi/security/authc/JwtAuthConfig.java
```diff
@@ -258,21 +263,37 @@ public org.springframework.security.web.AuthenticationEntryPoint unauthorizedEnt
...
 
-    httpSecurityShared.configureDefaults(http); 
- 
+    httpSecurityShared.configureDefaults(http);
+
     http
         .httpBasic(AbstractHttpConfigurer::disable)
-        .authorizeHttpRequests(auth -> auth
-            .anyRequest().permitAll())
+        .authorizeHttpRequests(auth -> {
```
chrisknoll, line 270: This is why we origionally allowed any access to the web service, and then rely on `@PreAuthorize` to harden endpoints that require permissions:

Doing it this way, we have to catalog all the different endepoitns that exist and stuff into this one location all the endpoints that exist on the application that don't require authentication.

The way we decided to handle this was to allow access to any endpoint, and lock down the permission using the pre-authorize annotations.  This also let's us handle anonymous access without special logic checks (other than the standard spring-framework which assigns anonymous identity to requests without bearer token).

All this was thoroughly thought through, showcased in the sandbox repo, and documented.

PR #2526, opened 2026-06-25
Implement API Keys.

### LINKED ISSUE #1884: Implement API Key for programmatic access to WebAPI without authentication

chrisknoll:
Fixes #1884.

This PR introduces API Keys by creating a new controller (for create/list/delete operations), a database table and repository (sec_api_key), and migration scripts to define the tables (in a migration script and in baseline).

### DIFF
--- a/articles/APPKey_HOWTO.md
+++ b/articles/APPKey_HOWTO.md
@@ -1,0 +1,459 @@
+# Personal API Keys — Configuration & Usage Guide
+
+This document describes how personal API keys work in WebAPI, how they are
+implemented, and how to use them for programmatic access to WebAPI services.
+
+---
+
+## Overview
+
+Personal API keys allow authenticated users to generate long-lived credentials
+for use by scripts, automated pipelines, and other programmatic clients that
+cannot perform an interactive login.
+
+An API key request is completely stateless — no session is created or
+maintained. Once a key is presented on a request, the system resolves it to
+the owning user's identity and the request is processed exactly as if that
+user had authenticated via their normal login method.
+
+Key properties of the implementation:
+
+- **No secrets stored in the database.** Only a BCrypt hash of the key's
+  secret component is persisted.
+- **O(1) database lookup during authentication.** The key format includes a
+  public identifier stored in a uniquely indexed column; this avoids scanning
+  every row to find a match before running the expensive BCrypt comparison.
+- **Scoped to the creating user.** Keys can only be managed (listed, revoked,
+  deleted) by the user who created them.
+- **Optional expiry.** Keys can be created with a specific lifetime in days, or
+  with no expiry (never expires).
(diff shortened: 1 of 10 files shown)


chrisknoll:
@p-hoffmann : this PR is independent of #2523, but there is some overlap with base migrations and the JWT auth filter. I will merge this branch with the webapi-3.0 branch and then update this PR.

chrisknoll:
The merge went through with no conflicts, and from my glance, I can see your changes were applied.  This should be ready for review.

Note: while I made IT tests, i had to run them manually because for some reason the Embedded PG instance can not initilize on my local enviornment. I'm hoping the tests will run on the git action.

chrisknoll:
Looks like tests pass and merge was clean, so @p-hoffmann if you want to give a quick review, this can be merged when ready.

PR #2527, opened 2026-06-25
Unify login pipeline to accept an AuthenticatedLogin which the login …

chrisknoll:
This PR unifies the login flow (from authentication to user session + JWT).   An un-implemented part of this is how to manage group synchronization. The OIDC attempted to do this, but there is infrastructure in place to manage Group->Role associations, however, it is tightly coupled to LDAP and UserImportJobs.  We can decide if we want to maintain the UserImport functioniality, but my prefernce for 3.0 would be to synchronize roles based on user login (at login time, certain auth providers will present claims, so we can do IdP claim->webapi roles mapping during the login process).   

I attempted to handle group synchronziaton, but then learned about how the RoleGroupEntity depends on LDAP, UserImportJobs, and I decided to put this into a phase 2 (either we can extend this PR to cover group sync or we can leave this PR to not do group synchronization and juse leave a place-holder for furter expansion to handle group mapping).

### DIFF
--- a/articles/LoginPipeline.md
+++ b/articles/LoginPipeline.md
@@ -1,0 +1,438 @@
+# Unified Login Pipeline
+
+## Overview
+
+WebAPI implements a **unified authentication pipeline** that consolidates all authentication types (Database, LDAP, Windows/Kerberos, and OIDC) through a single entry point. This design ensures consistent user lifecycle management, role synchronization, and JWT token generation across all authentication methods.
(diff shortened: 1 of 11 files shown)


chrisknoll:
I'm going to commit these changes as it unifies the login pipeline, and then my next PR will involve implementing role/claim/nt groups -> webapi Roles functionality.

